PR #37, opened 2018-04-24
Add index.d.ts

varHarrie:


iamkun:
We are just doing this actually, check out our road map #39 here. 

PR pending ...

iamkun:
thanks close #56

PR #52, opened 2018-04-25
Fix isLeapYear docs

CarlRosell:
Hi!
This project looks very promising! 
I will definitely give it a try on my projects.
I found a small error in the `isLeapYear` docs as the function does not take any arguments.


iamkun:
nice

PR #59, opened 2018-04-26
Get day and month name arrays by splitting strings

okonomiyaki3000:
Splitting a single string into an array of strings may be smaller than writing an array literal (depending on the length of the array).

If being tiny is your goal, this will shave off a few bytes.


iamkun:
nice point!  I am just wondering if might reduce code readability 

Lengthy lines of code vs readability

okonomiyaki3000:
@iamkun I think the difference in readability is minor but that's just me. I think a slightly bigger issue is that these arrays are being created (in whichever way) each time `format` is called. They could be pulled out to a higher scope and only created once. Same goes for the regular expression in that function I guess. I mean, it depends on performance difference between creating an array and looking up to a higher scope. I don't really know which is costly or by how many picoseconds.

schiem:
I agree with @okonomiyaki3000 about moving them to a higher level, it would also potentially open the door for creating a function that would let people override them with location specific versions at runtime, which could allow for localization without the needing to increase the overall size of the library.

iamkun:
I agree. 

Moving constants to a higher scope level may help us easily support i18n / locals.

Issue #60, opened 2018-04-26
Missing several format options


schiem:
The Moment [docs](https://momentjs.com/docs/#/parsing/string-format/) have a number of format options which are missing from DayJS.  Some of them, like the locality specific ones, I think it makes sense to exclude, but there are others that are missing that could be included without having a serious impact on the library size.

Here are the missing formats that I propose are added:

- Q  - Quarter of the year
- Do - Day of month with ordinal
- DDD/DDDD - Day of year
- X - Unix timestamp
- x - Unix timestamp with ms
- h/hh - 12 hour time
- k/kk - 24 hour time beginning at 1
- a/A - Post and ante meridiem
- S/SS/SSS - Milliseconds

However, this would nearly double the size of the format function, which isn't ideal.  Are there any of those that shouldn't be added?

Frondor:
In my humble opinion, the only useful ones in that list are `h/hh` and `k/kk`. The rest doesn't look like human-friendly formats.

schiem:
I would tend to agree, I think that Q, DDD/DDDD, X/x, and S/SS/SSS are definitely not necessary.   I think if h/hh is included, then a/A should probably be included as well, or you lose the ability to show specific times (1:00 vs 1:00 AM).

"Do" is kind of up in the air for me, because I've had clients who've insisted on having the ordinal (1st, 2nd, 3rd, etc) shown, but it's something that a user could implement on their own.  I think most of it comes down to "what should be provided by a library" vs. "what can a user be reasonably expected to write themselves," and I think that's mostly up to the owner of the repo.

iamkun:
I tend to agree adding h/hh & a/A(with local). And leave the rest formats to a separate plugin, maybe called `AdvFormat`
It will be nice if we implement a plugin system like this.
```javascript
import { AdvFormat } from 'dayjs/plugin'
dayjs.extend('AdvFormat') // a plugin might add a new API or overwrite the existing one
//here it just overwrite the core.format
dayjs.format('Do SSS ... something less useful format')
```

schiem:
I've updated #68 to only have AM/PM and 12 hour time support, with additional options included in another plugin (and the addition of a plugin system).

schiem:
The PR from #68 has been split into #91 and #92.

PR #68, opened 2018-04-27
Display format update

schiem:
This adds many of the missing display formats referenced in #60 .  It does add an additional ~30 lines of code, which is ~10% of the entire library size.  I think that some of the options are worthwhile, but there are some that might not be needed.

### DIFF
--- a/plugins/advancedFormat.js
+++ b/plugins/advancedFormat.js
@@ -1,0 +1,83 @@
+import * as Utils from "../src/utils"
+
+let advFormat
+export default advFormat = {
+  name : "format",
+  method : function(formatStr = 'YYYY-MM-DDTHH:mm:ssZ') {
+    const weeks = 'Sunday.Monday.Tuesday.Wednesday.Thursday.Friday.Saturday'.split('.')
+    const months = 'January.February.March.April.May.June.July.August.September.October.November.December'.split('.')
+    const suffixes = ['th', 'st', 'nd', 'rd']
+
+    return formatStr.replace(/Y{2,4}|Q|M{1,4}|Do|D{1,2}|d{1,4}|X|x|H{1,2}|h{1,2}|k{1,2}|a|A|m{1,2}|s{1,2}|S|Z{1,2}/g, (match) => {
+      switch (match) {
+        case 'YY':
+          return String(this.$y).slice(-2)
+        case 'YYYY':
+          return String(this.$y)
+        case 'Q':
+          return Math.ceil((this.$M + 1) / 3)
+        case 'M':
+          return String(this.$M + 1)
+        case 'MM':
+          return Utils.padStart(String(this.$M + 1), 2, '0')
+        case 'MMM':
+          return months[this.$M].slice(0, 3)
+        case 'MMMM':
+          return months[this.$M]
+        case 'Do': {
+          const digits = Utils.padStart(String(this.$D), 2, '0').slice(-2)
+          const ls = parseInt(digits[1], 10)
+          return this.$D + suffixes[digits[0] === '1' || ls > 3 ? 0 : ls]
+        }
+        case 'D':
+          return String(this.$D)
+        case 'DD':
+          return Utils.padStart(String(this.$D), 2, '0')
+        case 'd':
+          return String(this.$W)
+        case 'dddd':
+          return weeks[this.$W]
+        case 'X':
+          return Math.floor(this.$d.getTime() / 1000)
+        case 'x':
+          return this.$d.getTime()
+        case 'H':
+          return String(this.$H)
+        case 'HH':
+          return Utils.padStart(String(this.$H), 2, '0')
+        case 'h':
+          if (this.$H === 0) {
+            return 12
+          }
+          return this.$H < 13 ? this.$H : this.$H - 12
+        case 'hh':
+          if (this.$H === 0) {
+            return 12
+          }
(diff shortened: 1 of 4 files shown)


schiem:
I definitely agree.  I broke all of them out into separate commits so it would be possible to remove any of the formats that aren't needed.

schiem:
Would there possibly be a way to extend the format function so that it could accept new display format options from the user, so users could include any extra format options they need without increasing the library size?  Several users have mentioned specific display format options that they would use that are missing, but I've also seen several users say that they don't think any more are required since they don't personally use them.

That leads me to believe that the solution that would make the most users happy would be finding a way to extend it so users who need the extra format option could have them by including them separately, but the library would stay compact for users who don't need them.

Barring that, I think that at _most_ 1-2 of the newly introduced formats should be merged in, for the sake of the size of the library.

iamkun:
Think we might need a plugin system, like jQuery.

```
import dayjs from ''dayjs'
import { advancedFormat } from 'dayjs/plugin'
dayjs.extend(advancedFormat) // will overwrite core.format

//then

dayjs.format('full format options support')

```

warapitiya:
I completely agree with the library size, and the plugin system is a great idea. Anyway, we needed this sort of a formatting technics. I would love to see this get implemented soon.

schiem:
I started working on a plugin system.

It works by taking in objects in the format of:
```
{
  name : "methodName",
  method : function() { //plugin method }
}
```

So to override the format method, you would do something like:
```
{
  name : "format",
  method : ...
}
```

The plugins are then included in the core by doing:
```
import advancedFormat from "./plugins/advancedFormat"

dayjs().extend(advancedFormat);
dayjs().format(); //New format
```
This operates on the class prototype, so any calls to the function after the plugin is loaded will use the new function.

I also reduced the additional formats in the core down to h/hh and a/A for 12 hour time and AM/PM.  The rest of the new formats are included in the plugin.  The downside to this plugin approach is that all of the code for the format function had to be copied into the advancedFormat plugin as well, but I doubt that the core format function will be extended too much in the future.

Let me know what you think!

iamkun:
Looks cool.

But seems too many repeated code in this plugin system design (utils, old format func). 
What I am thinking is:  
1 assign core Utils to this.$u so that we can use it directly instead of imported it 
```javascript
this.$u = Utils  
```
2 while extend an API like `format`, we might could pass the argument to the core.format then pass to plugin.format. 
```javascript
dayjs.extend(function(args) {
var oldFormat = this.format
var oldresult = this.format(args) // something like this "2018-01-01 Do" 'Do' is not supported in core.format
var result = pluginFormat(oldresult) // process 'Do' return "2018-01-01 1st"
return result
})
```

3 global plugin
I don't have any idea yet. But could we extend our plugin globally ? Like
```javascript
dayjs.extend(advFromat) // instead of `dayjs().extend(advFromat)`
```

schiem:
I definitely agree with 1, I'll move the Utils to a member object so it will be accessible.

I really like the approach in 2, since it reduced the amount of duplicated code.  Since the current approach involves overwriting the method, I don't think directly calling the old method will work (but I'll need to test it).  For example:
```
function(args) {
  var oldFormat = this.format //format has already been overwritten, so this.format is the current function
  ... 
}
```

Instead, could we move the old functions into an `old` member object?  Like so:
```
extend(plugin) {
  this.old[plugin.name] = this[plugin.name]
  this[plugin.name] = plugin.method
}
```
And then in the plugin:
``` 
function(args) {
  var oldFormat = this.old['format']
  var oldResult = oldFormat(args)
  ...
}
```


For number 3, could you explain what you mean a bit more?  Even though the extend method is being called on an instance, it's overwriting the original prototype, which is about as global as it gets (it will retro-actively replace the method in any existing instances).
Here's some code I just ran to demonstrate:
```
var dayjs = require("./dist/dayjs.min.js");

var dayjsInstance = dayjs();
console.log(dayjsInstance.format()); //Output: 2018-05-01T13:31:07-04:00

dayjs().extend({
    name : "format", 
    method : function() { 
        return "firing new format"; 
    }
}); 

console.log(dayjsInstance.format()); //Output: firing new format
console.log(dayjs().format()); //Output: firing new format
``` 

That might not be what a user would expect to happen though, since it is called on an instance instead of the base model.

iamkun:
@schiem Would you mind cherry-pick your commits about  h/hh/a/A into a separate pr please? So that I could merge it before we finish our plugin system design.

iamkun:
For number 3, my mistake sorry. I haven't noticed that you've overtired it's prototype. I thought it was an instance level change not global.
Still, our user may have the same confusion like me. So how can we make this api more friendly? May be like this `dayjs.extend(advFromat)`

schiem:
@iamkun Absolutely, I'll move those commits over when I get home from work.

One of the issues with using  the `dayjs` object instead of `dayjs()` is that dayjs is just a function that returns a new instance, like so:

```
var dayjs = function(config) {
    return new Dayjs(config);
}
```

which means that the `dayjs` variable doesn't have access to the underlying Dayjs class.  When I get home I'll play around with doing something like:

```
let dayjs = (config) => {
    return new Dayjs(config);
}
dayjs.extend = (plugin) => {
    Dayjs[plugin.name] = plugin.method;
}
export default dayjs(config);
```

But I'll have to dig around with that before I can say for sure that that approach will work.

schiem:
@iamkun I've split that into 2 new separate PRs (#91 and #92), so I'll close this one out.  #92 cover all three points you brought up earlier.

PR #74, opened 2018-04-27
Query method supports Dayjs constructor args

huihuimoe:
like momentjs, it supports
`dayjs().isAfter("2020")` is now after 2020years
`dayjs().isBefore("2000-4-1")` is now before 2000-4-1


iamkun:
Not sure if we need this. Simply you can do `dayjs().isAfter(dayjs('2020-01-01'))` to meet your need.

Pending... , waiting for opinions.

devshawn:
For some input, I'd likely agree with @iamkun's suggestion here of not needing it... fairly easy to just make a new dayjs object with your second date for comparisons.

iamkun:
@huihuimoe Finally, agreed with you. But I'm afraid this pr is too old to get merged.

Would you please make a new pr? With the code only will be fine. I'll do the docs together.

Thanks

PR #75, opened 2018-04-27
Run Prettier & Update grammar in README

Mohamed3on:
Ran Prettier on the README, and changed things like `return Number` to `returns a Number` for better clarity.


Frondor:
That's how docblocks actually work, you declare the returning entities by their type, which is always singular, so no need to say `return a Number`. `Number` is singular already, you can't return multiple numbers as `Number` type, that would be an `Array` instead, and still, `Array` is singular as well. Just my opinion, tho.

iamkun:
Prettier is so cool! 

I may merge this pr first. But still, I'm looking for a better way declaring the returning value.

ReadMe.md
```diff
@@ -21,29 +22,38 @@ English | [简体中文](./ReadMe.zh-CN.md)
 > Day.js is a minimalist JavaScript library for modern browsers with a largely Moment.js-compatible API. If you use Moment.js, you already know how to use Day.js.
 
 ```js
-dayjs().startOf('month').add(1, 'day').set('year', 2018).format('YYYY-MM-DD HH:mm:ss');
+dayjs()
+  .startOf('month')
+  .add(1, 'day')
+  .set('year', 2018)
+  .format('YYYY-MM-DD HH:mm:ss');
```
iamkun: Can we change this code back to its original format? I think that looks cool. 😃

Mohamed3on:
@iamkun better?

iamkun:
@Mohamed3on Cool~ Thanks

PR #77, opened 2018-04-28
Implementing localization

### LINKED ISSUE #36: Locales?
will there be any locales avaliable?

Frondor:
First try, WIP, waiting for reviews

Besides of implementing localization, this PR adds a new way to work with the library with more flexibility without breaking changes.

A full example of the new APIs and method features

```js
// optionally, you can use a named import to get the Dayjs class and pass your own configuration object.
import { Dayjs }, dayjs from 'dayjs'
```

```js
// Instantiate your own Dayjs object, and use it anywhere in your app with the provided default config
// myDayjs.js
import { Dayjs }, dayjs from 'dayjs'
import ESP from 'dayjs/locales/es'

// You can still do
dayjs().format() // "2018-04-01T23:10:59Z"

// all of these settings are optional
const config = {
    date: new Date(), // optional
    locale: ESP, // default locale for this and cloned instances
    format: 'YYYY / MM / DD HH:ss', // default format, used for empty calls to format() and object serialization
    // more config...
}

export default new Dayjs(config)

/*
* ----------------------------------
*/
// Somewhere in your app
import myDayjs from '../myDayjs'

myDayjs.set('year', 1991).format() // "1991 / 12 / 1 09:30"
myDayjs.format('YYYY') // "2018"
myDayjs
    .set(new Date(2018, 5, 1)) // you can now pass a Date or Dayjs instance to set!
    .setLocale() // change the instance locale from spanish back to english
    .format('dddd - MMMM', {
        // override default format and WeekDays locale (only WD)
        WD: ['Mondy', 'Tuesdy', 'Wendy', 'Thursdy', 'Fridy', 'Saty', 'Sundy']
    }) // "Tuesdy - May"

// And now, objects with Dayjs instances are being serialized to a JSON string using the default format
JSON.stringify({ name: 'Jon', birthday: myDayjs }) //  '{name: "Jon", birthday: "2018 / 12 / 1 09:30"}'
```

closes #36


src/index.js
```diff
@@ -200,16 +209,16 @@ class Dayjs {
...
   }
 
   subtract(number, string) {
     return this.add(number * -1, string)
   }
 
-  format(formatStr = 'YYYY-MM-DDTHH:mm:ssZ') {
-    const weeks = 'Sunday.Monday.Tuesday.Wednesday.Thursday.Friday.Saturday'.split('.')
-    const months = 'January.February.March.April.May.June.July.August.September.October.November.December'.split('.')
+  format(formatStr = this.$format, L = {}) {
```
atilacamurca: I think this could be improved by using:

```js
format(formatStr = this.$format, L = this.$L)
```

Then I think you can get rid of next lines short circuits `||`. What do you think?

Frondor: Really depends, because you may do `format.('...', { WD: [0,1,2,3,4,5,6] })` to override the week days locale, but leave `MONTHS` untouched with the default instance locale. You're not setting a locale, just providing overrides for certain locales.

Frondor:
Switching target branch to @next

src/index.js
```diff
@@ -163,9 +172,15 @@ class Dayjs {
     return this
   }
 
-  set(string, int) {
+  set(unitOrDate, int) {
+    if (typeof unitOrDate === 'object' &&
```
iamkun: @Frondor 
> myDayjs.set(new Date(2018, 5, 1)) // you can now pass a Date or Dayjs instance to set!

Could you explain the meaning of you code here, please?  Don't really understand.

Frondor: Just for the record, this is the old PR, the new one is #90 
This code is kind of explained in the test case https://github.com/Frondor/dayjs/blob/5f574850d1cb75c6bd2a8b1684a94346f1d501b7/test/get-set.test.js#L69
In the past, you were limited to set certain time units over an already defined date (`$d`). Now that code checks whether the `unitOrDate` passed is either a time unit, a `Date` object, or a `Dayjs` instance. So you can set the whole date if needed, not just an unit. Useful for inheriting configurations like locale, plugins and default format, and create date objects out of a single source, and keeping the Dayjs immutable since that's how `set` works.

Frondor: set(unitOrDate, int) {
    // if unitOrDate is a Date or Dayjs object
    if (typeof unitOrDate === 'object' &&
      unitOrDate.constructor.name.indexOf(['Date', 'Dayjs'] > -1)) {
      const self = this.clone() // don't mutate this instance
     // set $d date object depending on whether its a native Date object, or a Dayjs instance
      self.$d = unitOrDate.toDate ? unitOrDate.toDate() : unitOrDate
      return self // done
    }
    // if this is not a date object nor an int just fail silently (this code was already here)
    if (!Utils.isNumber(int)) return this
   // unitOrDate is definitely a integer (unit)
    return this.clone().mSet(unitOrDate, int)
  }

iamkun: @Frondor Got it, thanks.

PR #81, opened 2018-04-30
Add .json extension to .eslintrc

ohbarye:
It's a kind of nitpicking though, `.eslintrc` without any extension is deprecated.

ref: https://eslint.org/docs/user-guide/configuring#configuration-file-formats


iamkun:
thanks

PR #90, opened 2018-05-02
Implementing localization

### LINKED ISSUE #36: Locales?

Frondor:
First try, WIP, waiting for reviews

Besides of implementing localization, this PR adds a new way to work with the library with more flexibility without breaking changes.

A full example of the new APIs and method features

```js
// optionally, you can use a named import to get the Dayjs class and pass your own configuration object.
import { Dayjs }, dayjs from 'dayjs'
```

```js
// Instantiate your own Dayjs object, and use it anywhere in your app with the provided default config
// myDayjs.js
import { Dayjs }, dayjs from 'dayjs'
import ESP from 'dayjs/locales/es'

// You can still do
dayjs().format() // "2018-04-01T23:10:59Z"

// all of these settings are optional
const config = {
    date: new Date(), // optional
    locale: ESP, // default locale for this and cloned instances
    format: 'YYYY / MM / DD HH:ss', // default format, used for empty calls to format() and object serialization
    // more config...
}

export default new Dayjs(config)

/*
* ----------------------------------
*/
// Somewhere in your app
import myDayjs from '../myDayjs'

myDayjs.set('year', 1991).format() // "1991 / 12 / 1 09:30"
myDayjs.format('YYYY') // "2018"
myDayjs
    .set(new Date(2018, 5, 1)) // you can now pass a Date or Dayjs instance to set!
    .setLocale() // change the instance locale from spanish back to english
    .format('dddd - MMMM', {
        // override default format and WeekDays locale (only WD)
        WD: ['Mondy', 'Tuesdy', 'Wendy', 'Thursdy', 'Fridy', 'Saty', 'Sundy']
    }) // "Tuesdy - May"

// And now, objects with Dayjs instances are being serialized to a JSON string using the default format
JSON.stringify({ name: 'Jon', birthday: myDayjs }) //  '{name: "Jon", birthday: "2018 / 12 / 1 09:30"}'
```

closes #36


iamkun:
I agreed with most of your code, thanks.

Still, I don't think we should change the `JSON.stringify` result, or let user set default format option.

JSON should be a standard data structure, to be as clear and simple. 

Compare these two JSON:
```javascript
{"date": "2011-10-05T14:48:00.000Z"}
{"date": "Monday - 11 AM - 2018 - 01 -1 "}
```
How do you think?

If our user have to change the format result, may be they could just do:
```javascript
const defaultFormat = 'YYYY-MM-DD A'
//then
dayjs().format(defaultFormat)
// somewhere else
dayjs().format(defaultFormat)
```

Frondor:
> Still, I don't think we should change the JSON.stringify result
JSON should be a standard data structure, to be as clear and simple.

Depends on how you're using this library.
Let's start by stating that, in my opinion, there's no point to do `JSON.stringify(dayjs())`. If you're doing that, then you should be using the correct methods such `dayjs().toString()` or simply `String(dayjs())` which shall return the standard format you're talking about.
But when you're working with collections of objects which contains `dayjs` instances in them, and you're about to convert each one to string because that's how your view templating engine works (angular/react/vue, etc...), you'll need to add an extra iteration over that collection and manually call `format(desiredFormat)` on each one manually.
I already provided [an example](https://codepen.io/Frondor/pen/BxpLZG?editors=0012) of how this works, and the benefit you can get from allowing the user to set a default format for an instance and its childs (cloned ones). It's inspired by how [Carbon](https://github.com/briannesbitt/Carbon) works in [Eloquent's](https://laravel.com/docs/5.6/eloquent-mutators#date-mutators) implementation of date casting.
For instance, imagine I want to retrieve a collection of `posts` from a database, and the `created_at` / `updated_at` fields are automatically casted to `dayjs` instances to make working with dates easier, and display a human readable format with localization support. But when you send that object directly through a restful api response, the `post` object would look like
```js
{
  "title": "A node.js example",
  "created_at": "",
  "updated_at": "",
}
```
As you can see, those are not human readable dates, and you'll need to instantiate `dayjs` once again in the frontend, once per each date property you're displaying in the view, and per post... So you can now reformat it to the format you need.
So the solution here would be to iterate over the collection of posts once retrieved, format each post dates and then you're good to send the data back to the client. Again, in my humble opinion, that could be easily avoided by providing a default format for those instances and letting `toJSON` use it (again, `toJSON` should be meant for this kind of use only. For direct `toString` casting, the user should go for `toString*` methods instead).

If you still need a real world example of what I'm talking about, just scaffold the most simple express server and try this by yourself; imagine the object passed to `res.send` is a `post` object which contains a dayjs instance.

app.get('/', function (req, res) {
  const dayjs = {
    toJSON () {
      return 123 // a default format for human readable dates
    }
  }
  res.send({created_at: dayjs});
});

This is my argument for the `toJSON` method and support for instance's default format. It's completely valid for you disagree, and if so, I'll remove that from this PR. I just hope you get the point of why `toJSON` shouldn't need to behave just like any other `toString` method.

iamkun:
I'm gonna to merge this pr fist. And put locale && plugin 's code altogether. Then we might could discuss the new release in one pr.

PR #92, opened 2018-05-02
Added a plugin system.

schiem:
The plugin system allows users to extend dayjs with additional plugins,
or change existing functionality.

The plugins work by calling the extend method:

```
dayjs.extend({
  n: "methodName",
  m: function(arguments) {
     //
  }
})
```

When overwriting a function, the original function can be referenced from
inside the plugin using:

```this.getOld(pluginName)```

For example:

```
dayjs.extend({
  n: "format",
  m: function(formatStr) {
    let oldFormat = this.getOld("format")
    let oldResults = oldFormat(formatStr)
    ...
})
```

This commit also introduces a plugin which extends the format options to include:
-Q: Quarter of the year
-k/kk: 1-24 hour time
-ddd: Date of the month with ordinal (1st-31st)
-S: Milliseconds

### DIFF
--- a/plugins/advancedFormat.js
+++ b/plugins/advancedFormat.js
@@ -1,0 +1,32 @@
+let advFormat
+export default advFormat = {
+  n : "format",
+  m : function(formatStr = 'YYYY-MM-DDTHH:mm:ssZ') {
+    let oldFormat = this.getOld('format')
+    let oldResult = oldFormat(formatStr)
+
+    const suffixes = ['th', 'st', 'nd', 'rd']
+
+    return oldResult.replace(/Q|ddd|X|x|k{1,2}|S/g, (match) => {
+      switch (match) {
+        case 'Q':
+          return Math.ceil((this.$M + 1) / 3)
+        case 'ddd': {
+          const digits = this.$u.padStart(String(this.$D), 2, '0').slice(-2)
+          const ls = parseInt(digits[1], 10)
+          return this.$D + suffixes[digits[0] === '1' || ls > 3 ? 0 : ls]
+        }
+        case 'k':
+          return this.$H === 0 ? 24 : this.$H
+        case 'kk':
+          return this.$u.padStart(String(this.$H === 0 ? 24 : this.$H), 2, '0')
+        case 'X':
+          return Math.floor(this.$d.getTime() / 1000)
+        case 'x':
+          return this.$d.getTime()
+        default: //'S'
+          return this.$u.padStart(String(this.$ms), 3, '0')
+      }
+    })
+  }
+}
--- a/plugins/advancedFormat.js
+++ b/plugins/advancedFormat.js
@@ -11,20 +11,20 @@ export default advFormat = Dayjs => {
         case 'Q':
           return Math.ceil((this.$M + 1) / 3)
         case 'ddd': {
-          const digits = this.$u.padStart(String(this.$D), 2, '0').slice(-2)
+          const digits = this.Utils().padStart(String(this.$D), 2, '0').slice(-2)
           const ls = parseInt(digits[1], 10)
           return this.$D + suffixes[digits[0] === '1' || ls > 3 ? 0 : ls]
         }
         case 'k':
           return this.$H === 0 ? 24 : this.$H
         case 'kk':
-          return this.$u.padStart(String(this.$H === 0 ? 24 : this.$H), 2, '0')
+          return this.Utils().padStart(String(this.$H === 0 ? 24 : this.$H), 2, '0')
         case 'X':
           return Math.floor(this.$d.getTime() / 1000)
         case 'x':
           return this.$d.getTime()
         default: //'S'
-          return this.$u.padStart(String(this.$ms), 3, '0')
+          return this.Utils().padStart(String(this.$ms), 3, '0')
       }
     })
   }
--- a/plugins/advancedFormat.js
+++ b/plugins/advancedFormat.js
@@ -1,9 +1,8 @@
 let advFormat
-export default advFormat = {
-  n : "format",
-  m : function(formatStr = 'YYYY-MM-DDTHH:mm:ssZ') {
-    let oldFormat = this.getOld('format')
-    let oldResult = oldFormat(formatStr)
+export default advFormat = Dayjs => {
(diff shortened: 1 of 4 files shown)

### CODE REVIEW

**Comment by schiem** [comment]
@frondor had suggested a system more like Vue js, where the prototype is passed into the plugin function and the plugin is able to alter it however it would like.  It would look something like:

`src/index.js`
```
Dayjs.prototype.extend = (fn) => {
  fn(Dayjs)
}
```
User's `app.js`
```
dayjs.extend( (Dayjs) => {
    let oldFormat = Dayjs.prototype.format
    Dayjs.prototype.format = (formatStr) => {
      let oldResponse = oldFormat(formatStr)
      ...
     }
})
```

Both that approach and the current one should work equally well, it seems to be a matter of preference and how much control plugin authors should be given over the default DayJS object.

**Comment by iamkun** [comment]
@schiem I agree giving plugin a full access to the Dayjs class. That makes it a real plugin 😋

**Comment by iamkun** [comment]
Plus, I thinks we might better use `Utils` rather than `this.$u` is our main code. It will reduce the code size after minimize.

**Comment by Frondor** [comment]
Just one thing, you may want to define whether you want to extend `prototype` or `__proto__` (class vs instance).

In Vue you can't do `new Vue({...}).use(myPlugin)`. Your are supposed to install the plugin first, and then instantiate the Vue object which shall use all registered plugins, like `Vue.use(myPlugin); const app = new Vue({...})`. That's why I wanted to export `Dayjs` class in first place.

I know its a total different approach, but just take a look, and tell later: https://codepen.io/Frondor/pen/MGopZx?editors=0010
Influenced by the current Vue source code. **It adds a new feature called `hooks`. While the plugin is executed as soon as it is registered / installed, the user can use `plugin hooks` to set information at instantiation time**.

```js
// The plugin
const lastNamer = {
  install (Person, opts = {}) {
    const formatter = s => {
      return typeof opts.formatter === 'function' ? opts.formatter(s) : s
    }

    Person.addHook(function (p) {
      p.lastName = formatter(p.cfg.lastName)
    })

    Person.prototype.getName = function () {
      console.log('I am', `${this.firstName} ${this.lastName}`)
    }
  }
}

// Plugin installation with options
Person.extend(lastNamer, {
  formatter: ln => ln.toUpperCase()
})

// Library usage
const person = new Person({
  firstName: 'Jon', // supported by default
  lastName: 'Snow' // supported after plugin
})

console.log(person)
person.getName() // new custom method
```

harrysarson:
If I can weigh in with my two cents:

From the README:

>  💪 Immutable

I believe it would be great to implement plugins in a way which extends this concept of immutability.

By this I mean, instead of mutating `dayjs.prototype` (which is effectively a global variable) when extending `dayjs` with plugin, they would instead write:

```js
 const extended_dayjs = dayjs.extend(plugin);
``` 

This would leave the original `dayjs` unchanged and the user can use their new, shiny `extended_dayjs` however they wish.

Frondor:
Have to agree with harry. After #77 you will be able to do
```js
import { Dayjs } from 'dayjs'

class MyDayjs extends Dayjs {
  constructor (cfg) {
    super(cfg)
  }
}
```

harrysarson:
Again just my personal views here:

@Frondor 

>  besides of the `setLocale `method

I think a `setLocal` method that returned a new instance of `dayjs` rather than modifying the current instance would be better. 

@schiem

> * Adding a plugin creates a new subclass with the extended method, and the extend method returns a new "factory" for that subclass.

I think this would be a sensible approach.

schiem:
@iamkun That makes sense to me.  Any thoughts on whether it makes more sense to have the Dayjs core handle installing the plugins vs. having the plugins install themselves using an install method?

Having the core install them would reduce the complexity for the plugin author, but it would also reduce the amount of flexibility the plugin author has.

iamkun:
@schiem Looks good.

1. Check out my comment above. 
> Plus, I thinks we might better use Utils rather than this.$u....

2. Another problem here, should we make a backup for our core APIs, may be like this.$old you used before. If on plugin modified the behavior of one core API, other plugin may also affected, if they relay on the same API. Here coms the question you've mentioned yesterday,.
> Any thoughts on whether it makes more sense to have the Dayjs core handle installing ...

schiem:
I've switched the `this.$u` to use the Utils library directly in the core, and I've swapped `this.$u` with a function that returns the Utils library for plugins to use, like so:
```
this.Utils().padStart(...)
```

The reason for that change is that `this.$u` had the Utils library assigned to it for every dayjs instance that was created, whereas `this.Utils()` should always return a reference to the same library.


For point 2, I think that since we're passing in the Dayjs prototype for plugins to modify, it makes sense for them to handle the install themselves (since there might be plugins, like locales, that might modify a variable instead of a method).  And if they're handling the install themselves, then I think it might make sense for them to keep track of the original function themselves.  This would also reduce conflicts when installing multiple plugins that operate on the same method.

For instance, say there were two different format plugins that added slightly different functionality.  If the core was responsible for keeping track of the original `format` method, then you might end up with something like:

```
dayjs.extend(format1) //dayjs().format() now calls original format, then format1
dayjs.extend(format2) //dayjs().format now calls original format, then format 2, but never format 1
```

If the plugins are responsible for saving the original method, then they should be storing whichever method is currently on the plugin.  For example:

```
plugin1 = dayjs => {
    const oldFormat = dayjs.prototype.format //oldFormat is the original format
    dayjs.prototype.format = (formatStr) => {
        ...
    }
}

dayjs.extend(plugin1) //dayjs().format() is now calling plugin1 format, which calls the original format

plugin2 = dayjs => {
    const oldFormat = dayjs.prototype.format //oldFormat is now plugin1 format
    dayjs.prototype.format = (formatStr) => {
        ...
    }
}

dayjs.extend(plugin2) //dayjs().format now calls plugin2 format, which calls plugin 1 format, which calls the original
```


It also gives the plugin authors the opportunity to _not_ call the original if they need to completely change the core functionality.  I think that not enforcing a specific install method gives the plugin authors a lot more freedom to extend in the way they need to.

iamkun:
I'm gonna to merge this pr fist. And put locale && plugin 's code altogether. Then we might could discuss the new release in one pr.

PR #99, opened 2018-05-03
Bug fix && bundle update

iamkun:


.babelrc
```diff
@@ -10,7 +10,8 @@
         [
           "@babel/preset-env",
           {
-            "modules": false
+            "modules": false,
+            "loose": true
```
hzoo: yeah if you want this to be smaller it's a good idea. (I'm a maintainer on Babel)
or upgrade to v7 which might have more savings

iamkun: THX, seems we are already using babel V7.

I'm not sure if it's a good approach turning loose mode on. But actually I've got what I want, a little bit small bundle size and no Object.defineProperty compatibility problem in IE8

PR #100, opened 2018-05-04
chore: add Node.js 8 and 10

DanielRuf:


iamkun:
Yes, I know what's this config used for.

I'm not sure if we need it. Actually we have all these tests in the past. You could see a `travis-deploy-once`  used in the `.travis.yml` at that time.

Just wondering if our unit test runs perfectly on Node6, a lower node version, dose that mean it shoud be good on 8 and 10?

Plus, will this expand the total CI process time?

DanielRuf:
> Plus, will this expand the total CI process time

No, not in general. These do normally use a test matrix.

> I'm not sure if we need it. Actually we have all these tests in the past. You could see a travis-deploy-once used in the .travis.yml at that time.

Why not? There can be always changes in Node.js which affect libraries.

> You could see a travis-deploy-once used in the .travis.yml at that time.

You can constrain it to a specific Node.js version / test.

DanielRuf:
Regarding different Node.js versions and testing: https://github.com/webpack/watchpack/pull/80#issuecomment-386285716

Also older versions (4 is EOL) often break and dependencies are often not compatible anymore (eg puppeteer requires Node.js 6+).

DanielRuf:
https://github.com/webpack/webpack-cli/blob/master/.travis.yml#L42

Travis Ci is very flexible.

iamkun:
All right, may be we could give it a try

iamkun:
Let's give it a try.

PR #101, opened 2018-05-04
chore: add version to package.json

DanielRuf:
The version in the package is meant for npm and should be set in general to prevent specific cases where this could be problematic.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "dayjs",
-  "version": "0.0.0-development",
+  "version": "1.5.19",
   "description": "",
   "main": "dist/dayjs.min.js",
   "scripts": {


iamkun:
There's no need. We will upload `package.json` with the updated version string as well as a `dist` folder with bundled files to npm in each release automatically.

DanielRuf:
> There's no need. We will upload package.json with the updated version string as well as a dist folder with bundled files to npm in each release automatically.
> 
> Try download a dayjs npm package and you'll see it.

Sure but what about contributors, other maintainers and people who want to create forks. And I have never seen a big package which does not set the version string in there. There are scripts to automatically set the version like semantic-release and others and update the package.json in the repository.

Personally I find it confusing as it is not clear in the repository history which is the current version when traveling through old commits and checking the package.json as reference (we would have to check the tags and compare them with the date of the current tree).

iamkun:
I agree, seems currently, the only way is to check git tags.

May be, we need a script that could generate a CHANGELOG file?

DanielRuf:
> May be, we need a script that could generate a CHANGELOG file?

Probably one of these
https://www.npmjs.com/package/semantic-release
https://www.npmjs.com/package/changelog
https://www.npmjs.com/package/conventional-changelog-cli

iamkun:
Will try. Thanks

iamkun:
Leave the version to semantic-release.

But a change log is needed.

PR #109, opened 2018-05-07
localization && Plugin pending...

iamkun:


aalises:
@iamkun perfect! I will then propose some changes to the locales (adding plural/singular forms and patterns for past, present and future), to implement the TimeAgo API with the locales effortlessly :)

iamkun:
@aalises Fantastic! Seems that I must seize the time 😁

PR #110, opened 2018-05-07
Add Hour and Minute to diff()

kyleknighted:
This PR adds the ability to get the `diff()` of hours and minutes in addition to the types already available.

I added the two units to the test file, but this test was failing for me on master locally, so I didn't investigate too deeply as to the cause. 

Feel free to reach out if you have any concerns with this PR.


iamkun:
Looks cool THX

PR #123, opened 2018-05-09
Locale && Plugin

iamkun:
Waiting for reviews.
@Frondor @schiem 

The main purpose for this PR is to add flexibility to our core library including localization as well as plugin system. (Also a format plugin)

In order to keep our core code clear and small, I removed some of your code, but not much, I'll explain this later.

## Locale
```javascript
import dayjs from 'dayjs'
import esLocale from 'someLocale' // named 'es'
dayjs.locale(esLocale) // global set local
dayjs().locale(esLocale) // instance set local
dayjs().format(xxx, 'en') // use local in an API
```
- Why global locale
   The most common usage with locale in most real project might just be :  
  ```javascript
  import dayjs from 'dayjs'
  dayjs.local('fr')
  // then just use dayjs with French locale support.
  ```
  That's all. That's why I think it's useful to have a global locale setting.
- Why named locale
   I did not use the locale object directly. Instead, each locale object has a `name` key. So that we could set local through a string later without importing the object in every page in a SPA.
## Plugin
```javascript
import dayjs from 'dayjs'
import somePlugin from 'somePlugin'
dayjs.extend(somePlugin) // only one API
```
extend method will invoke a callback with three arguments, `prototype, class and factory`. Sine class.prototype might used in every plugin, so I just pass it in to the callback for convenience.

## Reason for removing some code

> remove plugin extend(xxx, true) to a new class factory

I removed the code about keep immutability. I think we may don't need this. I can't think of a situation that just need to include a plugin to a new instance locally rather than globally. 

> remove format config && toJSON 

Really I don't think we need this, sorry. Maybe we could add this as a plugin(we have plugin sys now!) to satisfy your and some others' need.


iamkun:
@Frondor Just back home, Thanks.

> You're no longer exporting Dayjs

I just getting an error from rollup "Named export and default export at the same time". So that I have to remove the export. Maybe there is a better solution.

> I'm not 100% sure but I think L and Ls

I don't understand the memory leak here. Could you detail it pls?

> what's the alternative solution to create a new dayjs with locale

Yes, My mistake. I may need to add a `.locale(this.$L)` before return it. And your approach about set a whole new Object is a little bit heavy. And I'll check other APIs to ensure return with locale.

> Also, plugin(Dayjs.prototype, dayjs, Dayjs) this line looks weird

Yes, I do think so. I'm still looking for a better solution.A brief Explanation: 
```javascript
// we could extend dayjs like
dayjs.utc() // moment has this method
// we could also extend Dayjs like
dayjs().newAPI()
// And Dayjs,prototype is just for convenience as I mentioned above in PR content.
```
Besides, you are right that plugin need to have a option argument.

> And still, don't see the point of having toJSON do the same as toISOString

After we finish locale and plugin. As for now, I just want to keep it simple and results same with moment. We may could discuss this in a separate issue. Since it's not the main problem in this PR(but still interesting). How do you think.

Frondor:
> Named export and default export at the same time
Weird, it worked for me, and the test passed. Maybe you could try
```js
// export class Dayjs -> tried it?
export { Dayjs };
export default dayjs;
// import { Dayjs }, dayjs from 'dayjs'
```

> I don't understand the memory leak here. Could you detail it pls?

Neither do I, but I think that if you import dayjs in the browser (or any module other importing it), and then do `dayjs = undefined`, the garbage collector is not gonna free any memory because there's a global reference to `Ls` in `$locale` method. I'm no expert here, so we may need to ask for help probably.

> And your approach about set a whole new Object is a little bit heavy

What's the heavy thing in setting `$d` to a given `Date` object? That's what the library does on every immutable action (and that's even heavier, since at cloning you're not only instantiating a new Date, but the whole library).
Actually, if you have a dayjs instance with certain locale configured, and you want to clone that instance (h**ence the locale and other options added by plugins**) with a different instance, you'll need to manually instantiate everything again and thus forcing the user to fail at DRY.
Anyway, I understand that's not in the scope of this PR either, but consider it.

In conclusion, everything else looks fine. I just don't like how the plugin system behaves actually but that's because of my very personal preference for the [Vue approach](https://github.com/xx45/dayjs/pull/92#issuecomment-386266780).

iamkun:
screen shot here, no idea what to do
```javascript
// code 
export  { Dayjs }
export default dayjs
```
rolloup:
screen shot 2018-05-10 at 9 50 13 am

Frondor:
> screen shot here, no idea what to do

That's just a warning... if then you can `import { Dayjs }, dayjs from 'dayjs` and do `dayjs(new Dayjs({ date: Date.now()}))` (just for testing purposes) That's it, the warning is just telling you're doing what you want.
This should be the final code
```js
export class Dayjs {...}
export default dayjs
```
> I like your vue approach, but we have a discussion [here], and wants to leave more flexibility to the plugin author.

That's the issue I linked previously, and both approaches add the same flexibility as far as I know. 
It looks more elegant to me but I think we should forgive that, you've already taken a decision and that's totally fine.
> Just added a wrapper function to deal with config inheritance

I'm not sure what does it fix regarding to the issues I've pointed out but :+1:

iamkun:
@Frondor the added wrapper: 
```
const wrapper = (date, instance) => dayjs(date, { locale: instance.$L })
```

so that we could easily deal with config inheritance. Just pass `wrapper(new Date(), this)` like this

src/index.js
```diff
@@ -1,29 +1,76 @@
 import * as C from './constant'
-import * as Utils from './utils'
+import * as U from './utils'
+import en from './locale/en'
 
-const parseConfig = (config) => {
+let L = 'en' // global locale
+const Ls = {} // global loaded locale
```
harrysarson: Every time someone adds a locale here it is permanently stored by dayjs, conceivably people could do

``` js
const process = (obj) => {
  // some logic here
  const locale = generate_locale(obj);
  const time = dayjs();

  return time.locale(custom_locale);
}
```
If this function is called many times and each time a unqiue local is generated (unlikely I know but possible) then dayjs will end up storing all the locales indefinitely.

This would be a good place to use a weak map: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/WeakMap

src/index.js
```diff
@@ -361,4 +427,11 @@ class Dayjs {
   }
 }
 
-export default config => (new Dayjs(config))
+dayjs.extend = (plugin, option) => {
+  plugin(option, Dayjs, dayjs)
+  return dayjs
+}
```
harrysarson: What advantage does using `dayjs.extend` give users over just adding to the dayjs prototype directly?

iamkun: We call this a plugin system. 😁 Just like jQuery.extend()

This just give the ability to extend our core library. Since we can't put everything into our core code.

src/index.js
iamkun: Nice point, but I don't really understand (sad face). A little more detail, please?

src/index.js
iamkun: You can see my comment [here](https://github.com/xx45/dayjs/pull/83#issuecomment-387805390)

We will have some Day.js extensions with official support. And with out `extend`, it will be so wired like this in our document.
```javascript
// if you want to use advancedFormat, follow the instruction
import { Dayjs } from 'dayjs'
import { advancedFormat } from 'dayjs/plugin'
Dayjs.prototype.advancedFormat = advancedFormat
dayjs.advanced = () => {......}
// that all
```
Or:
```javascript
dayjs.extend(advancedFormat)
```

iamkun:
I'm going to merge this pr tomorrow if no other opinions.

src/index.js
harrysarson: That makes sense :)

src/index.js
harrysarson: I will make a PR with a fix once this is merged, might be simpler to think about it as a seperate issue.

iamkun:
@kfirba after #141 is merged. Still some fixing and building work todo.

PR #130, opened 2018-05-11
Fix #127 Call isBefore, isAfter and isSame without argument

### LINKED ISSUE #127: Default to current time for isBefore/isAfter
In the momentjs API, calling `momentObj.isBefore()` (without an argument) returns whether the `momentObj` is before the current time.

Calling `dayjsObj.isBefore()` results in `TypeError: Cannot read property 'valueOf' of undefined`.

It'd be nice to have the same behavior as moment in this case, or at least note that it's different in the README. 🙂

Imperat:
Add ability to call isBefore, isSame and isAfter methods
without any argument.

It has some copy-paste of code. Any ideas how to make it better?


iamkun:
cool thx, I deal with this after finish #123

Imperat:
Hi, @iamkun. I rebase my fix to fresh version of library

iamkun:
Found a better solution instead of copy-paste

Issue #113, opened 2018-05-07
TODO


iamkun:
# Checklist
- [ ] CHANGELOG
- [ ] Chinese doc 中文文档补齐 目前format表格，core#day 文档缺失
- [ ] Plugin && Locale (doing...)

guisturdy:
#168

guisturdy:
i update the plugin Chinese doc in #168

Issue #63, opened 2018-04-26
Alternate constructor support?


planttheidea:
I've been traipsing through the documentation and the code, and didn't see anything, so if it's there and I missed it I apologize.

Is there a plan to support other constructors for different use cases, the way `moment.utc` and `moment.unix` operate? Everything currently appears to be local timezones only.

gpsimmo:
+1

zgayjjf:
Need this too.

iamkun:
dayjs.unix is available now

PR #133, opened 2018-05-11
unix-timestamp-seconds

guisturdy:
Passing an integer value representing the number of seconds since the Unix Epoch.

```js
dayjs.unix(1526048872)
dayjs.unix('1526048872')
```


src/index.js
```diff
@@ -361,4 +361,13 @@ class Dayjs {
   }
 }
 
-export default config => (new Dayjs(config))
+function DayjsEntry(config) {
+  return new Dayjs(config)
+}
+
+DayjsEntry.unix = (unixTimestamp) => {
```
Imperat: I belive it's better to implement `unix` method like static function
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Classes/static
instead of writing this wrapper.

README.md
```diff
@@ -143,6 +144,15 @@ dayjs(Number);
 dayjs(1318781876406);
 ```
 
+### Unix Timestamp (seconds)
+
+Passing an integer value representing the number of seconds since the Unix Epoch (Jan 1 1970 12AM UTC).
+
+```js
+dayjs(Number);
```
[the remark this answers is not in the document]
guisturdy: sorry for made this stupid mistake.

src/index.js
Imperat: I'm sorry. I think you are right! Apologize for this comment

iamkun:
Pending, really have to decide such feature should be treated as core feature or plugin.  Need your voice.

guisturdy:
this ( #63 ) issues is the reason i made this . 
now i thought make this feature as a plugin is better , few people may need this function .

iamkun:
Thanks. check #63 and #113

Since we have plugin system in `1.6.0`, we are planning to support `dayjs.unix` and `dayjs.utc` as two separate plugins. Not in core.

It would be nice if you could modify your code as a plugin, we are happy to merge it to our code base.

Cheers.

guisturdy:
i already start writing a plugin for implementation `dayjs.utc` , maybe you can let me done this task.

PR #140, opened 2018-05-14
Add Russian locale

alik0211:


iamkun:
@alik0211 would you please update your code to pass our unit test?

[Reference here](https://github.com/xx45/dayjs/blob/master/docs/en/I18n.md#customize)

iamkun:
Thanks

PR #148, opened 2018-05-15
Add Brazilian Portuguese locale

jgalat:
This adds the Brazilian Portuguese locale file. 
I used the current Spanish locale file as template.


src/locale/pt-br.js
```diff
@@ -0,0 +1,12 @@
+import dayjs from 'dayjs'
+
+const locale = {
+  name: 'pt-br',
+  weekdays: 'Domingo_Segunda-feira_Terça-feira_Quarta-feira_Quinta-feira_Sexta-feira_Sábado'.split('_'),
+  months: 'Janeiro_Fevereiro_Março_Abril_Maio_Junho_Julho_Agosto_Setembro_Outubro_Novembro_Dezembro'.split('_'),
+  ordinal: n => n
```
iamkun: should this be `ordinal : '%dº'` ?  Not sure.

iamkun:
Looks good thx

src/locale/pt-br.js
jgalat: @iamkun The ordinal can be used or not. Here you can check examples that show that both ways are used in Brazil:
https://en.wikipedia.org/wiki/Date_and_time_notation_in_Brazil
http://www.maujor.com/w3c/datetime.html Here search for _Exemplos_ (Examples)

iamkun: After checking locale files in other JS date libraries, seems most are `nº`. But still it's up to you, native speaker. 😁

jgalat: @iamkun I've amended the commit with the fix. 
Thanks for the review!

iamkun: great THX

PR #150, opened 2018-05-16
Fix typescript declarations

### LINKED ISSUE #139: typings indicate default whereas the js code by default is export all
in your ts typings:

```typescript
declare function dayjs (config?: ConfigType): Dayjs

export default dayjs
```
in your distjs.min.js
```javascript
...
 module.exports = e() 
...
```

styfle:
Fixes #139

The problem was that users cannot actually run ESM today so there is no "default" export.

This follows the pattern of [tape](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/tape/index.d.ts) which is to put all types in a namespace and export the function.

### Usage

Before, the typings expected the user to import the default.

After, the typings expect the user to import all exports which is a single function.

```ts
import dayjs from 'dayjs'; // usage before this PR
import * as dayjs from 'dayjs'; // usage after this PR
```

### TS Output (JS)

```js
var dayjs = require('dayjs');
```


iamkun:
Thanks @styfle, looks better. 

I've added two more APIs in @1.6.0 release. And working on the docs, could you please help me add it to `.d.ts` file as well if you have time? I don't know where to add 😬 

### API one: plugin install `extend`
```
dayjs.extend(plugin: Function, option: Any (null string object ...))
```

### API two: i18n support `locale`
```
// 1 dayjs constructor accept two arguments
dayjs(date: ConfigType, config: optional object)
// e.g. dayjs('2018-4-28', { locale: es }


// 2 set global locale
dayjs.locale(arg1, arg2) 
// 3 combinations here
// arg1 string , arg2 null
// arg1 string, arg2 object
// arg1 object, arg2 null

// 3 set instance's locale
dayjs().locale(any, any) // the same argument as global
```

Many thanks. 🙏

iamkun:
@styfle Thanks a lot. I'll merge this to next release.

Issue #135, opened 2018-05-12
Improvements on the introduction.


AlvinTheDeveloper:
For the introduction in README.md I think we should mention what day.js can do. Shouldn't wait for the user scroll to the API reference and spend some time to understand what is going on.

I think we should add some introduction about the functions right before the sentence below:

> Day.js is a minimalist JavaScript library for modern browsers with a largely Moment.js-compatible API. If you use Moment.js, you already know how to use Day.js.

iamkun:
Umm, maybe a website is needed at this moment.

Or any suggestion to our readme structure?

dmckenna24:
I agree with @iamkun, I've previously worked on moment, and having the website with all required documentation and introductory information was useful before entering their GitHub. Perhaps create an issue on this? Or is it too big of a scale for just a regular issue?

iamkun:
@dmckenna24 agreed!

iamkun:
@AlvinTheDeveloper THX, that make sense.

iamkun:
@AlvinTheDeveloper added to readme, better?

AlvinTheDeveloper:
Will do. I was busying. Sorry about that.

AlvinTheDeveloper:
Sorry I have just found some grammatical mistakes in my sentence.
This should be better:

> Day.js is a minimalist JavaScript library that parses, validates, manipulates, and displays dates and times for modern browsers with a largely Moment.js-compatible API. If you use Moment.js, you already know how to use Day.js.

iamkun:
@AlvinTheDeveloper THX, I'll update it to readme.

iamkun:
Doc updated!

PR #151, opened 2018-05-16
Improved README.md introduction.

AlvinTheDeveloper:
Added few words as the issue mentioned. #135


iamkun:
@AlvinTheDeveloper What I mean in that issue is, I've added your words to readme already. Just check if it looks good?

PR #167, opened 2018-05-18
Add Japanese (ja) locale

ohbarye:
I added `ja` locale for Japanese.
Please review when you have time!

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -13,8 +13,8 @@ package-lock.json
 coverage
 
 # build
-locale
-plugin
+locale/
+plugin/
 dayjs.min.js
 
 #dev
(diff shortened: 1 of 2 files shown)


.gitignore
```diff
@@ -13,9 +13,9 @@ package-lock.json
 coverage
 
 # build
-locale
-plugin
+locale/
+plugin/
```
iamkun: what's this change for?

ohbarye: @iamkun 
To include `src/locale` & `src/plugin` folder. 

I found that changes under `src/locale` are ignored when the `.gitignore` has `locale`. But once it changes to `locale`, it just ignores `locale` and stopped to ignore `src/locale`.

iamkun: Sure, got it thanks.

Issue #143, opened 2018-05-14
parsing difference with time offset between moment and dayjs


tareqlol:
Hello, firstly thanks for your effort for this plugin. 
I found an issue with parsing the same date between moment and dayjs 
![image](https://user-images.githubusercontent.com/7700188/40005918-6a760198-5799-11e8-9133-16339eb4d569.png)
as you can see here dayjs is not adding the timezone shift when you call format method.
am i doing something wrong ?

iamkun:
Thanks, should be a bug. Will check it soon

tareqlol:
I have checked other related issues mentioning that replacing the dashes and "**T**" will solve it for now but I guess it should be fixed [!73](https://github.com/xx45/dayjs/issues/73)
`.replace(/-/g,'/').replace('T', ' ')`
![image](https://user-images.githubusercontent.com/7700188/40007342-c9a5e2de-579c-11e8-894b-864e881bf432.png)
Just wanted to reference it in case it helps. Thanks :)

Imperat:
Hi, @iamkun. I'm going to update regular expression and fix the function `parseDate`. But I misunderstand why we need to parse the string in ISO 8601 format in this place:
```
  if ((typeof date === 'string') && (reg = date.match(C.REGEX_PARSE))) {
    // 2018-08-08 or 20180808
    return new Date(
      reg[1], reg[2] - 1, reg[3] || 1,
      reg[5] || 0, reg[6] || 0, reg[7] || 0, reg[8] || 0
    )
  }
```
instead to pass this string into constructor directly:
```
if (typeof date === 'string') {
  return new Date(date);
}
```
Can you clarify it, please?

iamkun:
The regex is used to parse date string like 2018-01-01, 2018-01 that is not in ISO format.

But in the case, this date string passed our regex, which is incorrect.

Imperat:
Thank you, @iamkun. Just another question.
The last matching group from regular expression:
`.?(\d{1,3})`
It means amount of milliseconds?
We pass it as last argument to new Date() constructor.
I'm afraid that we have the bug in this place also.

Imperat:
For example:
12:20:12.1 means 1/10 part of second - 100 milliseconds.
12:20:12.10 means 10/100 (1/10 also )
12:20:12.123 means 123 milliseconds
But I see that we take this value from string and pass it as milliseconds.
Please, let me know if I'm wrong.

iamkun:
[...]

PR #168, opened 2018-05-18
Add UTC plugin

guisturdy:
```javascript
import utcPlugin from 'dayjs/plugin/utc'
dayjs.extend(utcPlugin)
// dayjs.extend(utcPlugin, { parseToLocal: Boolean })
```

### capability

```javascript

dayjs('2018-05-18T03:04:05+06:00')  // creating with utc

dayjs.utc()
dayjs.utc('2018-05-18T03:04:05+06:00') // (string | number | Date | Dayjs) support like dayjs()

dayjs().utc()
dayjs().local()

dayjs().isLocal()
dayjs().isUTC()

dayjs().utcOffset()   // get
dayjs().utcOffset(0)  // set

```

### TODO

- [x] doc
- [x] test cover


src/plugin/utc/util.js
```diff
@@ -0,0 +1,26 @@
+function timezoneOffsetToString(tz) {
+  if (tz === 0) return 'Z'
+  const h = Math.floor(Math.abs(tz) / 60)
+  const m = Math.abs(tz) % 60
+  return `${tz > 0 ? '-' : '+'}${h < 10 ? '0' : ''}${h}${m < 10 ? '0' : ''}${m}`
```
Imperat: Comment about ${h < 10 ? '0' : ''}${h}

Probably function padStart from utils of core code is better? :)
${padStart(h, 2, '0')}

guisturdy: thanks, i replaced this method with `padZoneStr` from core code.

Imperat:
In common I see that in this new plugin there is ability to parse timezone offset:
+06:00 or Z or -0530. 
According discussion in issue https://github.com/iamkun/dayjs/issues/143 do we want to add ability to parse timezone offset in core part of library? Just for have ability to parse next strings and get same results?
2018-05-18T03:04:05+06:00
2018-05-18T04:04:05+07:00

guisturdy:
the core part already had the bility to parse timezone offset right now
```javascript
dayjs('2018-05-18T03:04:05+06:00').valueOf() // 1526591045000
dayjs('2018-05-18T04:04:05+07:00').valueOf() // 1526591045000
```
the difference after load this plugin is get data method (like `.year()` `.date()` `. format()`) will return the data based of the time zone what you set not the time zone of you local 
- not load plugin
```javascript
dayjs('2018-05-18T03:04:05+06:00').format() // 2018-05-18T05:04:05+08:00
dayjs('2018-05-18T04:04:05+07:00').format() // 2018-05-18T05:04:05+08:00
```
- load plugin
```javascript
dayjs.extend(utcPlugin)
dayjs('2018-05-18T03:04:05+06:00').format() // 2018-05-18T03:04:05+06:00
dayjs('2018-05-18T04:04:05+07:00').format() // 2018-05-18T04:04:05+07:00
```

Imperat:
Got it. Thank you for clarification, @guisturdy!

guisturdy:
#113

iamkun:
Fantastic. Please allow me some time going through the code.

Plus, nice translation. Would you mind open a new PR with the translation of current Chinese Document only? So that I could merge it in advance.

guisturdy:
@iamkun remove the part no belong to this plugin from Plugin.md and as an new PR #173

iamkun:
@guisturdy  cool

iamkun:
@guisturdy In order to keep the compiled file small and simple, we should keep away from `Array. reduce`  `Object.assign` and `{...}`

guisturdy:
ok , i'll change this code to ES5 style

在 2018年5月22日，18:41，iamkun <notifications@github.com<mailto:notifications@github.com>> 写道：


@guisturdy<https://github.com/guisturdy> In order to keep the compiled file small and simple, we should keep away from Array. reduce Object.assign and {...}

iamkun:
@guisturdy Thanks, ES6 style is OK and welcome, just not those modern APIs that are not fully supported by browsers. Those polyfills are too heavy to our project.

guisturdy:
@iamkun finished

hlehmann:
any update on this ?

iamkun:
@hlehmann This is an import PR for our project. I'm trying to find some time go through this lot of code. And reviews are welcome.

antony:
Would be great to have this in soon. Trying to use dayjs on a new project and `.startOf` is giving me yesterday at 11pm!

RJiryes:
When will this be available? Thanks!

guisturdy:
before this PR merged, hope [_**this**_](https://github.com/guisturdy/dayjs-plugin-utc) may help you
@RJiryes @antony

prantlf:
Essentially, tis plugin retains the original time zone (offset) coming from the initialisation string. The date parts (year, month, ...) used in later formatting and other operation retain their original values. Am I right?

I am wondering, what would be a real-world use case for this. I mean no disrespect :-) Could you help me by explaining a bit, in what application and how you will use it?

Applications, which I am working on, need to display the time zone consistently, not to confuse the user. The current time zone parsing works well. Different date sources, potentially with different time zones in their strings, are parsed to standard `Date` object. It offers them in a single time zone - the local one - plus UTC support.

Some applications just work in the local time zone, because it is the most natural  for the users. No work needed, the `Date` object in Day.js supports it well.

Other applications let the users specify the time zone, which they wish to see and enter the date values. The `Date` object in Day.js does not support it. But this plugin does not support it either.

I opened a feature request for the full time zone support: #323.

prantlf:
I browsed opened issues to help me understand the need for the "UTC mode". I found #74 and #249 particularly interesting.

It appears, that there is a misunderstanding about how Day.sj constructs `dayjs` instances. I have no problem, because I am aware, that Day.js behaves like the native `Date` - it assumes the local time zone, if the time part is omitted in the input string:

```js
dayjs("2018-10-28")
new Date("2018-10-28")
```

Both of the expressions above initialise the date instances to the following value in Central Europe: "2018-10-28T01:00+01:00". However, it is not obvious and people may think, that Day.js works in a "date-only mode", if they omit the the time part. If you know `Date` well, Day.js will not surprise you :-) But Day.js was created to make the the date handling easier. We could do better here.

This can be solved by two ways:

* Introduce a "date-only mode" in Day.js. Specify it during the construction, for example.
* Introduce an "UTC mode" in Day.js. The time will be always set to 00:00 and no conversion to the local time zone will be performed.

Although I feel, that the former would be clearer and better fitting for the use case, I admit, that the latter can be used to work with UTC dates as well and thus is more flexible.

When I was looking at this PR, I could not help myself not thinking, why would I need a new `UTCDate` class? The native `Date` stores the UTC value too and offers methods to inspect and manipulate it too. Is it not possible to stay with the `Date` and make the change simpler? Day.js stores the date and time parts in private variables $Y, $M etc. anyway. Most of the `dayjs` methods do not need to be modified.

I opened a new PR #326 with a simpler proposal. Not as a plugin, at first, because overriding `wrapper` and `parseDate` would mean copying practically all code to the plugin. The change is small, but similar to the introduction of the locale parameter.

prantlf:
I suggest not to change behaviour of the constructor. If somebody includes the UTC plugin in an application, existing components already using `Day.js` should not get broken. I suggets making the "UTC mode" obvious by some parameter like `{ parseToLocal: Boolean }` or `{ assumeUTC: true }`.

Also, the `utcOffset` method should return the same sign, as `Date.prototype.getTimezoneOffset` for consistency. The method `month` returns also the month index instead of the month number to be consistent with `Date`.

iamkun:
After careful consideration, we shouldn't change the behaviour of the original constructor. Thus, a plugin with `dayjs.utc` and `dayjs().utc()` is preferred and will be available soon.

PR #179, opened 2018-05-20
fix bug: advencedFormat plugin

huruji:
**When I run `npm run test`**：
image

The two letters of  an ordinal number after the number are usually **the last two letters in the word denoting that number**. So it's a bug, and I fixed it.

After fixed:
image

### DIFF
--- a/src/plugin/advancedFormat.js
+++ b/src/plugin/advancedFormat.js
@@ -4,7 +4,7 @@ export default (o, c, d) => { // locale needed later
   const proto = c.prototype
   const oldFormat = proto.format
   // eslint-disable-next-line no-nested-ternary
-  d.en.ordinal = number => `${number}[${number === 1 ? 'st' : number === 2 ? 'nd' : number === 3 ? 'rd' : 'th'}]`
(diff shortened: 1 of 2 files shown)


Imperat:
Thank you, @huruji!
But I think it still has some bugs. For example look at "eleventh" number. Your code will return 11st, but correct result is 11th. Probably we can improve it also?:)

Imperat:
For example it can be fixed like that:
```
number => {
  let suffix;
  if ([11, 12].includes(number)) {
    suffix = 'th';
  } else {
    suffix = ['st', 'nd', 'rd'][number % 10] || suffix;
  }

  return `${number}[${suffix}]`;
}
```
Every time it wonderful to make code more elegant and simple, but I'm not sure how to write it in less amount of lines.

huruji:
Oh,great, thank you, @Imperat your code is very elegant

huruji:
I fixed the bug using your code and add some test, it's OK. Very thank you!! @Imperat

iamkun:
My mistake, sweet fix, thanks @huruji , @Imperat

Issue #175, opened 2018-05-20
API reference improvements


vilanz:
While the API reference right now works well, I think it could be more concise, especially when it comes to method typing and some inconsistent grammar/formatting.

I can't set up a pull request right now, but these are some sketches of what I came up with (still need to find a better way to include parameter types):

#### Now `Dayjs`
Set to the current date and time by default.

```js
dayjs();
```

#### Dayjs parsed from other date formats `Dayjs`

* [ISO 8601](https://en.wikipedia.org/wiki/ISO_8601) string

```js
dayjs(String);
dayjs('1995-12-25');
```

* Unix Timestamp (milliseconds since the Unix Epoch - Jan 1 1970 12AM UTC)

```js
dayjs(Number);
dayjs(1318781876406);
```

* native Javascript Date object

```js
dayjs(Date);
dayjs(new Date(2018, 8, 18));
```

[ ... stuff ... ]

## Get + Set

Get and set date.

#### Year `number`

```js
dayjs().year();
```

#### Month `number`

```js
dayjs().month();
```

PR #180, opened 2018-05-20
Reformatting the API Reference

ludvon:
Added some type & parameters information, examples, and changed the Markdown formatting for the subsections. #175 

I kept the parameters info in the table of contents. I don't mind it, but it might look too long.


iamkun:
@ludvon  Thanks, and looks better I think.

PR #182, opened 2018-05-21
immutable locale, docs update

iamkun:


iamkun:
Hi, @ohbarye @ludvon 

In order to keep the same behavior with other APIs, I've changed `dayjs().locale` to a immutable action.

And also updated one line of the English doc like this:

```md
Changing locales locally
- Returns a new Dayjs object by switching to new locale.
```

For less confusion, would you please update your Docs if you have some free time? 

Cheers.

ludvon:
@iamkun Great! The wording in my translation already implied it was immutable, so I don't need to change it.

iamkun:
@ludvon great!

PR #186, opened 2018-05-21
update: prevent rewriting and reduce bytes

a13821190779:


Imperat:
Who of us is familiar with eslint? It's needed to disable no-void rule in this line)

iamkun:
sad story from eslint 

> Expected 'undefined' and instead saw 'void'  no-void

a13821190779:
Vscode does not prompt...

iamkun:
Add this above at the end of the line could solve this:
```
// eslint-disable-line no-void
```

a13821190779:
i repr

Issue #185, opened 2018-05-21
 ideas about saving some bytes 


FoxDaxian:
https://github.com/iamkun/dayjs/blob/97763f83fc84ea86ca8a3adb0ada467127a4eb0e/src/utils.js#L27

iamkun:
Sounds interesting, any detailed information please?

Imperat:
Good idea to write it in less amount of symbols:
```
-const isUndefined = s => s === undefined
+const isUndefined = s => s === void 0
```

iamkun:
yes, it's shorter than undefined and save a little bytes 😁

a13821190779:
I thought you probably got it, so. . .

Imperat:
I had some ideas about saving some bytes.. example:
-return Utils.padStart(this.$H < 13 ? this.$H : this.$H - 12, match === 'hh' ? 2 : 1, '0')
+return Utils.padStart(this.$H < 13 ? this.$H : this.$H - 12, match.length, '0')
Let me investigate it when I will have free time ;)

Imperat:
@a13821190779 but we need to merge it, of course. Can you open PR?)

a13821190779:
ok @Imperat 
and I will continue to pay attention and make good suggestions. :)

iamkun:
This is an interesting topic and I've change the title a little bit.

huruji:
oh, a great idea. @a13821190779

a13821190779:
@huruji 哈哈

guisturdy:
rollup-plugin-uglify plugin will help us do this , an we already using this in `build/rollup.config.js` .

iamkun:
@guisturdy Just noticed your comment after I merge that PR.

waitingsong:
undefined is not save, and can be modified .
void 0 alwalys equal to the undifined  : `assert(typeof (void 0) === 'undeifned) `

iamkun:
great discussion

PR #187, opened 2018-05-21
update: prevent rewriting and reduce bytes

a13821190779:


Imperat:
Sorry, I forget about this:
https://babeljs.io/docs/plugins/transform-undefined-to-void/
Probably it's better to add this plugin to our babelrc config instead writing "eslint"-wrong code?

Imperat:
Not sure about right solution - we use "undefined" word only in this one line and it seems ugly to add separate babel plugin only for modify just one line..

iamkun:
> seems ugly to add separate babel plugin only for modify just one line..
Nice plugin, but maybe just do it manually this time.

a13821190779:
Haven't merged yet?

guisturdy:
as i said in #185 , `rollup-plugin-uglify` plugin we used in `build/rollup.config.js` is already help us do this, change code here would not saving any byte.

iamkun:
🎉 This PR is included in version 1.6.4 🎉@a13821190779

PR #189, opened 2018-05-22
Update fromNow plugin

iamkun:
@aalises 

I update the code of `fromNow` plugin a little bit. Here is the reason:

In previous code, we use `resabs !== P[i].v` to check if it is singular `a month` or `2 months`.

However, there is some disadvantage of this kind of check, it must be exactly the same time in millisecond, otherwise, it will result ` 1 month ` instead of `a month` 

So my solution is, still implement it in Moment.js's way. 

We set the output to a range of time. Check the table blew:

Range | Key | Sample Output
-- | -- | --
0 to 44 seconds | s | a few seconds ago
44 to 44 seconds | ss | 44 seconds ago
45 to 89 seconds | min | a minute ago
90 seconds to 44 minutes | mins | 2 minutes ago ... 44 minutes ago
45 to 89 minutes | h | an hour ago
90 minutes to 21 hours | hh | 2 hours ago ... 21 hours ago
22 to 35 hours | d | a day ago
36 hours to 25 days | dd | 2 days ago ... 25 days ago
26 to 45 days | M | a month ago
46 days to 10 months | MM | 2 months ago ... 10 months ago
11 months to 17months  | y | a year ago
18months+ | yy | 2 years ago ... 20 years ago

Also, added `from` `to` `toNow` APIs.

Should we come up a new name for this plugin, since it's not only `fromNow`  now


aalises:
Hi! Thank you very much for your work! It indeed is better that way. When I designed the plugin I had timeago.js in mind and a very basic approach to it, not as complex as moment.js

iamkun:
Thanks @aalises.

Any review here?

PR #195, opened 2018-05-24
fixed `toObject` return type

ingver:
Just noticed that `toObject` return type was too generic


iamkun:
THX

iamkun:
🎉 This PR is included in version 1.6.4 🎉

PR #220, opened 2018-06-01
Fix missing colon in locale template

mahammedzkhan:
Fix missing colon in locale template


iamkun:
@mahammedzkhan Thanks, I am about to update it.

Issue #223, opened 2018-06-01
ss [plural for seconds]


diomed:
I don't see string for plural of seconds anywhere in translations,
and I'm pretty sure if there's d, dd and m, mm, and h, hh, there 
should also be s, **ss**

iamkun:
Yes, we have this.



s | 0-59 | The second
-- | -- | --
ss | 00-59 | The second, 2-digits
SSS | 000-999 | The millisecond, 3-digits

diomed:
it isn't in any translate file however.
it should be, right?

aalises:
@diomed @iamkun we don't need the 'ss' string as our RelativeTime plugin does not use it. As differentiating from 1 second to 9 seconds is quite specific, we just provide the 'few seconds' generalization for this case. See how the locales are written.

iamkun:
@aalises I skipped the `ss` string in our RelativeTime plugin just because I want to keep the same behavior as moment.js.  However, if we think it necessary, we could add it back.

aalises:
@iamkun I don't think it's necessary to be fair :+1:

prantlf:
@aalises, using just `s` with "a couple of seconds" is fine. I agree, that there's no need to make it more complicated to "count the seconds".

But there is a problem with using your localization functionality in real-world applications, which are not in English. Assuming, that "*in* a couple of seconds" and "a couple of seconds *ago*" uses the same common text "a couple of seconds" is wrong. Other languages need to apply declension specific to preposition used. I opened #302 about it.

iamkun:
discussion 302

PR #225, opened 2018-06-04
Update hr.js

diomed:
ss seconds added in translation


iamkun:
@diomed Hi, there might be a little misleading in #223.

`ss` is supported in `.format()` API for 2-digits second. However, the `ss` in `relativeTime ` in this pr is not necessary, since it's is not used by our `relativeTime`  plugin.

PR #226, opened 2018-06-05
Add Catalan (ca) and French (fr) locales

naulacambra:
Added Catalan (ca) and French (fr) locales


src/locale/fr.js
```diff
@@ -0,0 +1,27 @@
+import dayjs from 'dayjs'
+
+const locale = {
+  name: 'es',
```
iamkun: `fr` here?

PR #228, opened 2018-06-06
update format API

iamkun:


src/locale/th.js
```diff
@@ -3,7 +3,7 @@ import dayjs from 'dayjs'
 const locale = {
   name: 'th',
   weekdays: 'อาทิตย์_จันทร์_อังคาร_พุธ_พฤหัสบดี_ศุกร์_เสาร์'.split('_'),
-  weekdaysShort: 'อา._จ._อ._พ._พฤ._ศ._ส.'.split('_'),
+  weekdaysShort: 'อาทิตย์_จันทร์_อังคาร_พุธ_พฤหัส_ศุกร์_เสาร์'.split('_'),
```
iamkun: @saknarak
In order to pass the unit test compared to moment.js, I've changed this locale a little bit. Is this still OK in thai?

saknarak: Yes, it still OK in Thai.

```
// from moment.js
        weekdaysShort : 'อาทิตย์_จันทร์_อังคาร_พุธ_พฤหัส_ศุกร์_เสาร์'.split('_'),
        weekdaysMin : 'อา._จ._อ._พ._พฤ._ศ._ส.'.split('_'),
```

iamkun: Thanks @saknarak , just wondering if we should add weekdaysMin as well?

PR #229, opened 2018-06-06
fix 'locale' method name in index.d.ts typings file

rubenbp:
There is an error in the typings file with the 'locale' method of the namespace 'dayjs'
This pull request solves the problem of the name


iamkun:
Cool thanks.

PR #230, opened 2018-06-06
buddhistEra plugin

saknarak:
- BBBB for B.E. year
- BB for B.E. year 2 digit

B.E. is 543 year before C.E.


iamkun:
@saknarak Thanks, interesting plugin.

Since buddhistEra is a fresh concept to most of us. Would you please add a simple docs introducing your plugin?  

`master/docs/en/Plugin.md` is a good place to add.

Cheers.

src/plugin/buddhistEra/index.js
```diff
@@ -0,0 +1,21 @@
...
+  const oldFormat = proto.format
+  // extend en locale here
+  proto.format = function (formatStr, localeObject) {
+    const yearBias = 543
+    const locale = localeObject || this.$locale()
+    const str = formatStr || FORMAT_DEFAULT
+    const result = str.replace(/BBBB|BB/g, (match) => {
+      switch (match) {
+        case 'BB':
+          return (this.$y + yearBias) % 100
```
iamkun: @saknarak we should check and add the leading zero using `padStart`, consider this case `2000 % 100`

src/plugin/buddhistEra/index.js
```diff
@@ -6,14 +6,15 @@ export default (o, c) => { // locale needed later
...
     const locale = localeObject || this.$locale()
     const str = formatStr || FORMAT_DEFAULT
     const result = str.replace(/BBBB|BB/g, (match) => {
       switch (match) {
         case 'BB':
-          return (this.$y + yearBias) % 100
+          return utils.padStart(String((this.$y + yearBias) % 100), 2, '0')
         default: // BBBB
-          return this.$y + yearBias
+          return utils.padStart(String(this.$y + yearBias), 4, '0')
```
iamkun: @saknarak we don't need this in four-digits year right? e.g. year 998 should be 998 not 0998

and also come up with a new idea, `String(this.$y + yearBias).slice(-2)` seems easier to understand.

saknarak: base on moment
console.log(moment('0123-05-07').format('DD MMM YYYY')) // 07 MAY 0123
console.log(moment('123-05-07').format('DD MMM YYYY')) // error

iamkun: I see

PR #231, opened 2018-06-06
Add Taiwan (zh-tw) locale

LeoLiu1209:
add in src/locale/zh-tw.js

### DIFF
--- a/src/locale/zh-tw.js
+++ b/src/locale/zh-tw.js
@@ -1,0 +1,30 @@
+import dayjs from 'dayjs'
+
+const locale = {
+  name: 'zh-tw',
+  weekdays: '星期日_星期一_星期二_星期三_星期四_星期五_星期六'.split('_'),
+  weekdaysShort: '週日_週一_週二_週三_週四_週五_週六'.split('_'),
+  weekdaysMin: '日_一_二_三_四_五_六'.split('_'),
+  months: '一月_二月_三月_四月_五月_六月_七月_八月_九月_十月_十一月_十二月'.split('_'),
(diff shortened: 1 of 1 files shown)


src/locale/zh-tw.js
```diff
@@ -0,0 +1,30 @@
+import dayjs from 'dayjs'
+
+const locale = {
+  name: 'zh-tw',
+  weekdays: '星期日_星期一_星期二_星期三_星期四_星期五_星期六'.split('_'),
+  weekdaysShort: '周日_周一_周二_周三_周四_周五_周六'.split('_'),
```
iamkun: Not sure if this should be `週日_週一_週二_週三_週四_週五_週六` ?

LeoLiu1209: I think it should be ok!!
After i translated for it, I have been to Moment.js locale to check for [zh-tw ](https://github.com/moment/moment/blob/develop/src/locale/zh-tw.js)and [zh-cn](https://github.com/moment/moment/blob/develop/src/locale/zh-cn.js).
The translate are the same :)

iamkun: But the `zh-tw` link you provided is `週日` not `周日`?

LeoLiu1209: oops is my mistake!!!!
i will fix for it!!!!!

LeoLiu1209:
I merge multiple commits into single one.
If the PR need to be revised or corrected, i will try my best to make it more completely :blush:

iamkun:
Thanks @LeoLiu1209, LGTM.

PR #250, opened 2018-06-22
Added Norwegian Bokmål "nb" locale.

cowai:


iamkun:
@cowai  thanks cool

cowai:
@iamkun Wow that was fast!
Looking forward for this to be available in a release.

iamkun:
@cowai Will be available in next release.

Plus, according to other libs, weekdaysShort in Norwegian "nb" locale is not the first three letters.  Any update here?

cowai:
@iamkun Thanks.
Both first 2 and 3 works in Norwegian.
I wondered what I should take but ended up doing the same as Swedish and Danish (90% same language) is in this project: https://github.com/iamkun/dayjs/blob/master/src/locale/da.js and  https://github.com/iamkun/dayjs/blob/master/src/locale/sv.js

cowai:
@iamkun Oh, I misunderstood you at first.
Yes, all weekdays in Norwegian can absolutely be cut to use only first 2 and 3 letters.

iamkun:
@cowai It's up to you, native speaker 😬.

PR #259, opened 2018-06-26
feat: Added method to retrieve week of the year

naulacambra:
Added API method `week()` to retrieve week of the year

### DIFF
--- a/docs/en/Plugin.md
+++ b/docs/en/Plugin.md
@@ -144,6 +144,17 @@ List of added formats:
 | `BBBB` | 2561             | Full BE Year (Year + 543)             |
 | `BB`   | 61               | 2-digit of BE Year                    |
 
+### WeekOfYear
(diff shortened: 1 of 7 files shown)


iamkun:
@naulacambra Nice code, thanks. How about added this as a plugin `weekOfYear`?

naulacambra:
Since moment.js has this method, and I personally think is quiet usefull, I've not thought adding it as a plugin.

Right now, the package has all `get` methods, but `week`, and it kinda feels right to add it here.

What we could win adding it as a plugin?

iamkun:
The reason why I'm not adding this to the core code at the first beginning is because it's not that widely used compared to other APIs.

However, as a useful API, we should add to our lib. That's why I think a plugin might be a better choice.

iamkun:
@naulacambra No problem 😁

naulacambra:
Moved code from core to weekOfYear plugin

naulacambra:
Updated Plugin.md doc file (english version)

Issue #269, opened 2018-07-06
Not able to set the day of the week


bas080:
Moment has this awesome feature where you can set the day of the week.

https://momentjs.com/docs/#/get-set/day/

PR #270, opened 2018-07-06
Add setting the day of the week

bas080:


iamkun:
@bas080 Great, thanks, I missed this. Plus, would you mind change this commit to a bugfix: instead of feat: cause a feat: would trigger a minor update

bas080:
Happy to help. I needed the feature too.

Changed the commit message.

#269

PR #272, opened 2018-07-06
Extend the property based unit tests

bas080:
- add a seed value which changes everytime the tests succeed.
- use pseudrandom functions for generating test data,

This PR contains tests for #270.

Issue: #272

### DIFF
--- a/test/get-set.test.js
+++ b/test/get-set.test.js
@@ -1,9 +1,15 @@
 import moment from 'moment'
 import MockDate from 'mockdate'
+import { gen } from 'testcheck'
+import jasmineCheck from 'jasmine-check'
 import dayjs from '../src'
 
+jasmineCheck.install()
+
+const seed = Number(process.env.TEST_SEED)
+
 beforeEach(() => {
-  MockDate.set(new Date())
+  MockDate.set(seed)
 })
 
 afterEach(() => {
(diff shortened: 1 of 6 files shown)


bas080:
The date `new Date` is not deterministic yet. That still requires a solution.

https://github.com/iamkun/dayjs/pull/272/files#diff-df4745a4855110035c597b7240e40042R9

bas080:
That should fix the date. It is now deterministic.

So I ran the tests for a short while. I found a seed that results in a difference between moment's and dayjs' month set behavior.

```
  ● Set Month ( 1 )

    expect(received).toBe(expected) // Object.is equality
    
    Expected value to be:
      956306927487128
    Received:
      956307013887128
```

You can run the tests yourself by running the following from your bash cli.

```bash
TEST_SEED='956330601087128' jest ./test/get-set.test.js
```

test/get-set.test.js
```diff
@@ -42,36 +48,25 @@ it('Millisecond', () => {
...
+  ['Month', gen.intWithin(0, 12)],
+  ['Year', gen.intWithin(-5000, 5000)],
+  ['Hour', gen.intWithin(0, 24)],
+  ['Minute', gen.intWithin(0, 59)],
+  ['Second', gen.intWithin(0, 59)],
+  ['Millisecond', gen.intWithin(0, 999)]
+]
+
+unitsAndQuantities.forEach(([unit, genInt]) => {
+  check.it(`Set ${unit}`, { seed: Number(seed), numTests: 1000 }, genInt, (quantity) => {
```
bas080: 1000 might be a bit on the high side. It defaults to 100. Less is also fine.

iamkun:
@bas080  `*.test.js` are unit tests based on jest, while `*.spec.js` are browser tests based on karma and saucelab. However, we have to write the same test logic twice because of the different test framework, and I'm looking for a way to simplify it.

bas080:
@iamkun, this PR is ready for a review. Note that the tests found a difference between moment's and dayjs' set behavior.

iamkun:
I'll close this PR since it's been a while since it's been opened. Feel free to reopen if you have updates on this

bas080:
@iamkun, would you like me to rebase, fix things if necessary and propose a review? It might detect more differences in behavior between momentjs and dayjs.

iamkun:
Sorry for the late reply, to be honest, I do not really know what's this PR is used for. Can you please detail it?

bas080:
@iamkun, the assumption I'm making is that moment.js and dayjs have the same behavior. I basically test if methods of moment.js output the same or something similar to day.js using a reproducible testdata-set for each test-run.

iamkun:
Yes, as far as I can see, all the tests in dayjs are compared to moment.js' result. Why should we add this additional test?

bas080:
Because it implements "generative testing", resulting in testing many more cases in a reproducible manner.

> https://en.wikipedia.org/wiki/Test_generation

iamkun:
I see. One question here does "generative testing" take much more time than the current test, or do we have to test them both?

bas080:
@iamkun, The generative testing with a random seed does not have to take much time. This depends on how much test data is generated for a test run.

The amount of test data can be configured. [`jasmine-check`][1] comes with options to configure this. Jasmine check uses [`testcheck-js`][2].

The reporter I wrote can be simplified. No need for a `.seed` file. Only an environment variable should be enough. That removes the tedious `fs` code. I'll focus on that. 

ALSO, It appears that the jasmine check has moved into the testcheck-js project. It has become an "integration" I would have to do some updating.

Just to take away your concern. The tests do not have to take more time. See https://github.com/leebyron/testcheck-js/tree/master/integrations/jasmine-check#options

I guess I have to do more work before this PR is ready for a merge.

[1]:https://github.com/leebyron/testcheck-js/tree/master/integrations/jasmine-check#options
[2]:https://github.com/leebyron/testcheck-js

iamkun:
Thanks, I finally know what this test is. Just wondering do we really need this type of test?

Issue #274, opened 2018-07-09
bug(locale): Dutch months and weekdays should be lowercase


donroyco:
## Type
Bug

## Where?
The Dutch locale 🇳🇱:
```dayjs/blob/master/src/locale/nl.js```

## What happened?
The Dutch language doesn't have capitalised months (`januari`, `februari`, `maart` etc.) and weekdays (`maandag`, `dinsdag`, `woensdag`, etc.), like the English language has (`January, February`, `March`, etcetera and `Monday`, `Tuesday`, `Wednesday`, etcetera).

Explanation from 'Our Language':

> But in the Green Book from 1954 the names of months and the days of the week got a lowercase letter, because they are not unique names, and that has remained the same since then.  _[Source](https://onzetaal.nl/taaladvies/oktober-october/)_

Reference to Moment `nl` locale: [moment/nl.js](https://github.com/moment/moment/blob/develop/locale/nl.js)

## And now?
I will open a PR soon to solve this bug. Cheers.

iamkun:
Great fix, thanks @donroyco

iamkun:
fix in 1.7.3

PR #275, opened 2018-07-09
fix(locale-nl): set correct weekdays and months

donroyco:
This will fix issue #274


bas080:
As a native dutch speaker I can confirm that this is correct.

iamkun:
cheers @donroyco @bas080

donroyco:
Thanks @bas080 and @iamkun !

PR #276, opened 2018-07-10
chore: update set week logic

iamkun:


iamkun:
@bas080 Updated the set logic a little bit to make it easy to understand.

Issue #277, opened 2018-07-10
Module has no has no default export with ts and rollup


waitingsong:
```ts
import dayjs from 'dayjs'
```
compile with rollup , got error `Module 
> ".../source/node_modules/dayjs/index" has no default export.

kaseyhinton:
any updates on this?

Kikisek:
Try `import * as dayjs from 'dayjs'`, it helped me 😃

aibrahim3546:
@Kikisek I did like what you mentioned...

like this:

`import * as dayjs from dayjs`

I am hitting this error now:

> Uncaught TypeError: dayjs is not a  function

samtsai:
@aibrahim3546 to be sure, do you have quotes around the second `dayjs`?

`import * as dayjs from 'dayjs'`

wangyiz4262:
@waitingsong @iamkun I re-did the Typescript definition in #466 which should fix the problem of default import. The current `index.d.ts` file does not make much sense to expose class to the user.

iamkun:
Updates:
After v1.8.1 we could simply use `import dayjs from 'dayjs'` in a TypeScript project.

Smirking:
Looks like this was reverted again here https://github.com/iamkun/dayjs/pull/476, so it's broken again as of 1.8.4 for rollup and things that use rollup, ng-packagr etc

conradolk:
As @Smirking said, this is broken as of now and we have to use @Kikisek workaround.

orige:
Even with @Kikisek workaround, I'm getting:
`TypeError: dayjs is not a function`

**version**:  1.8.28

blestab:
Also getting this **version:** 1.8.35

Lacroute:
For those who are experiencing the `Error: 'default' is not exported by node_modules/dayjs/dayjs.min.js` issue, add `@rollup/plugin-commonjs` as a dev dependency and configure your `jest.config.js` like below:

```
import { nodeResolve } from '@rollup/plugin-node-resolve';
import commonjs from '@rollup/plugin-commonjs';
import { terser } from "rollup-plugin-terser";

export default {
  input: 'src/index.js',
  output: {
    file: 'dist/index.min.js',
    format: 'cjs',
    plugins: [
      terser()
    ]
  },
  plugins: [
    nodeResolve(),
    commonjs()
  ]
};
```

risalfajar:
1.10.7, error still happened

Have tried the workaround but still not working

[...]

PR #278, opened 2018-07-10
fix(typings): default export issue#277

waitingsong:
ref: https://github.com/iamkun/dayjs/issues/277

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -1,6 +1,6 @@
-export = dayjs;
 
 declare function dayjs (config?: dayjs.ConfigType, option?: dayjs.OptionType): dayjs.Dayjs
+export default dayjs
 
 declare namespace dayjs {
   export type ConfigType = string | number | Date | Dayjs


iamkun:
Thanks. 

I'm new to TS, so reviewers are wanted. 

Related issue #277

waitingsong:
think about rewriting with ts?

prantlf:
This fixes a clear inconsistency in Day.js by making the TypeScript declaration file use the default export. The original JavaScript file used default export, but the TypeScript declaration file used named export.

Do you want to use default exports or named exports in Day.js? The latter is better for refactoring and ensuring the right naming conventions, when a module is used. However, it is not so important for Day.js, regarding its size and role as a 3rd-party library in other projects. It can be discussed any time again.

wangyiz4262:
Please refer to my comments [here](https://github.com/iamkun/dayjs/issues/277#issuecomment-457966636)

PR #283, opened 2018-07-18
add a plugin named "preciseDiff" and its test code for "dayjs"

huangjinlin:
this is how about it:https://github.com/huangjinlin/dayjs-precise-range

### DIFF
--- a/test/plugin/preciseDiff.test.js
+++ b/test/plugin/preciseDiff.test.js
@@ -1,60 +1,65 @@
 import dayjs from '../../src'
 import preciseDiff from '../../src/plugin/preciseDiff'
+
 dayjs.extend(preciseDiff)
 
 it('Diff of two time', () => {
-  expect( dayjs.preciseDiff(dayjs('2014-01-01 12:00:00'),dayjs('2014-02-03 15:04:05')) ).toBe("1 month 2 days 3 hours 4 minutes 5 seconds");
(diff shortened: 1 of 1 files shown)


huangjinlin:
`
var dayjs = require('dayjs');
var preciseDiff = require('dayjs/plugin/preciseDiff')
dayjs.extend(preciseDiff);
var m1 = dayjs('2014-01-01 12:00:00');
var m2 = dayjs('2014-02-03 15:04:05');
var diff = dayjs.preciseDiff(m1, m2); // '1 month 2 days 3 hours 4 minutes 5 seconds'
`

iamkun:
CI failed. Please run `npm run lint` before open a pull request. Thanks.

huangjinlin:
@iamkun i have modify through eslint

iamkun:
@huangjinlin Cool thanks. I'm waiting for others opinions before merging. Please be patient.

bas080:
I would expect extensions to be an opt-in thing. I am more for a separate library approach.

The code/tests you have written look great. You might not want maintenance responsibility, or you would like it to be part of the dayjs ecosystem. Maybe @iamkun can start an organization and create separate repositories for these libraries that build on dayjs. If that is the case, you would need to document and point to the libraries in the dayjs docs, similar to what moment.js docs do.

iamkun:
@huangjinlin Thanks for your code.

Still, I tend to agree not to merge this into our main code base.

Instead, you may could publish your plugin by yourself. 

And I'll consider @bas080 's words, and still, a lot work to do. 😬

cloydlau:
That preciseDiff plugin is under ISC license...

Issue #287, opened 2018-07-30
Not support isBetween() API


wangjianio:

lamartire:
@iamkun what about it? I think, I can implement that method.

iamkun:
@lamartire  That would be nice. Just one question, a core API or a isBetween plugin?

lamartire:
@iamkun , no matter, if it question about implementation. What will be better in your opinion?

lamartire:
Got it. I will try to do it in few days.

iamkun:
@wangjianio added isBetween plugin

iamkun:
Update: Support isBetween() API  https://day.js.org/docs/en/query/is-between

PR #296, opened 2018-08-04
Implements isBetween plugin

lamartire:
Sets testURL in jest config for resolve issue with failing tests
#287

### DIFF
--- a/package.json
+++ b/package.json
@@ -26,6 +26,7 @@
       "test"
     ],
     "testRegex": "test/(.*?/)?.*test.js$",
+    "testURL": "http://localhost",
     "coverageDirectory": "./coverage/",
     "collectCoverage": true,
     "collectCoverageFrom": [
--- a/src/plugin/isBetween/index.js
+++ b/src/plugin/isBetween/index.js
@@ -1,16 +1,10 @@
 export default (o, c, d) => {
   const proto = c.prototype
   proto.isBetween = function (a, b) {
-    const dayA = d(a)
-    const dayB = d(b)
+    const dA = d(a)
+    const dB = d(b)
 
-    if (this.isAfter(dayA) && this.isBefore(dayB)) {
-      return true
-    } else if (this.isBefore(dayA) && this.isAfter(dayB)) {
-      return true
-    }
-
-    return false
+    return (this.isAfter(dA) && this.isBefore(dB)) || (this.isBefore(dA) && this.isAfter(dB))
   }
 }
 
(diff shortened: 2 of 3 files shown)


src/plugin/isBetween/index.js
```diff
@@ -0,0 +1,11 @@
+export default (o, c) => {
+  const proto = c.prototype
+  proto.isBetween = function (a, b) {
+    if (!(a instanceof c && b instanceof c)) {
+      throw new Error('isBetween arguments must be dayjs-like!')
+    }
+
```
iamkun: `dayjs(a), dayjs(b)` might be better.

src/plugin/isBetween/index.js
```diff
@@ -0,0 +1,11 @@
...
+    return this.isAfter(a) && this.isBefore(b)
+  }
```
iamkun: also do we need `this.isAfter(b) && this.isBefore(a)` ?

src/plugin/isBetween/index.js
lamartire: Of course.

src/plugin/isBetween/index.js
lamartire: What you mean? It's just re-using exist methods. Tell me more 🙂

iamkun: Just not sure if we should check the two compared date in different order.

```
// the first one could be true, if we think it's useful
moment('2018-05-01').isBetween('2019-05-01', '2017-05-01') // false 
moment('2018-05-01').isBetween('2017-05-01', '2019-05-01') // true
```

lamartire: Oh, I see. Give me few time.

lamartire:
I changed logic, I think it is more readable and clear now.

src/plugin/isBetween/index.js
iamkun: see your latest commit, cool.

still, it's just a discussion of my idea not a decision

lamartire: Why? If we working just with two dates it is really simple to compare our date with they. Or you mean more than 2 arguments?

iamkun: no, just a discussion about if we should ignore the given order or not.

package.json
```diff
@@ -26,6 +26,7 @@
       "test"
     ],
     "testRegex": "test/(.*?/)?.*test.js$",
+    "testURL": "http://localhost",
```
iamkun: what's this config used for?

lamartire: On osx last versions of jest throws error, `localStorage is not available for opaque origins`. It solves only with this option in config.

lamartire:
@iamkun whats next? 🙂

lamartire:
Wow! 🙌

PR #300, opened 2018-08-14
Fix UnitType strings in Api docs

idready:


iamkun:
@idready any description?

idready:
@iamkun, my bad. This will only fix readme api description, the corresponding **UnitType** was false for diff method. There was 'days' instead of 'day' for example. I replaced all false **UnitType** with the one exported.

iamkun:
Thanks @idready. At the moment, we support both singular and plural nouns. And this PR seems not necessary. 

Feel free to re-open it if I am wrong.

PR #301, opened 2018-08-15
fix: locale return type should be 'Dayjs'

jovey-zheng:


iamkun:
@jovey-zheng THX. 

However, `dayjs.locale(es)` returns the locale string `es`, and `dayjs().locale(es)` returns the dayjs instance.

PR #305, opened 2018-08-16
Add LocalizableFormat plugin supplying format options known from Moment.js

prantlf:
High-level options `L LT LTS ...` can be configured to patterns comprised of low-level options `h m y ...` in the `formats` object in a locale definition.

Default format options are English [defined in the plugin source code](https://github.com/prantlf/dayjs/blob/localisable-formats/src/plugin/localizableFormat/index.js). [German localization overrides](https://github.com/prantlf/dayjs/blob/localisable-formats/src/locale/de.js) them accordingly.


tcrands:
@prantlf @iamkun I notice this hasn't been worked on for a while but has recently been tagged for next release. Are there plans for this to go in soon? Currently migrating a project away from moment and this would be very useful.

iamkun:
@tcrands will be available in next release soon.

PR #304, opened 2018-08-16
Allow grammatically correct localisation of the RelativeTime plugin to fusional languages

prantlf:
* Redefine the locale object structure to separate expressions for duration, future and past. Allow the proper declension of the noun coming after the numeral.
* Automatically convert the old locale object structure to the new one during the first call to a method of the RelativeTime plugin.

Redefining the locale object structure from the old one:

```js
  relativeTime: {
    future: 'in %s',
    past: '%s ago',
    s: 'a few seconds',
    ...
    yy: '%d years'
  }
```

to the new one solves the problem 1 from #302:

```js
  relativeTime: {
    duration: {
      s: 'a few seconds',
      ...
      yy: '%d years'
    },
    future: {
      s: 'in a few seconds',
      ...
      yy: 'in %d years'
    },
    past: {
      s: 'a few seconds ago',
      ...
      yy: '%d years ago'
    }
  }
```

Adding second expressions with three-letter keys for numerals greater or equal to 5 solves the problem 2 from #302:

```js
    duration: {
      s: 'a few seconds',
      m: 'a minute',
      mm: '%d minutes',
      mmm: '%d minutes',
      ...
    },
```

[My Czech localisation](https://github.com/prantlf/dayjs/blob/fusional-languages/src/locale/cs.js) is an example how to use the new locale structure to support grammatically correct formatting.

Issue #302, opened 2018-08-15
Localisation capabilities of the RelativeTime plugin are insufficient for fusional languages


prantlf:
What do you think about re-working the current locale definition of the RelativeTime plugin to support fusional languages without grammatical mistakes?

[Moment.js has more complicated locale definitions](https://github.com/moment/moment/blob/develop/locale/cs.js), but it allows grammatically correct localisation for Slavic languages, for example. [Your Russian localisation](https://github.com/iamkun/dayjs/blob/master/src/locale/ru.js) suffers from the second problem below and [my Czech localisation](https://github.com/prantlf/dayjs/blob/czech-locale/src/locale/cs.js) from both, which makes the localised expressions look so wrong, that they cannot be used in real-world applications.

## Problem

1. Localisation expressions for *future* ("after %s") and *past* ("before %s") contain the same rest of the expression. However, they contain different prepositions and thus they may need different declension in the rest of the expression.
2. There is only one localisation expression for multiple seconds ("%d seconds"), minutes ("%d minutes") etc. However, different numerals may need different declension of the following noun.

For example, Slavic languages do not use just one noun form for singlular (one second) and one for plural (many seconds). Usually there are different forms after numerals 1, 2-4 and 5+. Additionally, the noun changes according to to the grammatical case, which is needed in the particular expression. The grammatical case depends on the preposition, for example.

### Years - Past

```txt
 a year ago    vor einem Jahr   в прошлом году   vloni (před rokem)
 2 years ago   vor 2 Jahren     2 года назад     před 2 roky
 5 years ago   vor 5 Jahren     5 лет назад      před 5 lety
```

### Years - Future

```txt
 in a year     in einem Jahr   через год      za rok
 in 2 years    in 2 Jahren     через 2 года   za 2 roky
 in 5 years    in 5 Jahren     через 5 лет    za 5 let
```

### Days - Past

```txt
 yesterday    gestern          вчера          včera
(a day ago    vor einem Tag)   день назад     před jedním dnem)
 2 days ago   vor 2 Tagen      2 дни назад    před 2 dny
 5 days ago   vor 5 Tagen      5 дней назад   před 5 dny
```

### Days - Future

```txt
 tomorrow    morgen         завтра         zítra
(in a day    in einem Tag   через день     za den)
 in 2 days   in 2 Tagen     через 2 дни    za 2 dny
 in 5 days   in 5 Tagen     через 5 дней   za 5 dní
```

### Seconds - Past

```txt
[...]

PR #312, opened 2018-08-24
Add relative locales to czech language

NikolaRusakov:
Add czech language locale fuctionality

### DIFF
--- a/src/locale/cs.js
+++ b/src/locale/cs.js
@@ -1,0 +1,27 @@
+import dayjs from 'dayjs'
+
+const locale = {
+  name: 'cs',
+  weekdays: 'Pondělí_Úterý_Středa_Čtvrtek_Pátek_Sobota_Neděle'.split('_'),
+  months: 'Leden_Únor_Březen_Duben_Květen_Červen_Červenec_Srpen_Září_Říjen_Listopad_Prosinec'.split('_'),
+  relativeTime: {
+    future: 'po %s',
+    past: '%s před',
+    s: 'vteřin',
+    m: 'minuta',
+    mm: '%d minut',
+    h: 'hodina',
+    hh: '%d hodin',
+    d: 'den',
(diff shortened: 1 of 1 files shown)


prantlf:
Coincidentally, I [added the Czech locale](https://github.com/prantlf/dayjs/blob/czech-locale/src/locale/cs.js) too :-) As a part of PR #305 with the plugin called [LocalisableFormat].

Differences to this PR:

* The key "s" translated as "several seconds" and not "second".
* Keys for the single value (s, m, h, d, M, Y) translated with the correct declension:
  - if used in the expression for the future.
  - if used in the expression with duration; only if the duration is used in accusative - in the expression "takes ..."., for example 

In my case, expressions for the past cannot be translated correctly,because the preposition needs a different declension. In your case, you decided to translate correctly expressions of the past (although moving the preposition "before" after the value feels weird) and durations in nominative.

I opened issue #302 to explain and discuss the problem of the correct localisation. I sent PR #304, which localises the three scenarios (duration, future and past) with three different texts, if needed in a particular locale. I [updated the Czech locale](https://github.com/prantlf/dayjs/blob/fusional-languages/src/locale/cs.js) in this PR accordingly.

[LocalisableFormat]: https://github.com/prantlf/dayjs/blob/localisable-formats/docs/en/Plugin.md#localisableformat

greegus:
@NikolaRusakov beware that the weekdays list should be indexed from Sunday, not Monday.
https://github.com/iamkun/dayjs/blob/master/src/locale/de.js

stepanh:
Czech language requires more complex rules as highlighted by @prantlf above

iamkun:
I'll merge #461 first, cause weekdays list should be indexed from Sunday.

BTW, @stepanh I'll do it later to support more complex rules like prantlf mentioned above.

stepanh:
cc @prantlf

PR #315, opened 2018-08-28
Undo the capital letters for Russian locale

mvshmakov:
It's uncommon that all weekdays and months are starting from the capital letters. It would be much easier to migrate from moment.js on day.js if they were in lower case (https://github.com/moment/moment/blob/2e2a5b35439665d4b0200143d808a7c26d6cd30f/locale/ru.js#L38).


iamkun:
THX

Issue #323, opened 2018-09-03
Add a plugin with the time zone support


prantlf:
## Business Case

A typical application, which displayed dates and expects dates entered by the users handles the time zone:

1. Consistently. The user should not be confused by different components working in different time zones.
2. Either using the local time zone from the web browser (operating system).
3. Or using the time zone specified in the user settings of the web application.

## Interface Synopsis

A minimal support could include the following two scenarios, typical for applications displaying and editing dates:

1. From the storage to the presentation. Format a UTC date to show it converted to the time zone chosen by the user.
2. From the presentation to the storage. Parse a date entered by the user in their chosen time zone and convert it to UTC.

The API for these scenarios could use the `constructor` and the `format` method:

```js
const timeZone = 'Europe/Berlin' // Canonical time zone name

// String without a time zone from the date picker
const enteredDate = '2018-09-02 23:41:22'
const storedDate = dayjs(enteredDate, { timeZone }).toISOString()
// Will contain "2018-09-02T21:41:22Z"

// String in UTC or anything that Day.js accepts
const storedDate = '2018-09-02T21:41:22Z'
const userFormat = 'D.M.YYYY H:mm:ss [GMT]Z (z)' // Standard Day.js format
const displayedDate = dayjs(storedDate).format(userFormat, { timeZone })
// Will contain "2.9.2018 23:41:22 GMT+02:00 (CEST)"
```
## Implementation Proposal

Day.js uses an embedded `Date` object. That is why this library is so lightweight, but it limits its functionality. The `Date` object supports only local time zone and UTC. It does not work in other time zones. It is possible to continue leveraging this principle. without making more complicated by replacing internal `Date` object with something else. All other methods continue working as expected.

### dayjs(input: any, { timeZone: boolean }?)

Day.js already supports an optional `options` object in the constructor and the `parse` method. The time zone parameter in the constructor is meant only for converting the parsed input string correctly to UTC. The embedded `Date` object will be initialised with UTC and offer the local time zone representation as usual. The original time zone offset will not be remembered. It is usually not important, because dates should be rendered consistently in user's time zone; not in various time zones, which their string sources referred to.

### format(format: string, { timeZone: boolean }?)

The `options` object is optionally recognised by the overridden `format` method. The time zone parameter will extract the date parts (year, month, ...) from the embedded `this.$d` object in UTC and convert them to the specified time zone, before producing the output string.

[...]

Issue #46, opened 2018-04-24
Olson Timezone Support


alanhamlett:
Update: Day.js Time Zone Plugin https://day.js.org/docs/en/timezone/timezone

------

This is great, but without timezone support it's not usable in real-world applications.

lhcgoncalves:
Agreed!

iamkun:
+ 1
I am just thinking if we should add this to our core code, or as a plugin.

Plus, can you share some of timezone use cases? Or which API do we need?

alanhamlett:
> Plus, can you share some of timezone use cases? Or which API do we need?

The ability to parse and render dayjs objects from/in any timezone. Matching the [Moment Timezone API](https://momentjs.com/timezone/docs/#/using-timezones/) probably makes sense for portability.

kamranayub:
A plugin would be acceptable, especially to keep core low (great for perf-minded folks). Especially since then we could take advantage of code-splitting, dynamic imports, etc. where it makes sense to do TZ related stuff (which admittedly, can be for the entire app but at least you're opting in).

iamkun:
@kamranayub we are working on is, PR #168 

Any ideas?

prantlf:
I opene a feature request #323 for introducing just parsing and formatting support with a specific time zone. I can imagine implementation using both Olson time zone database or the new `Intl` support in the browser.

I do no know, if you wanted to make Day.js more complicated by remembering the original time zone in the Day.js object itself and supporting functions like `.tz()` or `.utcOffset()`. But I believe, that it is possible with still retaining the internal `Date` object, with some limitations.

xxyuk:
I do this:

```

dayjs(new Date().toLocaleString("en-US", {timeZone: "America/New_York"})).format('h:mA')


dayjs(new Date().toLocaleString("en-US", {timeZone: "America/Los_Angeles"})).format('h:mA')

```

 demo -> https://plnkr.co/edit/KjrKnwvxBktn3GJmDwDw

Refer: https://medium.com/@bmarti44/10-helpful-tips-i-wish-i-knew-before-i-started-using-angular-2-c25bcaa6fe2f

IssuehuntBot:
@issuehunt has funded $200.00 to this issue.

---
- Submit pull request via [IssueHunt](https://issuehunt.io/repos/128907699/issues/46) to receive this reward.
- Want to contribute? Chip in to this issue via [IssueHunt](https://issuehunt.io/repos/128907699/issues/46).
[...]

PR #325, opened 2018-09-04
TimeZone - a plugin with time zone support for parsing and formatting strings

prantlf:
Attempts to implement #323 by introducing the [TimeZone plugin](https://github.com/prantlf/dayjs/blob/format-timezone/docs/en/Plugin.md#timezone).

It uses the [IANA time zone](https://www.iana.org/time-zones) database, packed to a JSON format by [moment-time-zone](https://github.com/moment/moment-timezone) and with a simple interface to it provided by [timezone-support](https://github.com/prantlf/timezone-support).

Although the `Intl` time zone support has been introduced, there are still browsers or platforms, which do not support it and would not mind "paying for" the support by including more JavaScript code. The size of the time zone supporting library without the Day.js plugin: source 185 kB, minified 182 kB, gzipped 24 kB.

I can imagine adding support for remembering the time zone for future calls to `.format()` as discussed at #46.

Time zone conversions are possible already without an additional API. Jut use UTC date from one Day.js object and format it the target time zone using the other (or the same) Day.js object.

### DIFF
--- a/docs/en/Plugin.md
+++ b/docs/en/Plugin.md
@@ -201,6 +201,22 @@ Day.js uses an embedded `Date` object. This object supports only local time zone
 * The time zone parameter in the constructor is meant only for converting the parsed input string correctly to UTC. The embedded `Date` object will be initialised with UTC and offer the local time zone representation as usual. The original time zone offset will not be remembered. It is usually not important, because dates should be rendered consistently in user's time zone; not in various time zones, which their string sources referred to.
 * The time zone parameter in the `format` method will extract the date parts (year, month, ...) from the embedded `Date` object in UTC and convert them to the specified time zone, before producing the output string.
 
+#### Installation
+
+This plugin has a peer dependency on the `timezone-support` NPM module. If you are going to use it in your project, add it to your dependencies too, along the dependency on `Day.js`:
+
+```sh
+npm i --save dayjs timezone-support
+```
+
+If you use `Day.js` and this plugin on a web page, include the time-zone module too:
+
+```html
+<script arc="https://unpkg.com/dayjs/dayjs.min.js"></script>
+<script arc="https://unpkg.com/dayjs/plugin/timeZone.js"></script>
+<script arc="https://unpkg.com/timezone-support/dist/index.umd.js"></script>
+```
+
 ## Customize
 
 You could build your own Day.js plugin to meet different needs.
--- a/docs/en/Plugin.md
+++ b/docs/en/Plugin.md
@@ -203,13 +203,7 @@ Day.js uses an embedded `Date` object. This object supports only local time zone
 
 #### Installation
 
-This plugin has a peer dependency on the `timezone-support` NPM module. If you are going to use it in your project, add it to your dependencies too, along the dependency on `Day.js`:
-
-```sh
-npm i --save dayjs timezone-support
-```
-
-If you use `Day.js` and this plugin on a web page, include the time-zone module too:
+This plugin has a dependency on the [`timezone-support`](https://www.npmjs.com/package/timezone-support) NPM module. If you are going to use it on a web page directly, add its script to your section of `<script>`s too, along with the `Day.js`, for example:
 
 ```html
 <script arc="https://unpkg.com/dayjs/dayjs.min.js"></script>
--- a/docs/en/Plugin.md
+++ b/docs/en/Plugin.md
@@ -166,6 +166,41 @@ dayjs.extend(isBetween)
 dayjs('2010-10-20').isBetween('2010-10-19', dayjs('2010-10-25')); // true
 ```
 
+### TimeZone
+ - TimeZone extends `dayjs()` and `dayjs().format` APIs to support the most important usage scenatrios - parsing from a specific time zone and formatting in other time zone.
+
+```javascript
+import timeZone from 'dayjs/plugin/timeZone'
+
+dayjs.extend(timeZone)
+
+const timeZone = 'Europe/Berlin' // Canonical time zone name
+
+// String without a time zone from the date picker
+const enteredDate = '2018-09-02 23:41:22'
+const storedDate = dayjs(enteredDate, { timeZone }).toISOString()
+// Will contain "2018-09-02T21:41:22Z"
+
+// String in UTC or anything that Day.js accepts
+const storedDate = '2018-09-02T21:41:22Z'
+const userFormat = 'D.M.YYYY H:mm:ss [GMT]Z (z)' // Standard Day.js format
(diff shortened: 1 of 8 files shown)


prantlf:
Build fails only on Node.js 6, because of the osage of `async` keyword in `iamkun/dayjs/build/index.js`. Do you want to drop the Node.js 6 support, or use only JavaScript features supported by Node.js 6?

iamkun:
we build `dayjs` only on CI (node 10)

prantlf:
@iamkun , this is your configuration in .travis.yml:

```yml
node_js:
  - '10'
  - '8'
  - '6'
```

I opened PR #327, which drops the Node.js 6 line, for your consideration. Thanks!

prantlf:
Oh, I am sorry, @iamkun. I implemented other workaround for not having the built files for testing.

I will need to figure out the code coverage yet, which does not show me any report.

iamkun:
@prantlf you may could run `npm run test` on your local machine.

prantlf:
Yes, that helped, thanks! :-) I think, that I confused codecov by rebasing the PR branch to the current master and forcing the re-push.

pedro-mass:
While this does switch the dates and times, it seems to not change the day of the week.

Example:
```js
const timestamp = 1536901200000
const timezone = 'America/Denver'
const formatString = 'MM/DD/YYYY dddd hh:mma'

const dayString = dayjs(timestamp).format(formatString, {
  timeZone: timezone,
})

console.log(dayString) 	// returns: '09/13/2018 Friday 11:00pm'
						// should be: '09/13/2018 Thursday 11:00pm'

```

prantlf:
@pedro-mass, true, I ignored the day of the week in the original submission. I added it by the latest update. All properties of the `days` object should be covered now.

pedro-mass:
@prantlf This looks awesome, thanks for the hard work and contribution! 
Your `timezone-support` library also seems pretty interesting. Keep up the good work 👏

prantlf:
Thank you! Let's see how it goes on :-)

I was thinking about introducing an interface compatible with Moment.js:

```js
// Alternative to the constructor:
// dayjs('2018-09-29 21:34', { timeZone: 'Europe/Berlin' })
dayjs.tz('2018-09-29 21:34', 'Europe/Berlin')

// Convert an existing Day.js instance to other time zone
dayjs().tz('2018-09-29 21:34', 'Europe/Berlin')
```

The alternative to the constructor means no problem to the further usage. You might question if it is worth adding, because you need to pass the time zone as a parameter anyway, unlike the static method `dayjs.utc(...)`. Another thing might be the lack of the instance method `dayjs().tz(...)`, because it may not be wanted because of its problems. Anyway, no technical no problems, just some artificial subjective opinions ;-)

The time zone conversion for a Day.js instance may confuse the end-user. Day.js stores a `Date` instance internally and caches the date and time parts in `this.$*` properties used by getting and formatting methods. The `Date` object does not support arbitrary time zones. How to store a date in a particular time zone?

1. Overhaul Day.js not to use the `Date` object. The converting method `toDate()` would return the native `Date` object in the local time, possibly in a different time zone, than the Day.js instance. This would break the plugins, which depend on `this.$d`.
2. Add a new object to store the time-zone-aware date. This would make the Day.js instance more complicated and need some synchronization with `this.$d` to prevent confusion.
3.  Do not store the time zone information in Day.js instances. Use it only in the constructor/parse method to convert the input to UTC and store it in the Day.js instance, and in the formatting method to output a date string in a request ed time zone. This would not allow to carry the *original* time zone information together with the date value.

The first approach would be interesting to prototype. It could turn up lightweight and quick. I chose the third approach, because it was not intrusive, it was feasible as a plugin and the interface could be kept if the first approach won later.

However, The third approach does store the time zone information *publicly* in a `Day.js` instance, which means, that converting one Day.js instance to another one in an arbitrary time zone is not possible. When I could not add the instance method `tz`, I did not find adding the static method `tz` a good idea.

aralroca:
@prantlf Thanks for that great contribution! The interface compatible with Moment.js would be very great. 👏

prantlf:
@aralroca, I'm afraid, that I went for the option 3 from my comment above, which doesn't allow implementing the interface fully compatible with Moment.js... :'-{

I could implement the static `tz` method parsing the input in the specified time zone. However, the instance method `tz` cannot be implemented with the the option 3, because this option doesn't store the time zone in the `dayjs` object.

The trouble with the static `tz` from the option 3 is, that it converts the date to the local time zone, which is the only one supported by the native `Date` object.

```js
const date = dayjs.tz('2018-09-29 21:34', 'Europe/Berlin')
// Some code here, or the date may be passed to a distant function
const output = date.format('DD/MM/YYYY' )
```

What would you expect from such `dayjs` object?  Would you expect the output in the original time zone, or in the local time, as it is done now?

petermikitsh:
Hi, I'm inquiring to the status of this PR. Would be incredibly useful to have timezone support in dayjs. Thanks.

danielb2:
curious about status too

petermikitsh:
@danielb2 You might not need dayjs. `Intl.DateTimeFormat` may give you what you need.

https://caniuse.com/#search=datetimeformat

For older browsers, there's a polyfill (npm module `intl`):

https://bundlephobia.com/result?p=intl@1.2.5

bissolli:
@iamkun @prantlf, any news on it? Would be amazing having this plugin.

trylovetom:
Any Update? I can't wait for this plugin.

iamkun:
This pr introduced a heavy dependency which is not really suitable to dayjs. We will introduce a time zone plugin base on `Intl.DateTimeFormat`.

trylovetom:
@iamkun so you will adopt #608 ?

iamkun:
Day.js Time Zone Plugin https://day.js.org/docs/en/timezone/timezone

docs/en/Plugin.md
```diff
@@ -166,6 +166,51 @@ dayjs.extend(isBetween)
...
+const userFormat = 'D.M.YYYY H:mm:ss [GMT]Z (z)' // Standard Day.js format
+const displayedDate = dayjs(storedDate).format(userFormat, { timeZone })
+// Will contain "2.9.2018 23:41:22 GMT+02:00 (CEST)"
+```
+
+List of added formats:
+
+| Format | Output | Description             |
+| ------ | ------ | ----------------------- |
+| `z`    | CEST   | Time zone abbreviation  |
```
akoskm, line 195: @iamkun any chance to implement this as part of the timezone plugin?

zazzaz: @akoskm  I have the same need and draft a PR here https://github.com/iamkun/dayjs/pull/1069

PR #326, opened 2018-09-05
Introduce an UTC mode simpler than #168

prantlf:
The "UTC mode" means, that he constructor assumes the UTC date in the input string and the instance methods return and accept date parts (year, month etc.) in UTC.

## Additions

* Constructor accepts a boolean flag to enable the UTC mode: `{ utc: true }`.
* Method `utc()` returns a clone of the instance in the UTC mode.
* Method `local()` returns a clone of the instance in the non-UTC (local) mode.
* Method `isUTC()` checks, if the instance is in the UTC mode.
* Method `utcOffset()` returns the time zone offset to UTC consistently with `Date.prototype.getTimezoneOffset()`

##  Differences to the "normal" (local) dayjs mode

* Assume UTC as the time zone of the string passed to the constructor - for example, "2018-10-28" will be parsed as "2018-10-28T00:00:00Z". *But only if the time zone is omitted.* If you end the string tith a TZ offset - "2018-10-28T00:00:00+02:00", it will be retained and the whole date will be converted to UTC, when iniitalizing the datejs object.
* Methods returning the date parts like `year()`, `month()` etc. return UTC parts using `Date.propertotype.getUTC*()` methods.
* Methods manipulating the date - `set()`, `add()` etc. - work with the UTC values of the date parts, as the getters mentioned above.
* The `format()` method uses the UTC getters too and always formats the time zone as "+00:00".
* The `utcOffset()` method always returns zero.

If a Day.js instance is constructed without the time part, it will appear, as if the instance runs in a "date-only" mode, where no conversion from UTC to the local time takes place, like the `Date` instance would do it. It can be used for attributes, where the time does not make sense and only the date should be maintained. If only the date applies, it should not be affected by time zone changes.

### DIFF
--- a/docs/en/API-reference.md
+++ b/docs/en/API-reference.md
@@ -45,6 +45,12 @@ The `Dayjs` object is immutable, that is, all API operations that change the `Da
     - [Is Same `.isSame(compared: Dayjs)`](#is-same-issamecompared-dayjs)
     - [Is After `.isAfter(compared: Dayjs)`](#is-after-isaftercompared-dayjs)
     - [Is a Dayjs `.isDayjs()`](#is-a-dayjs-isdayjscompared-any)
+  - [UTC Mode](#utc-mode)
+    - [Date-Only Mode`](#date-only-mode)
+    - [Is UTC `.isUTC()`](#is-utc-isutc)
+    - [To UTC `.utc() | dayjs(original: Dayjs, { utc: true })`](#to-utc-utcdayjsoriginal-dayjs-utc-true)
+    - [To Local `.local() | dayjs(original: Dayjs, { utc: false })`](#to-local-localdayjsoriginal-dayjs-utc-true)
+    - [UTC Offset `.utcOffset()`](#utc-offset-utcoffset)
   - [Plugin APIs](#plugin-apis)
     - [RelativeTime](#relativetime)
     - [IsLeapYear](#isleapyear)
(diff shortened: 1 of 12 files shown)


prantlf:
I do intend to provide the tests and documentation. Later :-)

prantlf:
I wonder if you liked more the interface:

```js
dayjs.utc('2018-09-25')
```

instead of the current:

```js
dayjs('2018-09-25', { utc: true })
```

I introduced the latter, because static methods may feel like having a global effect on `dayjs`, like `locale`.

prantlf:
I prototyped an implementation of the UTC mode as a plugin the branch [https://github.com/prantlf/dayjs/commits/utc-mode-plugin](utc-mode-plugin). It is rather an example, what should not be a plugin :-)

Leaving `Day.js` intact is impossible, because functions like `wrapper` are not accessed vial `Utils.wrapper` and thus not overriddable. I made a minimum change to `Day.js` to be able to write the plugin in the first commit. The first version of the plugin in the second commit was a brutal copy & paste, especially `parseDate`, `startOf` and `$set`.

Plugins are supposed to extend Day.js with additional functionality. Not to copy & paste `Day.js` logic to be able to add a feature to it. It would tightly couple the plugin to a particular `Day.js` version. Any fixes and enhancements done in `Day.js` would have to be copied to the plugin too. If such extensibility is desired, an interface in `Day.js` should be introduced to allow it.

I tried to put the differences between the local time zone and UTC modes on the interface. Basically, it is the `Date` construction and date part setters. I ended up with a couple of methods, which no other plugin will use, than the UTC mode support. I added two other commits for the date creation and the setters, but such interface feels weird. Moreover, using utilities like `wrapper` always with the source object (`Utils.wrapper`) will cause mistakes, because it is not usual., Especially in times, when JavaScript developers embrace object destructuring.

UTC mode does not feel such an exotic feature, or to increase the size so much, to leave it out of the core library.

sahibjotsaggu:
Any update on when this will be merged and released?

prantlf:
Eventually , I added the static method `utc` as an alternative way to create a UTC instance.

```js
dayjs.utc('2018-09-25')
```

It fits the purpose of Day.js to help migration from Moment.js. And the code is shorter, that the constructor option.

prantlf:
@sahibjotsaggu, I released my Day.js fork including this feature as a new [`day-ext`](https://www.npmjs.com/package/dayjs-ext) module for the time being. You can try depending on it, until the "fate" of this functionality is decided in the official module. (I needed to create a prototype with this and other features, which I opened PRs for. Depending on github branches causes problems with NPM registry proxies.)

sahibjotsaggu:
@prantlf Are there any known bugs/issues with `day-ext`? Is it safe to use in a commercial setting?

prantlf:
@sahibjotsaggu, I fixed all issues, that I've found so far, to be able to use Day.js in a DMS application, enterprise quality: features #304, #305, #325, #326, #340, bug #360. I do parsing, formatting, 15 locales, different time zones. No querying or manipulating dates; I don't write a date picker. I use plugins customParseFormat, localizableFormat, relativeTime, timeZone.

I intend to maintain the result as [`dayjs-ext`](https://www.npmjs.com/package/dayjs-ext), as long as my changes are no accepted to the original package. All of them are needed for a serious usage in an internationalized application.

iamkun:
I think put in a plugin is better, #517.   

Referenced some logic and test suits from this PR, great thanks.

PR #327, opened 2018-09-05
Do not build and test with Node.js 6 any more

prantlf:
Build scripts depend on JavaScript features available in Node.js 8 and newer.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,7 +6,6 @@ cache:
 node_js:
   - '10'
   - '8'
-  - '6'
 install:
   - npm install -g codecov
   - npm install


iamkun:
No need. We do unit test use 6 8 10 and build use 10.

prantlf:
I see no condition restricting the build to Node.js 10 in `.travis.yml` or `package.json`. [The log](https://travis-ci.org/iamkun/dayjs/jobs/424663980#L541) points to the build task in the Node.js 6 environment. How would you otherwise explain the build failure?

Do you want me to introduce such build condition instead?

iamkun:
Why adding this `npm run build` in `.travis.yml`? This build task makes your CI failure.

prantlf:
Oh, I am sorry, @iamkun. I did add the `npm run build` there. And forgot about it already :-)

I was not able to get jest working with a module imported from `node_modules` in the TimeZone plugin. I had to build the plugin and test the built version. I will look at it once more and deal with it somehow.

I am sorry for the trouble and thanks for your help!

PR #331, opened 2018-09-07
[add] add support of isSameOrBefore and isSameOrAfter methods

rajdee:
Add support for original methods such as isSameOrBefore and isSameOrAfter


prantlf:
The conditions can be written using the existing ones:

```js
d.isSameOrBefore() === (d.isSame() || !d.isAfter())
d.isSameOrAfter() === (d.isSame() || !d.isBefore())
```

Is it worth extending the interface of `dayjs`?

iamkun:
@rajdee Thanks for the PR. 

I'd like to add this as a plugin rather in our core bundle.

Besides, @prantlf's advice is useful.

iamkun:
added as plugin

PR #332, opened 2018-09-07
Fixed french ordinal

francisbourquestudyo:


src/locale/fr.js
```diff
@@ -19,7 +19,10 @@ const locale = {
     y: 'un an',
     yy: '%d ans'
   },
-  ordinal: n => `${n}º`
+  ordinal: (n) => {
+    const o = n === 1 ? 'er' : ''
```
iamkun: Checked the French ordinal in momnet.js, should this be `const o = n === 1 ? 'er' : 'e'` here?

francisbourquestudyo: It would be the case with an ordinary ordinal, but dates are different. 

> In French, to state the date, the ordinal numbers are not used, but we use the regular number, or 'cardinal' number (like one, two, three, and so on). For example, in English, we say 'July  4th,' but in French, we would say le 4 juillet. [...] For the first day of the month, we use the ordinal number for 'first,' premier. So January 1st is le 1er janvier in French.

[Reference](https://study.com/academy/lesson/ordinal-numbers-in-french.html) (it's the best one I could find in English).

PR #333, opened 2018-09-10
Add sr and sr-cyrl locales

limonte:
There are 2 official alphabets for Serbian language - latin and cyrilic.

Naming is chosen to be the same as in moment.js


iamkun:
THX, @limonte

limonte:
Thanks for reviewing @iamkun! Can't wait toswitch from `moment` to `dayjs` in my project, will be able to do that right after this change will be released 🎉

iamkun:
@limonte cheers.

PR #335, opened 2018-09-10
Add relative date helpers (#299)

lamartire:
Hey, @iamkun! After much time I finally made some semantic plugins.
Main question – may be move them into one plugin, because 90% of code are really similar?

### DIFF
--- a/src/plugin/isToday/index.js
+++ b/src/plugin/isToday/index.js
@@ -1,0 +1,11 @@
+export default (o, c, d) => {
+  const proto = c.prototype
+  proto.isToday = function () {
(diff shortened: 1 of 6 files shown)


prantlf:
I myself would prefer getting them only optionally by importing a plugin. But that may be just me - I would move `startOf`, `endOf` and `daysInMonth` to a plugin too :-)

lamartire:
@prantlf hello! I think it will be good if you will open new issue with your desires ;)

iamkun:
I'll close this issue cause I think this plugin is not that useful. Feel free to reopen if you have updates on this

PR #340, opened 2018-09-14
CustomParseFormat - a plugin for the constructor to parse strings in custom formats

prantlf:
[Only alpha-numeric non-localizable format tokens are supported](https://github.com/prantlf/dayjs/blob/custom-parse-format/docs/en/Plugin.md#list-of-all-available-format-tokens), which are usually the only ones needed in enterprise applications. Recognition of localizable tokens, like "January" is missing.

I "scavenged" the parsing implementation [from timezone-support's parse](https://github.com/prantlf/timezone-support/blob/master/src/parse-format/parse.js), which is a simplified implementation from Moment.js.


tmrony:
I want to use this plugin. But the PR is not yet merged, can you give any idea when this goes live?

Regards.

iamkun:
Looks great. One question here, is it OK to make second argument as a string instead of a config object?  That will make full compatible with moment.
And the problem is we might have to change our existing config object logic.

How do you think  @prantlf

PR #348, opened 2018-09-23
Add Norwegian Nynorsk Locale (nn)

semonje:
Adds support for Norwegian Nynorsk (`nn`).

_Not to be confused with the already existing Norwegian Bokmål (`nb`)._


iamkun:
THX @semonje

PR #351, opened 2018-09-25
Add Finnish locale

Haprog:
This adds basic support for Finnish locale but atm. proper Finnish support can't be implemented until issue https://github.com/iamkun/dayjs/issues/302 is fixed.

`relativeTime` properties are now configured for past tense usage only because I think it's a more common use case with this kind of library.

For now using future tense with this locale will result in incorrect forms and text that doesn't make sense.


iamkun:
@Haprog THX

ttsirkia:
@Haprog it seems that the problem with future times can be now fixed: https://github.com/iamkun/dayjs/pull/767

PR #360, opened 2018-09-29
Keep Day.js instance immutable, when constructed with a Date instance

prantlf:
The Dayjs constructor called from the `dayjs` function stores the original `Date` instance, which was passed to it. It does not clone it. If that `Date` instance is modified later, it will affect the `dayjs` instance too. It will modify its internal `this.$d` object and make it inconsistent with the other `this.$*` variables. Such modification from outside is unexpected and it will break the immutability of `dayjs` instances too.

I made the `dayjs` function - the public interface - clone the input `Date` instance, before passing to the constructor. I added a boolean flag to the `wrapper` function to prevent double cloning. Premature optimization is evil, but being ignorant to unnecessary repeated operations too :-) The default behaviour is to clone to prevent hidden errors.

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -32,11 +32,14 @@ const dayjs = (date, c) => {
     return date.clone()
   }
   const cfg = c || {}
+  if (date instanceof Date && cfg.cloneDate !== false) {
+    date = new Date(date)
+  }
   cfg.date = date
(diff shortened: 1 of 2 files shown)


iamkun:
Thanks, updated in #517 just call `new Date(date)`

PR #368, opened 2018-10-06
add setWeek fucntion to weekOfYear

aquaminer:


naulacambra:
You should add [tests](https://github.com/iamkun/dayjs/blob/master/test/plugin/weekOfYear.test.js) so the code coverage remains at 100%.

naulacambra:
You're comparing a `dayjs` instance, with a `moment` instance, so they will never be the same.

The first test, works because the function `.week()` return an integer representing to which week of the year belongs the current date. Your new method, (I assume) sets the week, changing the current date, and returns the instance itself.

``` javascript
expect(dayjs().week()).toBe(moment().week()) // expect([int]).toBe([int]) - apples to apples

expect(dayjs().setWeek(week)).toBe(moment().week(week)) // expect([dayjs]).toBe([moment]) - apples to oranges
```

You should change the test to

``` javascript
expect(dayjs().setWeek(week).week()).toBe(moment().week(week).week()) // expect([int]).toBe([int]) - apples to apples
```

And also, add another test like 
``` javascript
expect(dayjs().setWeek(week).date()).toBe(moment().week(week).date()) // expect([int]).toBe([int]) - apples to apples
```

I hope it helps

src/plugin/weekOfYear/index.js
```diff
@@ -2,7 +2,12 @@ import { MS, Y, D, W } from '../../constant'
 
 export default (o, c, d) => {
   const proto = c.prototype
-  proto.week = function () {
+  proto.week = function (week = 0) {
+    if (week > 0) {
```
naulacambra: Instead of checking if `week` is positive, you should check if it's undefined.

Momentjs allows set a negative week

``` javascript
console.log(moment('2018-01-20').week(-1).format()); // "2017-12-23T00:00:00+01:00"
```

So I think that we should as well

src/plugin/weekOfYear/index.js
```diff
@@ -2,7 +2,12 @@ import { MS, Y, D, W } from '../../constant'
...
+      // eslint-disable-next-line prefer-template
+      if (week > 52) throw new Error('In one year 52 weeks, the resulting value: ' + week)
```
naulacambra: As well, we should allow the user set any number of weeks.

``` javascript
console.log(moment('2018-01-20').week(55).format()); // "2019-01-19T00:00:00+01:00"
```

src/plugin/weekOfYear/index.js
```diff
@@ -12,4 +15,7 @@ export default (o, c, d) => {
     const diffInWeek = this.diff(compareDay, W, true)
     return Math.ceil(diffInWeek)
   }
+  proto.weeks = function (week = 0) {
```
naulacambra: Be aware that, if user call `weeks` alias, without arguments, the default value `week = 0` will set the week to 0. You should change the default value to `null`, the same for `week()`

```javascript
proto.weeks = function (week = null) {
    return this.week(week)
}
```

aquaminer:
@naulacambra sorry, did not pay attention. Fixed

PR #373, opened 2018-10-11
Feature: add plugin members

mushan0x0:
```js
dayjs('2010-10-1').members('2010-10-7'); // [dayjs('2010-10-1'), ..., dayjs('2010-10-7')]
```


naulacambra:
I think that this functionality should be inside a [`duration`](https://momentjs.com/docs/#/durations/) or [`recurrence`](https://momentjs.com/docs/#/plugins/recur/) plugin. 

This library tries to use the same api as `moment.js`, and even though there is a method `days()`, this one returns the [day of the week](https://momentjs.com/docs/#/get-set/day/).

My suggestion, would be come up with another name for the method, or crete a whole new plugin like the mentioned before, and include this functionality in it.

And as an improvement, it would be nice adding a second parameter to let the user decide if he wants to include the first and last dates into the result.

``` javascript
dayjs('2010-10-1').days('2010-10-3'); // [dayjs('2010-10-1'), dayjs('2010-10-2'), dayjs('2010-10-3')]
dayjs('2010-10-1').days('2010-10-3', false); // [dayjs('2010-10-2')]
```

mushan0x0:
How about the API name `members`?

naulacambra:
I don't think that `members` would be a meaningful or intuitive name.

I would bet for something like `daysBetween`, based on the method [`isBetween`](https://momentjs.com/docs/#/query/is-between/)

Also, I would like to hear from @iamkun about the `duration` or `recurrence` plugin.

mushan0x0:
@iamkun 这个会合并吗

iamkun:
We will not merge this pr this time.

But still, you can publish your plugin by your own.

PR #380, opened 2018-10-14
add spanish documentation & extra locale

soulchainer:
Note: I cloned `src/locale/es.js` into `src/locale/es-es.js`,
similarly to how was done with `pt` and `pt-br`.
The reason is obvious: so many Spanish dialects. Leaving it with only
`es` would be problematic, especially when someone from another
Spanish-speaking country come to suggest editions because of this
Spanish not being the Spanish he is used to (understandable). And this is, as far as I can see,
already done and accepted with, at least, `pt`, so I think this
will be OK. Tell me if not and I will edit it
(remove it `/src/locale/es-es`, docs better have that way, so much text, could
be so different between countries) but have in mind the previous reasoning ^^.


iamkun:
Great, thanks.

iamkun:
I'll merge this PR first. If there is any updates or grammar corrections, feel free to send new PR.

Issue #377, opened 2018-10-12
weekdayShort localisation request


XXXVII:
You need to implement weekdayShort field in localisation files instead of just trimming weekday values. Otherwise you'll have awkward results for different locales. Currently, Russian is unsupportive because of this issue. For instance, Friday (Пятница) trims to Пя instead of Пт that is used in the language.
```
const getShort = (arr, index, full, length) => (
  (arr && arr[index]) || full[index].substr(0, length)
)
```
is a wrong way handling in multy-localisation project.

xxyuk:
there are `weekdaysShort` and `weekdaysMin` options in dayjs's locale config.

You could send a PR with the correct code in your language.

PR #381, opened 2018-10-14
Issue 377

denisinvader:
#377 

Texts from [this site](http://tetran.ru/SiteContent/Details/4)


denisinvader:
I've provided wrong weekdays order and creating another pull request with correct order

PR #383, opened 2018-10-14
Add shot months and weekdays to some locales

denisinvader:
In addition to #377


prantlf:
@denisinvader, German substantive start with an upper-case letter. You should follow this rule for shortened words too.

PR #387, opened 2018-10-17
Add install dayjs with yarn in readme

shaminmeerankutty:
I thought it would be great if we also give dayjs installation via yarn in the README and Installation guide. So it will be useful for developers, like me, who uses yarn to know that dayjs is available in yarn as well.

- Added yarn installation command in README.md
- Added yarn installation command in Installation.md

### DIFF
--- a/README.md
+++ b/README.md
@@ -42,8 +42,13 @@ dayjs().startOf('month').add(1, 'day').set('year', 2018).format('YYYY-MM-DD HH:m
 
 ### Installation
 
+Install dayjs using `npm`:
 ```console
 npm install dayjs --save
+```
+Or via `yarn`:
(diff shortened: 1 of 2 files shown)


iamkun:
Not sure if we need to mention this.

As to me, yarn is always available if the package is published to npm.

shaminmeerankutty:
When I usually came up with a library I do copy paste the installation script if I am using yarn installing the library with npm it causes some conflict. So I thought it will be good if there is also an installation script available for yarn as well in the readme, because it will be really useful for people who uses yarn as their package manager.

iamkun:
seems we don't need this. Fell free to re-open it if you have any question.

PR #388, opened 2018-10-18
Added Japanese weekdaysShort and weekdaysMin

Liooo:
Added Japanese weekdaysShort and weekdaysMin

### DIFF
--- a/src/locale/ja.js
+++ b/src/locale/ja.js
@@ -3,7 +3,7 @@ import dayjs from 'dayjs'
 const locale = {
   name: 'ja',
   weekdays: '日曜日_月曜日_火曜日_水曜日_木曜日_金曜日_土曜日'.split('_'),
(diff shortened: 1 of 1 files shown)


src/locale/ja.js
```diff
@@ -3,6 +3,8 @@ import dayjs from 'dayjs'
 const locale = {
   name: 'ja',
   weekdays: '日曜日_月曜日_火曜日_水曜日_木曜日_金曜日_土曜日'.split('_'),
+  weekdaysShort: '日曜_月曜_火曜_水曜_木曜_金曜_土曜'.split('_'),
```
iamkun: should this be `weekdaysShort : '日_月_火_水_木_金_土'.split('_')`, according to momentjs's locale?

Liooo: Yea if we stick to full compatibility with moment, fixed.

Issue #386, opened 2018-10-16
Diff for months results in negative zero when no difference.


archemiro:
In momentjs I had a months comparison of dates, when the difference is zero months, the result was `0`. 

With dayjs the result is `-0`, which breaks test cases using something like `expect(diff).toBe(0)` or `expect(diff).toEqual(0)`

```js
var moment = require("moment")
var dayjs = require("dayjs")

var momentDiff = moment("2018-01-01").diff("2018-01-01", "months")
var dayjsDiff = dayjs("2018-01-01").diff("2018-01-01", "months")

console.log(momentDiff) // 0
console.log(dayjsDiff) // -0
```

suspectpart:
It also returns `-0` for unit `'years'`, does the fix cover that as well?

iamkun:
fix #390

PR #390, opened 2018-10-19
[Fixing #386] - Added logic to return just 0

sadarshannaiynar:
Fixing #386 

Just a if condition checking whether the difference is 0 and added tests to make sure it isn't broken later.


sadarshannaiynar:
@iamkun Should modify something?

suspectpart:
From an algorithmic point of view, is this a good solution? Extracting the expression

```javascript
const diff = Number(-(wholeMonthDiff + ((b - anchor) / (c ? (anchor - anchor2) : (anchor2 - anchor)))))
```

and performing a `Math.abs()` on the result is fixing a possibly broken value we just created deliberately by putting a sign in front of the diff expression. If you split the expression this way

```javascript
const diff = (wholeMonthDiff + ((b - anchor) / (c ? (anchor - anchor2) : (anchor2 - anchor))))
return Number(-diff)
```

it becomes a little clearer.  We calculate a `diff` which can be either
- exactly `0` if the dates are the same or
- positive if `a < b` or
- negative if `a > b`

Because of dayjs' semantics, we need to swap the sign of `diff`, but only if it represents a non-zero value.

As we have milliseconds as the finest granularity, the smallest possible diff value other than `0` is for two dates with 1 ms difference, so if two dates differ by 1 ms we get either
- `-3.7335722819593786e-10` (`a > b`) or
- `3.7335722819593786e-10` (`a < b`)

Which means `diff` can never be `-0`, just `0` or a positive or negative float. So the cleaner fix to swap the sign on any non-zero value but return a zero for both negative and positive zero IMHO would be:

```javascript
const monthDiff = (a, b) => {
  // function from moment.js in order to keep the same result
  const anchor = a.clone().add(wholeMonthDiff, 'months')
  const c = b - anchor < 0
  const anchor2 = a.clone().add(wholeMonthDiff + (c ? -1 : 1), 'months')
  const d = (wholeMonthDiff + ((b - anchor) / (c ? (anchor - anchor2) : (anchor2 - anchor))))
  return Number(-d || 0)
}
```
or even inlining it


```javascript
  /// ...
  return Number(-(wholeMonthDiff + ((b - anchor) / (c ? (anchor - anchor2) : (anchor2 - anchor))) || 0)
}
```

This works as `-0` and `0` both evaluate as 'falsy', so we do not have to mess with `Math.abs()`.

suspectpart:
@iamkun @xxyuk What about this issue? Shoule we address it soon? Tests start turning red if you test on diffs with a newer version of node.js' `assert` lib, as they now do `Object.is(...)` in the implementation of `assert.strictEqual(...)`, and that distinguishes between `0` and `-0`. From an arithmetic point of view, it is not a big deal, but it can be quite hard to figure out what went wrong.

Just changing the return line to `  return Number(-(wholeMonthDiff + ((b - anchor) / (c ? (anchor - anchor2) : (anchor2 - anchor))) || 0)` will fix it, see my comment above. Or you take the more verbose solution proposed by this PR, which is more explicit, but also more code.

xxyuk:
@suspectpart your node version please?

suspectpart:
> @suspectpart your node version please?

It was introduced in [Node v10.0.0](https://github.com/nodejs/node/blob/master/doc/changelogs/CHANGELOG_V10.md#2018-04-24-version-1000-current-jasnell) on 2018-04-24, I am using v11.something.

You can find the PR for the change in https://github.com/nodejs/node/commit/493340f56e05d5eced47669f6f1047c29691d0b7

xxyuk:
little weird, `monthDiff ` function is just copied from moment.js in order to keep the same result

suspectpart:
> little weird, `monthDiff ` function is just copied from moment.js in order to keep the same result

Yes, and they have the `|| 0` appended to the end that I proposed in my comment, see here

https://github.com/moment/moment/blob/develop/src/lib/moment/diff.js#L39

So whoever copied it accidentally left out the default to `0`.

sadarshannaiynar:
Sure will update this PR by today

iamkun:
Look great.

Think it's my copy & paste mistake.

Thanks @sadarshannaiynar @suspectpart

iamkun:
hi @sadarshannaiynar, any update here?

suspectpart:
> hi @sadarshannaiynar, any update here?

This is really just 4 characters of code and one additional test case. I could issue a pull request, if you can't find the time.

sadarshannaiynar:
@iamkun Hey sorry will raise it in sometime

sadarshannaiynar:
@iamkun Deeply regretting the delay I have caused. I have changed the return statement as suggested.

iamkun:
@sadarshannaiynar Great thanks.

suspectpart:
> @iamkun Deeply regretting the delay I have caused. I have changed the return statement as suggested.

Nevermind, great you fixed it, thanks!

PR #393, opened 2018-10-25
Fix .diff() UnitType for Typescript.

biigpongsatorn:
```
export type UnitType = 'millisecond' | 'second' | 'minute' | 'hour' | 'day' | 'week' | 'month' | 'quarter' | 'year' | 'date'
```


xxyuk:
Good point.

It would be nice if you could update all the code in docs in different languages in your commit.

biigpongsatorn:
@xxyuk Done :)

PR #396, opened 2018-10-26
Convert switch statements to object literal and size reduction

kingstenbanh:
### Convert switch statement
I'm a big fan of `switch` statement since it is much cleaner when checking multiple scenarios. In the use case of `dayjs`, we want to keep the library as light as possible while still maintaining good development experience. A few switch statements can be converted to object literal to reduce a lot of `case`s and `break`s in the code while the results are still legible. As a result, the file sizes are

| Name  | Before (B)  | After (B)  | Change (%)   |
|---|---|---|---|
| src/index.js   | 10,472  |  9,321  | 11%  |
| dayjs.min.js  | 6,701  | 6,170  |  8% | 

### Destructure multiple property access
1. `this.$d` is used in repeatedly so it will save space by destructuring the `$d` in a few cases
2. Repeating method access
```js
// Before
isStartOf ? argumentStart.slice(slice) : argumentEnd.slice(slice)

// After
(isStartOf ? argumentStart : argumentEnd).slice(slice)
```

3. The method in the `buddhistEra/index.js` has a lot of thing in common and a few differences. 
```js
// Before
const result = str.replace(/BBBB|BB/g, (match) => {
  switch (match) {
    case 'BB':
      return utils.padStart(String(this.$y + yearBias).slice(-2), 2, '0')
    default: // BBBB
      return utils.padStart(String(this.$y + yearBias), 4, '0')
  }
})

// After
const result = str.replace(/BBBB|BB/g, (match) => {
  const year = String(this.$y + yearBias)
  const args = match === 'BB' ? [year.slice(-2), 2] : [year, 4]
  return padStart(...args, '0')
})


kingstenbanh:
@iamkun: Do you have any feedbacks for the changes?

PR #404, opened 2018-11-06
Add plugin advancedCompare

suspectpart:
We switched from moment.js to dayjs on a project that made heavy use of moment.js, especially comparisons considering units as well. Thought I'll share this a plugin. All tests taken from moment.js and adjusted to run with the dayjs test suite.


src/plugin/advancedCompare/index.js
```diff
@@ -0,0 +1,40 @@
...
+  const oldSame = proto.isSame
+  const oldIsAfter = proto.isAfter
+  const oldIsBefore = proto.isBefore
+
+  const v = (l, r) => l.isValid() && r.isValid()
+
+  dayjsClass.prototype.isSame = function (o, u) {
+    if (!v(this, o)) return false
+
+    return (u && u === 'millisecond')
```
iamkun: better use constant

src/plugin/advancedCompare/index.js
```diff
@@ -0,0 +1,40 @@
...
+      ? oldSame.bind(this)(o)
+      : this.startOf(u).valueOf() <= o && o <= this.endOf(u).valueOf()
+  }
+
+  dayjsClass.prototype.isAfter = function (o, u) {
```
iamkun: use `this.$utils().prettyUnit(u)`  to deal with 'ms', 'millisecond' and 'milliseconds'

src/plugin/advancedCompare/index.js
```diff
@@ -0,0 +1,42 @@
...
+
+    return (u && this.$utils().prettyUnit(u) === C.MS)
+      ? oldSame.bind(this)(o)
+      : this.startOf(u).valueOf() <= o && o <= this.endOf(u).valueOf()
+  }
+
+  dayjsClass.prototype.isAfter = function (o, u) {
+    if (!v(this, o)) return false
+
+    return (u && this.$utils().prettyUnit(u) === C.MS)
```
xxyuk: why not `const prettyUnit = this.$utils().prettyUnit` to decrease bundle size

src/plugin/advancedCompare/index.js
```diff
@@ -0,0 +1,42 @@
...
+      ? oldIsAfter.bind(this)(o)
+      : o.valueOf() < this.startOf(u).valueOf()
+  }
+
+  dayjsClass.prototype.isBefore = function (o, u) {
+    if (!v(this, o)) return false
+
+    return (u && this.$utils().prettyUnit(u) === C.MS)
```
xxyuk: and seems we don't need `u &&` scince prettyUnit has already done this check

src/plugin/advancedCompare/index.js
iamkun: nice

src/plugin/advancedCompare/index.js
suspectpart: You are right, that check was redundant. Fixed with bba77e4e246effc0070fb5fb6ae4f8b8f8e30f96.

src/plugin/advancedCompare/index.js
suspectpart: Good point. Didn't work exactly like that as `this.$utils()` is not bound in the outer context properly, but `proto.$utils()` works. Fixed with bba77e4e246effc0070fb5fb6ae4f8b8f8e30f96.

src/plugin/advancedCompare/index.js
suspectpart: Right, my bad. Fixed with bba77e4e246effc0070fb5fb6ae4f8b8f8e30f96.

src/plugin/advancedCompare/index.js
suspectpart: Fixed with bba77e4e246effc0070fb5fb6ae4f8b8f8e30f96.

src/plugin/advancedCompare/index.js
```diff
@@ -0,0 +1,43 @@
...
+
+    return pretty(u) === C.MS
+      ? oldIsAfter.bind(this)(o)
+      : o.valueOf() < this.startOf(u).valueOf()
+  }
+
+  proto.isBefore = function (o, u) {
+    if (!v(this, o)) return false
+
+    return pretty(u) === C.MS
```
xxyuk: seems we could compare this directly, just `return this.endOf(u).valueOf() < o.valueOf()`, since 'ms' is the default value of `endOf` and `startOf`

suspectpart: Good point, that leaves us with just

```javascript
proto.isSame = function (o, u) {
    return v(this, o) && this.startOf(u).valueOf() <= o && o <= this.endOf(u).valueOf()
  }

  proto.isAfter = function (o, u) {
    return v(this, o) && o.valueOf() < this.startOf(u).valueOf()
  }

  proto.isBefore = function (o, u) {
    return v(this, o) && this.endOf(u).valueOf() < o.valueOf()
  }
```

Tests are green, should I push that?

suspectpart: Integrated your suggestions with a101e3e437724f7498c8f651fd90061c35606389.

iamkun: Looks great. 

We might could modify `isSame`  `isAfter ` and `isBefore` in our main bundle instead of a separate plugin. 

And leave the other two as two individual plugins like `isBetween` plugin. 

How do you think?

suspectpart: Sounds great to me. I wasn't sure if you consider these comparisons to be important enough to be in the main bundle, but given how you helped me reduce the implementation to just oneliners, I think this brings great benefits with relatively low costs in terms of size.

What would be a proper plugin name for the remaining two methods `isSameOrAfter()` and `isSameOrBefore()`? Do you want me to do the changes or will you do them?

iamkun: If we skip the validation check, seems it won't increase bundle size too much (or the same maybe). 

Just have a try if you have spare time, and let us know the size of the modified main bundle.

We better use API name as its plugin name as well as file name.

iamkun: Yes, and two separate plugins might be a good choice I think

suspectpart: Alright, so I am going to

- swap `isBefore(other)`, `isAfter(other)` and `isSame(other)` with `isBefore(other, unit?)`, `isAfter(other, unit?)` and `isSame(other, unit?)` in the main bundle
- integrate the test suite into the main bundle test suite (should it be in a separate file? it is a rather large body of tests on its own...)
- add a plugin `isSameOrBefore` named `is-same-or-before.js`
- add a plugin `isSameOrAfter` named `is-same-or-after.js`

Agree?

iamkun: Agreed,

I like your large test suite, and better keep it as a separate file.

Plugin file name use camel-like case like the other.

Plus, we might need to update `isBetween` plugin as well.

Also our `index.d.ts` file.

suspectpart: - [ ] Move `isSame(other, units)`, `isAfter(other, units)`, `isBefore(other, units)` to main bundle
- [ ] Adjust tests to run against the main bundle
- [ ] Move `isSameOrAfter(other, units)` to `plugin/isSameOrAfter.js`
- [ ] Move `isSameOrBefore(other, units)` to `plugin/isSameOBefore.js`
- [ ] Adjust `isBetween()` to also support units
- [ ] Adjust Typescript typings in `index.d.ts` to reflect the changes

suspectpart: Apart from `isBetween`, all tasks are done as of c13c93d280f24c970db5e72752156fc1a4d5e7b2. Please check that I did not miss anything with Typescript typings, not too sure about that one, especially the tests for it.

Size of the main bundle is now:

```bash
> size-limit && gzip-size dayjs.min.js

  Package size: 2.37 KB
  Size limit:   2.99 KB
  With all dependencies, minified and gzipped

2.57 kB
```

suspectpart: `isBetween()` added with d0eeb32. Please do proper review, those are bigger changes. If no objections from your side, that should be it.

Liooo:
such a hot topic for me, is this going to be merged and released soon?

index.d.ts
```diff
@@ -18,68 +18,68 @@ declare namespace dayjs {
     seconds: number
     milliseconds: number
   }
-
```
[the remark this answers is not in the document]
suspectpart: I guess reformatting just removed unnecessary whitespace?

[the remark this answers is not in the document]
xxyuk: think you might could revert these changes, and create a new PR with related reformatting changes.

and it's easy to review either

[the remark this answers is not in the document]
suspectpart: Agreed, I will do that tonight. Are we good to merge then?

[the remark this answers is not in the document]
suspectpart: I am trying to revert those whitespaces, but webstorm won't let me. Actually, I see no reason to do it anyway as blank linens with whitespaces shouldn't be in there anyway according to linting (which is just not checked by `npm lint` due to the `index.d.ts` file being Typescript). Do you really want me to force this back to whitespaces? I think it overcomplicates things. If you finally decide you want that reverted, I will do, but I think it is rather not worth it.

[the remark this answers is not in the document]
suspectpart: So I'll resolve this discussion and we are good to go?

suspectpart:
@xxyuk @iamkun Can we merge this soon? It just got outdated with master (again). What is the status on your side?

suspectpart:
> Thanks. There's a little problem with our npm account. And I'll merge this into next release after I get the account fixed.

Alright, thanks for the update. Can you make any projection on how long that might take? Is it days, weeks or months?

iamkun:
Not sure, but as soon as possible.

xxyuk:
great

suspectpart:
Thanks a lot, great support and discussion with all of you!

PR #407, opened 2018-11-10
refactor(style): make the code less and more convenient to maintain

BlackHole1:
1. use object instead of the switch


BlackHole1:
Unit tests have all run through

iamkun:
duplicate #396

Issue #364, opened 2018-10-03
Add plugins methods to typescript definition


naulacambra:
Should plugins's api methods be added in the core `d.ts` file? Or they should have their owns?

If they have their own, they should be merge into one when bundling the package?

I'm trying to use this library at another library, which is written with Typescript, and trying to use plugin methods, it's driving me crazy. Any thoughts?

suspectpart:
In my opinion it would be the proper approach to place `d.ts` definitions in the plugin's folder as a plugin should be self-contained. By keeping plugins self-contained it is easy to add or remove plugins by just adding / removing a folder in `plugin/`, making sure that a plugin does not alter the main bundle in anyway. I would even suggest to place tests right with the plugin, so every plugin has a structure like this:

```
plugin
└── somePlugin
    ├── index.js
    ├── index.d.ts
    └── somePlugin.test.js
```
That could be achieved by altering the jest config to something like

```javascript
"jest": {
    "roots": [
      "test",
      "src/plugin"
    ],
    "testRegex": "(.*?/)?.*test.js$",
    "testURL": "http://localhost",
    "coverageDirectory": "./coverage/",
    "collectCoverage": true,
    "collectCoverageFrom": [
      "src/**/*"
    ]
  }
```

iamkun:
@suspectpart  Seems good.

suspectpart:
Should I restructure the plugins that way and make a pull request so we can have the discussion based on the changes?

iamkun:
We have to consider structure of our release code.

Currently:

plugin
└── advancedFormat.js
└── isBetween.js

And I'm not sure if code editor could support  typescript definition file like this:
plugin
└── advancedFormat.js
└── advancedFormat.d.ts
└── isBetween.js
└── isBetween.d.ts

suspectpart:
I guess it works by making the compiled `plugin/` folder look like

```
plugin
└── isBetween
    ├── index.js
    └── index.d.ts
```

i.e. compiling plugin code into separate folders and copying the corresponding `index.d.ts` to reside next to the build artifact. Type definitions in those `index.d.ts` files need to extend the `Dayjs` class by doing something like this

```typescript
declare namespace dayjs {
  interface Dayjs {
    isBetween(d1: dayjs.Dayjs, d2: dayjs.Dayjs)
  }
}
```
I quickly sketched this and could get Typescript to not complain, but I had no time to create a full working prototype.

suspectpart:
[...]

PR #418, opened 2018-11-29
Add plugin type definitions

### LINKED ISSUE #297: Property 'fromNow' does not exist on type 'Dayjs'
I'm using the `relativeTime` plugin in my `.ts` file

Importing is done in below manner : -
`import * as dayjs from 'dayjs';`
`import relativeTime from 'dayjs/plugin/relativeTime';`
`dayjs.extend(relativeTime);`

I've also tried to change the import statement as:
`import dayjs from 'dayjs';` But started throwing new error as `no exported member`

Now when I'm using in the `fromNow` function as: -
`dayjs(this.created_at).fromNow();` the compiler is throwing an error 

**` Property 'fromNow' does not exist on type 'Dayjs'`**
 I'm using  `"dayjs": "^1.7.4",` version

What's the with usage? How can I fix this?

@iamkun Please address this issue

ypresto:
- Place all type definition files in `types/` dir and merge into root of package on build.
- Fix lack of `string | number | Date` in arguments
- Add argument types to locale() methods and fix plugin option type
- Added type definitions of plugins
- Bumped rollup but still have `@types/***` in its `dependencies`, causes `tsc` compile error without workaround...

Fixes #297


yashwp:
@ypresto Thanks, bro :) Anyways I've rolled back to `moment.js`

iamkun:
Hi @ypresto , we are working on this PR. And there is a discussion on #364. 

Join us please if you have time, so that we can see what we can do to make this pr get merged.

types/index.d.ts
```diff
@@ -1,14 +1,14 @@
 export = dayjs;
 
-declare function dayjs (config?: dayjs.ConfigType, option?: dayjs.OptionType): dayjs.Dayjs
+declare function dayjs (date?: dayjs.DateType, option?: dayjs.OptionType): dayjs.Dayjs
 
 declare namespace dayjs {
-  export type ConfigType = string | number | Date | Dayjs
+  export type DateType = string | number | Date | Dayjs
```
dmk1111: this could be a breaking change for those, who use TS and rely on semver. I suppose it is better to increase at least minor version, or to not change naming right now at all, but leave comments that it is deprecated and will be replaced in newer versions

ypresto: Added type alias with `@deprecated` marker!

dmk1111:
Looks good to me

iamkun:
@ypresto I like this plugin type definition in module augmentation。
Come and join me here, if you have time.

https://gitter.im/iamkun/dayjs

ypresto:
I added comment here :) https://github.com/iamkun/dayjs/issues/364#issuecomment-465923460

ypresto:
Fixed missing OptionType signature, especially for `dayjs('05/02/69 1:02:03 PM -05:00', 'MM/DD/YY H:mm:ss A Z')`.

PR #421, opened 2018-12-06
Extendable wrapper function

farnabaz:
if we call `Utils.wrapper` instead of `wrapper` we gave plugins a great capability to overwrite it and pass additional data to dayjs constructor
I had a plugin called jalaliday that add jalali calendar support for dayjs, and I need to keep `calendar` on date manipulation just like we keep locale in `wrapper`
```js
// original
const wrapper = (date, instance) => dayjs(date, { locale: instance.$L })
// what I need
const wrapper = (date, instance) => dayjs(date, { locale: instance.$L, calendar: instance.$C })
```
with this changes I could overwrite `Dayjs.proto.$utils().wrapper` and achieve my goal, without this there is no way to keep calendar in date manipulation like calling `startOf` method


iamkun:
More detail please. I still don't understand why doing this. `Utils.wrapper`

farnabaz:
In current version of dayjs there is no way for plugins to keep their properties on object clone, because cloning dayjs is done in `wrapper` function that isn't accessible for plugins
if we use `Utils.wrapper` we can overwrite this function in dayjs plugins and  add more logic to this function
for example if my plugin I could overwrite this function like this:
```js
const myCustomWrapper = (date, instance) => dayjs(date, { locale: instance.$L, calendar: instance.$C })
Dayjs.proto.$utils().wrapper = myCustomWrapper

const oldInit = Dayjs.proto.init
Dayjs.proto.init = function (cfg = {}) {
    oldInit.bind(this)(cfg)

    this.$C = cfg.calendar || 'default value'
}
```
this way I could keep value of property `$C`

also here is a simple simulation of two senario
```js
const wrapper = () => console.log('original wrapper')
const utils = {
	wrapper: wrapper
}
function $utils() {
	return utils;
}
// change utils wrapper on plugin
$utils().wrapper = () => console.log('custom wrapper')

// dayjs clone
wrapper() // calls original wrapper

utils.wrapper() // calls custom wrapper (what I want)
```

iamkun:
Seems we only need to update the `clone` API to meet your need.

Or we may could looking for a new way of overriding our `init`  function.

PR #452, opened 2019-01-15
Added deltaZone in the equation when calculating diff

### LINKED ISSUE #384: diff() in "days" broken after making add() "days" aware of daylight-saving changes
Computing difference between two dates, which one was subtracted 1000 days from, returns 999 days, if the date is on the boundary of daylight-saving changes:

```
> d1 = dayjs('2018-10-28')
2018-10-27T22:00:00.000Z
> d2 = dayjs('2018-10-28').add(-1000, 'days')
2016-01-31T23:00:00.000Z
> d1.diff(d2, 'hours')
23999
> d1.diff(d2, 'days')
999
```

Moment.js computes the difference correctly. Notice, that the difference in hours is correct in Day.js too:

```
> m1 = moment('2018-10-28')
moment("2018-10-28T00:00:00.000")
> m1.toDate()
2018-10-27T22:00:00.000Z
> m2 = moment('2018-10-28').add(-1000, 'day')
moment("2016-02-01T00:00:00.000")
> m2.toDate()
2016-01-31T23:00:00.000Z
> m1.diff(m2, 'hours')
23999
> m1.diff(m2, 'days')
1000
```

This problem has been caused by 969acedc94461dca2b1d9313299a2c9db4aed7a9. It adds/subtracts days using the local time zone. Previously, days were added/subtracted in UTC (using UNIX time), like hours and shorter units. While it helps people, who were not aware of Day.js working in the local time zone (#262 and #329), it also made Day.js inconsistent, because computing difference in days is still done in UTC (using UNIX time).

Once `add()` and `subtract()` methods were modified to expect the day count in the local time zone, the `diff()` method should be modified to return the day count in the local time zone too.

Adding/subtracting hours and shorter units happens in UTC, computing the difference in these units happens in UTC as well. Handling hours and shorter units by Day.js is consistent.

The tests were executed in Node.js 8 in the time zone "Europe/Prague" using Day.js 1.7.7 and Moment.js 2.22.2.

naulacambra:
I've looked into `momentjs` code, and tried to understand why method `dayjs().diff()` was returning different values than `moment().diff()` when the unit was `day`. Happens that `momentjs` uses the offset of each element to calculate hour zone difference, and adds it to the equation.

I've had to add the `utcOffset()` method in the core (where the `diff()` method is), but only the `get` version.

This should resolve #384

### DIFF
--- a/src/constant.js
+++ b/src/constant.js
@@ -34,4 +34,3 @@ export const en = {
   weekdays: 'Sunday_Monday_Tuesday_Wednesday_Thursday_Friday_Saturday'.split('_'),
   months: 'January_February_March_April_May_June_July_August_September_October_November_December'.split('_')
 }
-
--- a/src/index.js
+++ b/src/index.js
@@ -212,7 +212,6 @@ class Dayjs {
     return this
   }
 
-
   set(string, int) {
     return this.clone().$set(string, int)
   }
@@ -303,9 +302,14 @@ class Dayjs {
     })
   }
 
+  utcOffset() {
+    return -Math.round(this.$d.getTimezoneOffset() / 15) * 15
+  }
+
   diff(input, units, float) {
     const unit = Utils.prettyUnit(units)
     const that = dayjs(input)
+    const zoneDelta = (that.utcOffset() - this.utcOffset()) * C.MILLISECONDS_A_MINUTE
     const diff = this - that
     let result = Utils.monthDiff(this, that)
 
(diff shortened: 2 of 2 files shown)


iamkun:
Oh I see, nice catch. 

Besides, I think we could add `utcOffset` as a public api to our user.

Docs and test should be included. 

How do you think? Or I could make a pr if I got some time.

iamkun:
PS, could you please re-open a new pr to merge your code to our dev branch.

It might good to have a dev branch waiting for next release.

naulacambra:
> Besides, I think we could add utcOffset as a public api to our user.

Sure, I'll add the `utcOffset` documentation and tests

> PS, could you please re-open a new pr to merge your code to our dev branch.

Ok, i'll close this PR and open a new one

iamkun:
Thanks

PR #453, opened 2019-01-16
Added deltaZone in the equation when calculating diff and added utcOffset api method

### LINKED ISSUE #384: diff() in "days" broken after making add() "days" aware of daylight-saving changes

naulacambra:
I've looked into momentjs code, and tried to understand why method dayjs().diff() was returning different values than moment().diff() when the unit was day. Happens that momentjs uses the offset of each element to calculate hour zone difference, and adds it to the equation.

I've had to add the utcOffset() method in the core (where the diff() method is), but only the get version.

This should resolve #384

As mentioned at #452 I've added test and documentation for `utcOffset` method

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -212,7 +212,6 @@ class Dayjs {
     return this
   }
 
-
   set(string, int) {
     return this.clone().$set(string, int)
   }
@@ -303,9 +302,14 @@ class Dayjs {
     })
   }
 
+  utcOffset() {
+    return -Math.round(this.$d.getTimezoneOffset() / 15) * 15
+  }
+
   diff(input, units, float) {
     const unit = Utils.prettyUnit(units)
     const that = dayjs(input)
+    const zoneDelta = (that.utcOffset() - this.utcOffset()) * C.MILLISECONDS_A_MINUTE
     const diff = this - that
     let result = Utils.monthDiff(this, that)
 
@@ -313,8 +317,8 @@ class Dayjs {
       [C.Y]: result / 12,
       [C.M]: result,
       [C.Q]: result / 3,
-      [C.W]: diff / C.MILLISECONDS_A_WEEK,
-      [C.D]: diff / C.MILLISECONDS_A_DAY,
+      [C.W]: (diff - zoneDelta) / C.MILLISECONDS_A_WEEK,
+      [C.D]: (diff - zoneDelta) / C.MILLISECONDS_A_DAY,
       [C.H]: diff / C.MILLISECONDS_A_HOUR,
       [C.MIN]: diff / C.MILLISECONDS_A_MINUTE,
       [C.S]: diff / C.MILLISECONDS_A_SECOND
--- a/src/index.js
+++ b/src/index.js
@@ -303,6 +303,8 @@ class Dayjs {
   }
 
   utcOffset() {
+    // Because a bug at FF24, we're rounding the timezone offset around 15 minutes
+    // https://github.com/moment/moment/pull/1871
     return -Math.round(this.$d.getTimezoneOffset() / 15) * 15
   }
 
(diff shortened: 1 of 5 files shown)


iamkun:
Cool

src/index.js
```diff
@@ -303,18 +302,23 @@ class Dayjs {
     })
   }
 
+  utcOffset() {
+    return -Math.round(this.$d.getTimezoneOffset() / 15) * 15
```
iamkun: why `15` here, can this be any other number?

naulacambra: In the `momentjs` code, there was this comment

> // On Firefox.24 Date#getTimezoneOffset returns a floating point.
    // https://github.com/moment/moment/pull/1871

It seems there was a bug with FF24, and in order to avoid issues with the floating point, they round it around 15, which are minutes.

I'm not fully sure about how this solves a possible bug, but `momentjs` is fully tested, and I think we can trust its code. Maybe we should add some sort of comment?

iamkun: I see, thanks. And add some comments please.

src/index.js
```diff
@@ -303,18 +302,25 @@ class Dayjs {
...
     let result = Utils.monthDiff(this, that)
 
     result = {
       [C.Y]: result / 12,
       [C.M]: result,
       [C.Q]: result / 3,
-      [C.W]: diff / C.MILLISECONDS_A_WEEK,
-      [C.D]: diff / C.MILLISECONDS_A_DAY,
+      [C.W]: (diff - zoneDelta) / C.MILLISECONDS_A_WEEK,
+      [C.D]: (diff - zoneDelta) / C.MILLISECONDS_A_DAY,
```
iamkun: Why we just need zoneDelta only in week and day? 
Could this change added to  `diff` in line 315?
```
const diff = this - that - zoneDelta
```

naulacambra: I'm not sure 😅 but substracting the zoneDelta to other units than `week` and `day`, makes the tests broke.

Maybe because units under `day` do not consider daysaving?

PR #454, opened 2019-01-16
Created dayOfYear plugin

naulacambra:
Created `dayOfYear` plugin, continuing the `ofYear` plugin series.

This plugin adds method 'dayOfYear()'  which return the day of the year of the current object.

This PR goes after #453 because `diff()` was needed to work properly to pass all tests.


iamkun:
THX

PR #455, opened 2019-01-16
Created DaysInYear plugin

naulacambra:
Created daysInYear plugin, which adds `daysInYear()` API method to retrieve how many days has the `dayjs` instance year.


iamkun:
It's not that useful I think. Pending.

naulacambra:
> It's not that useful I think. Pending.

It is not 😅  but there is this lib, [Dayspan Vuetify](https://github.com/ClickerMonkey/dayspan-vuetify) which relies on `momentjs`. I want to use this lib, and at the same time, keep `dayjs` as a `momentjs` substitute. 

Using a `webpack` alias, I can "deceive" `dayspan-vuetify` to load `dayjs` instead of `momentjs`, but there are some functions that are needed to work, thats why I'm creating this plugins, since I've already made the code for my project.

iamkun:
Agreed that we should add the missing functions to dayjs  (core / plugin ).

But it seems momentjs does not have `DaysInYear` either?

naulacambra:
> But it seems momentjs does not have DaysInYear either?

My bad, you're right. It's a method needed for `weeksInYear` (plugin I'll create after we resolve #456), and I blindly created this. 

Closing this PR

PR #441, opened 2019-01-04
perf(format): reuse matches instead of created when replacing

whatwewant:
as title saying. @iamkun

PR #457, opened 2019-01-16
Parse month from string in customParseFormat

Kreozot:
to handle 'MMM' and 'MMMM' formats (February, Feb, etc)


babel.config.js
```diff
@@ -1,20 +1,20 @@
...
-          {
-            "modules": false,
-            "loose": true
-          }
-        ]
-      ]
-    }
-  }
-}
+module.exports = {
```
Kreozot: Otherwise it causes tests compilation on Windows

src/index.js
```diff
@@ -64,15 +64,16 @@ const parseDate = (date) => {
 
 class Dayjs {
   constructor(cfg) {
+    this.$L = this.$L || parseLocale(cfg.locale, null, true) || L
```
Kreozot: to make locale available in parse() function

babel.config.js
iamkun: why we have to change file name && code here?

Kreozot: It's strange, but babel v7+ doesn't work with .babelrc on Windows (I tested it on Windows 10) - only with babel.config.js. JS files just stops transpiling and all tests fails because of syntax errors on import statements. You can test this configuration - it is completely same.

Kreozot:
@iamkun Is it ok now?

iamkun:
I'll test `babel.config.js` change if I got some time first.

src/plugin/customParseFormat/index.js
```diff
@@ -80,7 +43,55 @@ function correctHours(time) {
   }
 }
 
-function makeParser(format) {
+function makeParser(format, instance) {
+  const expressions = {
```
iamkun: @Kreozot 

Everything else looks fine in this PR, and will be merged into next release.

Please move this object to outer scope for a better performance.

ref #441

iamkun:
`babel.config.js` tested on windows, works good.

src/plugin/customParseFormat/index.js
Kreozot: This object contains two properties that depends of current instance locale.
But I can move everything but this two properties and then merge it together. Is it ok?

Kreozot: Something like this:
```
  const expressions = {
    ...baseExpressions,
    MMM: [matchWord, function (input) {
      const locale = instance.$locale()
      const { months } = locale
      this.month = months.findIndex(month => month.substr(0, 3) === input) + 1
    }],
    MMMM: [matchWord, function (input) {
      const locale = instance.$locale()
      const { months } = locale
      this.month = months.indexOf(input) + 1
    }]
  }
```

src/plugin/customParseFormat/index.js
```diff
@@ -80,7 +43,55 @@ function correctHours(time) {
...
+    h: [match1to2, addInput('hours')],
+    HH: [match2, addInput('hours')],
+    hh: [match2, addInput('hours')],
+    D: [match1to2, addInput('day')],
+    DD: [match2, addInput('day')],
+    M: [match1to2, addInput('month')],
+    MM: [match2, addInput('month')],
+    MMM: [matchWord, function (input) {
+      const locale = instance.$locale()
+      const { months } = locale
```
iamkun: also consider `monthsShort`  in some locales

src/plugin/customParseFormat/index.js
iamkun: It's better to get `this.locale` before `makeParser ` called, and set it to a variable in the outer scope.

src/plugin/customParseFormat/index.js
Kreozot: I just started to doubt: is there a way to change locale before date parsing? According to documentation, the only way to parse is in the constructor. Am I missing something?

iamkun: why `change locale before date parsing`?

src/plugin/customParseFormat/index.js
Kreozot: Don't you think, it is a little bit dirty? For now the only way to change locale is to set it globally. But if it will be changed in future then global variable may cause failure.

src/plugin/customParseFormat/index.js
```diff
@@ -80,7 +81,31 @@ function correctHours(time) {
...
+  const expressions = {
+    ...baseExpressions,
+    MMM: [matchWord, function (input) {
+      const locale = instance.$locale()
+      const { months, monthsShort } = locale
+      const matchIndex = monthsShort
+        ? monthsShort.findIndex(month => month === input)
+        : months.findIndex(month => month.substr(0, 3) === input)
+      if (matchIndex < 0) {
+        throw new Error(`Failed to parse "${input}" as MMM`)
```
iamkun: no error, just skip it like others

Kreozot: done

Kreozot: But suddenly it broke the tests on "Invalid Date" after parse corrupted string

src/plugin/customParseFormat/index.js
```diff
@@ -9,6 +9,7 @@ const matchUpperCaseAMPM = /[AP]M/
 const matchLowerCaseAMPM = /[ap]m/
 const matchSigned = /[+-]?\d+/ // -inf - inf
 const matchOffset = /[+-]\d\d:?\d\d/ // +00:00 -00:00 +0000 or -0000
+const matchWord = /[^\d\s]+/ // Word
```
iamkun: let locale;

src/plugin/customParseFormat/index.js
```diff
@@ -114,9 +133,9 @@ function makeParser(format) {
   }
 }
 
-const parseFormattedInput = (input, format) => {
+const parseFormattedInput = (input, format, instance) => {
   try {
-    const parser = makeParser(format)
+    const parser = makeParser(format, instance)
```
iamkun: locale = instance.$locale()

src/plugin/customParseFormat/index.js
```diff
@@ -80,7 +81,25 @@ function correctHours(time) {
   }
 }
 
-function makeParser(format) {
+function makeParser(format, instance) {
+  const expressions = {
+    ...baseExpressions,
+    MMM: [matchWord, function (input) {
+      const locale = instance.$locale()
+      const { months, monthsShort } = locale
```
iamkun: and move these out of this function,
```
 MMM: [matchWord, function (input) {
      const { months, monthsShort } = locale
```

src/plugin/customParseFormat/index.js
Kreozot: got it

src/plugin/customParseFormat/index.js
Kreozot: So, am I should to bring errors back?

test/plugin/customParseFormat.test.js
```diff
@@ -72,3 +75,47 @@ it('fails with an invalid format', () => {
...
+  const inputUk = '2018 трав 03'
+  const format = 'YYYY MMM DD'
+  dayjs.locale(uk)
+  expect(dayjs(inputUk, format).valueOf()).toBe(moment(input, format).valueOf())
+})
+
+it('return Invalid Date when parse corrupt string', () => {
+  const input = '2018 Februaru 03'
+  const format = 'YYYY MMMM DD'
+  expect(dayjs(input, format).format()).toBe('Invalid Date')
```
iamkun: where's `'Invalid Date'`?

Kreozot: What do you mean?

Kreozot: As a user I expect to know if something goes wrong when parsing my date string.
In first iteration I raised Error in parsing function in result was "Invalid Date". But now it returns correct date (maybe it previous value). So I consider to bring back errors raising.

Kreozot: Take a look at test details
https://travis-ci.org/iamkun/dayjs/jobs/495012516#L543

iamkun: May be you are right, let's throw this error 😬

Kreozot: Looks like this done at last :)

test/plugin/customParseFormat.test.js
```diff
@@ -72,3 +75,47 @@ it('fails with an invalid format', () => {
...
+  const format = 'YYYY MMM DD'
+  expect(dayjs(input, format).valueOf()).toBe(moment(input, format).valueOf())
+})
+
+it('parse month from short string in other locales', () => {
+  const input = '2018 May 03'
+  const inputUk = '2018 трав 03'
+  const format = 'YYYY MMM DD'
+  dayjs.locale(uk)
+  expect(dayjs(inputUk, format).valueOf()).toBe(moment(input, format).valueOf())
```
iamkun: this is not a good test .

we should compare with 
```
moment.locale('uk')
```

test/plugin/customParseFormat.test.js
```diff
@@ -72,3 +75,47 @@ it('fails with an invalid format', () => {
...
+})
+
+it('return Invalid Date when parse corrupt string', () => {
+  const input = '2018 Februaru 03'
```
iamkun: I don't know why `moment('2018 Februaru 03', 'YYYY MMMM DD')` returns a valid date. 

But we might should change `Februaru` to another error string

Kreozot: done

test/plugin/customParseFormat.test.js
Kreozot: done.
I also write test with passing locale in config (btw, it is undocumented for now)

Kreozot:
@iamkun Take a look, please!
And I also added case with passing locale in config to docs.

PR #458, opened 2019-01-18
fix: fix it locale error

iamkun:
@anbi can you review this pr please?

### DIFF
--- a/src/locale/es-do.js
+++ b/src/locale/es-do.js
@@ -1,9 +1,12 @@
 import dayjs from 'dayjs'
 
 const locale = {
-  name: 'es-es',
-  weekdays: 'Domingo_Lunes_Martes_Miércoles_Jueves_Viernes_Sábado'.split('_'),
+  name: 'es-do',
+  weekdays: 'domingo_lunes_martes_miércoles_jueves_viernes_sábado'.split('_'),
+  weekdaysShort: 'dom._lun._mar._mié._jue._vie._sáb.'.split('_'),
+  weekdaysMin: 'do_lu_ma_mi_ju_vi_sá'.split('_'),
   months: 'Enero_Febrero_Marzo_Abril_Mayo_Junio_Julio_Agosto_Septiembre_Octubre_Noviembre_Diciembre'.split('_'),
+  monthsShort: 'ene_feb_mar_abr_may_jun_jul_ago_sep_oct_nov_dic'.split('_'),
   relativeTime: {
     future: 'en %s',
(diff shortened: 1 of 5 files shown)


anbi:
Done.

iamkun:
@naulacambra Just updated `locale/es.js` according to momentjs. Can you review please.

Plus, after compare to momentjs locale file, there are `es.js`, `es-do.js` and `es-us.js`. But Dayjs has a `es-es.js` locale file. Is it something wrong?

naulacambra:
> @naulacambra Just updated locale/es.js according to momentjs. Can you review please.
If it's ok, I'll update the test file accordingly.

It's ok 😀

> Plus, after compare to momentjs locale file, there are es.js, es-do.js and es-us.js. But Dayjs has a es-es.js locale file. Is it something wrong?

I'll try to explain it correclty. Spanish is spoken worldwildly, but there are some little differences between some regions. `es-es` is the Spain's spanish (from where am I), `es-us` United States's spanish, and `es-do` is the Dominican Republic's spanish. While there are some differences in the vocabulary, the name of months or weekdays, are the same, so for this reason there is no need to have one file for each region.

We should decide which standard we use. Language nomenclature (`es` for all Spanish regions) or language-region nomenclature (`es-es`, `es-us`,...)

iamkun:
Oh I see.
Not sure if I understood you correctly

It seems `es-es`  is equal to `es` (Castilian Spanish), and `es-us`, `es-do`...  has some different.

So that `es-es.js` is the same with `es.js` in dayjs, and we should just delete `es-es` and consider adding `es-us`, `es-do` like moment does?

iamkun:
@naulacambra  any suggestions here for my comment above? thanks https://github.com/iamkun/dayjs/pull/458#issuecomment-456000908

iamkun:
@naulacambra thanks. 

According to your replay, at present, we should just delete `es-es`.

And if we add some more cases in locales that might have different between `es` and `es-us`, we could add `es-do` or `es-us` at that time accordingly.

Right？

naulacambra:
@iamkun exactly 😀

iamkun:
Finally I decided to add `es-do` and `es-us`.
So that can keep the same with moment, on the other hand, this could also increase our total locale count. 😬

naulacambra:
> So that can keep the same with moment, on the other hand,

Ok, makes sense.

> this could also increase our total locale count. 

Since they need to be loaded by the user, I don't increase the "default" size, so I think we're fine with that 😀

PR #465, opened 2019-01-27
feat(): add DynamicLocale plugin

adrai:


src/plugin/dynamicLocale/index.js
```diff
@@ -0,0 +1,14 @@
+function extendLocaleFn(orgFn) {
+  return function (...theArgs) {
+    if (typeof theArgs[0] === 'string' && theArgs.length === 1) {
+      // eslint-disable-next-line no-empty, global-require, import/no-dynamic-require
+      try { require(`../../locale/${theArgs[0]}`) } catch (e) {}
```
adrai: is this ok for you?

iamkun:
well, instead of doing dynamic require, seems we'd better export all locale in one file such as `dayjs/locale/index.js`, so that we could switch locale easily.

albertms10:
Hey, is there any alternative to this in order to dynamically load a locale?

I get an error when I try something like this:
```ts
const selectedLocale = "ca";

import(`dayjs/locale/${selectedLocale}`).then(() => {
  dayjs.locale(selectedLocale);
});
```

```
TypeError: Module specifier, 'dayjs/locale/ca' does not start with "/", "./", or "../".
```

PR #467, opened 2019-01-27
Add prettier

wangyiz4262:
Use Prettier to format source files


iamkun:
Thanks, but I'm not sure if this is a must do change, since the format rules of prettier has a little different with es-lint in .js files. But markdown format looks great.

iamkun:
@wangyiz4262  one more thing, it would be great if you could make a new PR with only prettier added to the project as well as the pre-commit lint or something alike, and we could do prettier during our CI running. A PR 104 files change make it impossible to review.

wangyiz4262:
@iamkun Very happy to. I will get that done tomorrow. Actually it is not necessary to make a new PR. I can force push the changes so that this PR will be updated accordingly.

iamkun:
@wangyiz4262 that would be nice,  take your time please.

wangyiz4262:
Hi @iamkun, I added Prettier into CI but disabled pre-commit because it wouldn't let me commit without updating all the files. Therefore I need you to approve the PR first, and then I will create another PR to re-enable eslint in the pre-commit hook.

wangyiz4262:
@iamkun I just realized this won't work because just adding Prettier to CI will only prettify files on the flight. The real source files still don't change. There has to be a big change contain all the prettified files.

iamkun:
How about we just start from `.md` files. Simply add prettier and a npm script, and that's enough. After that we could consider this should be added to precommit or ci or something else.

wangyiz4262:
@iamkun I added Prettier configuration without hook up to eslint, and also formatted all the markdown files in a separate commit.

iamkun:
Looks good, but better revert the last commit and leave the only change with add prettier. There's no need to review diffs made by prettier.

wangyiz4262:
@iamkun Got you. I have reverted the last commit. The source files can be prettified at any time later.

PR #470, opened 2019-01-31
Added shorthand units to UnitType

dotlouis:
Thanks for this useful library!

When using Typescript, shorthand units were throwing errors because not recognized.

On a side note, I wonder if the same type (UnitType) should be used for `set()` and `add()/subtract()`methods. Because the former does not take a `week` as a unit (which would count as valid in TS typings). Just a thought.

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -7,7 +7,11 @@ declare namespace dayjs {
 
   export type OptionType = { locale: string }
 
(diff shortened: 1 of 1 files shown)


iamkun:
@dotlouis thanks you are right, and advice here? Or just copy to another unitType?

dotlouis:
Well you can make union types to avoid copying code. Something in the likes of:
```ts
type OpUnitType = 'day' | 'hour' // ... all common unit types
type SetUnitType = OpUnitType | 'week' // all additional types
```

I can make PR with the changes if you'd like

PR #476, opened 2019-02-05
D2M

iamkun:


index.d.ts
```diff
@@ -1,6 +1,5 @@
-
+export = dayjs;
 declare function dayjs (config?: dayjs.ConfigType, option?: dayjs.OptionType): dayjs.Dayjs
-export default dayjs
```
anka-213: Why was this reverted? Importing it as `import * as dayjs from 'dayjs'` doesn't seem to work. What is the correct way to import this from typescript?

iamkun: you can check https://day.js.org/docs/en/installation/typescript for installation guide.

anka-213: Yes, I did that, but as I said, they’re incorrect. I should probably open a new issue about that though.

PR #478, opened 2019-02-05
Added en-gb-locale

tcrands:
Added a new locale for en-gb that supports the time and date formats.

### DIFF
--- a/src/locale/en-gb.js
+++ b/src/locale/en-gb.js
@@ -1,0 +1,49 @@
+import dayjs from 'dayjs'
+
+const locale = {
+  name: 'en',
+  weekdays: 'Sunday_Monday_Tuesday_Wednesday_Thursday_Friday_Saturday'.split('_'),
+  weekdaysShort: 'Sun_Mon_Tue_Wed_Thu_Fri_Sat'.split('_'),
+  weekdaysMin: 'Su_Mo_Tu_We_Th_Fr_Sa'.split('_'),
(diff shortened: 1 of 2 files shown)


src/locale/en-gb.js
```diff
@@ -0,0 +1,49 @@
+import dayjs from 'dayjs'
+
+const locale = {
+  name: 'en',
```
iamkun: en-gb

iamkun:
What's the difference between en-gb and built in en locale, please?

tcrands:
The difference is the time and data formats. It says the built in en locale is US so this won't work for the UK.

Alternatively we could make the default en locale UK based?

iamkun:
Is `formats` the only different part?

tcrands:
Yes I believe formats is the only difference

src/locale/en-gb.js
```diff
@@ -0,0 +1,50 @@
...
+    const b = number % 10
+    let output = 'th'
+    if (b === 1) {
+      output = 'st'
+    } else if (b === 2) {
+      output = 'nd'
+    } else if (b === 3) {
+      output = 'rd'
+    }
+    return number + output
```
iamkun: ordinal = (number) => {
    const s = ['th', 'st', 'nd', 'rd']
    const v = number % 100
    return `[${number}${(s[(v - 20) % 10] || s[v] || s[0])}]`
  }

This looks better

iamkun:
npm run lint first, pls

PR #485, opened 2019-02-10
Create my.js

pete-arif:
suggestion for Bahasa Melayu (Malaysia) locale


src/locale/my.js
```diff
@@ -0,0 +1,36 @@
+import dayjs from 'dayjs'
+
+const locale = {
+  name: 'id',
```
iamkun: please keep name the same with file name

iamkun:
Should this be `ms` instead of `my`? https://en.wikipedia.org/wiki/List_of_ISO_639-2_codes

iamkun:
and file name pls

pete-arif:
> and file name pls
> Should this be `ms` instead of `my`? https://en.wikipedia.org/wiki/List_of_ISO_639-2_codes

It should be... apologies. 👍  
_(personally, not really agree to that ISO naming for the language, but standards is standards)_

Issue #486, opened 2019-02-11
Refactor (and redesign?) parseDate()


suspectpart:
@iamkun Here is the issue with my comments, following the discussion in #417.

https://github.com/iamkun/dayjs/blob/a7e05e0b7fa529e8d81ebec3f007eb817d970fdd/src/index.js#L53-L56
Here I want to get rid of the warning `eslint-disable-next-line no-cond-assign`, I think that is not good style. Also, the check whether there is a `z` in the end of the string or not does not need to be a regex, which is also hinted by the comment `looking for a better way`. I propose these changes:

```javascript
if (typeof date === 'string' && !date.toLowerCase().endsWith('z')) {
    const d = date.match(C.REGEX_PARSE)
    if (d) {
      return new Date(d[1], d[2] - 1, d[3] || 1, d[4] || 0, d[5] || 0, d[6] || 0, d[7] || 0)
    }
}
```
this also gets rid of the `let reg` in the first line of `parseDate`, moving the variable down where it is needed, renaming it to just `d` as `reg` was also quite inprecise as this is not a regex but rather matched parts of a date.

https://github.com/iamkun/dayjs/blob/a7e05e0b7fa529e8d81ebec3f007eb817d970fdd/src/index.js#L63

Here the comment is misleading. This is **not** the default return for timestamps only, but all of the following inputs will end up in this last return statement:

- `parseDate({}) // objects` (invalid)
- `parseDate(function() {}) // function` // invalid
- `parseDate(Infinity) // non-finite number` (invalid)
- `parseDate(12345) // finite number` (valid, interpreted as unix timestamp)
- `parseDate('Mon Jan 01 2018 00:00:00 GMT+0100 (Mitteleuropäische Normalzeit)') // RFC 2822 date string` (valid)
- `parseDate(() => '2018-01-01') // arrow func` (valid in Chrome, invalid in Firefox)
- `parseDate([2018, 1, 1]) // Array` (valid in both Firefox and Chrome!)

So I got two issues with this:
- The comment is misleading and most of the cases I mention above are not explicitly tested by the dayjs test suite
- We do not abstract the Date constructor default behavior, even causing incosistencies between different browsers

In my opinion, the second point is what I expect dayjs to do - provide a clear and straightforward API to create dates without any weird special cases. Instead almost everything is forwarded to the Date constructor, and adding more complexity even:
[...]

PR #487, opened 2019-02-11
Refactor parseDate

suspectpart:
See #486

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -45,21 +45,17 @@ Utils.isDayjs = isDayjs
 Utils.wrapper = wrapper
 
 const parseDate = (date) => {
-  let reg
-  if (date === null) return new Date(NaN) // Treat null as an invalid date
-  if (Utils.isUndefined(date)) return new Date()
+  if (date === null) return new Date(NaN) // null is invalid
+  if (Utils.isUndefined(date)) return new Date() // today
   if (date instanceof Date) return date
-  // eslint-disable-next-line no-cond-assign
-  if ((typeof date === 'string')
-    && (/.*[^Z]$/i.test(date)) // looking for a better way
-    && (reg = date.match(C.REGEX_PARSE))) {
-    // 2018-08-08 or 20180808
-    return new Date(
-      reg[1], reg[2] - 1, reg[3] || 1,
-      reg[4] || 0, reg[5] || 0, reg[6] || 0, reg[7] || 0
-    )
+  if (typeof date === 'string' && !date.toLowerCase().endsWith('z')) {
+    const d = date.match(C.REGEX_PARSE)
+    if (d) {
+      return new Date(d[1], d[2] - 1, d[3] || 1, d[4] || 0, d[5] || 0, d[6] || 0, d[7] || 0)
+    }
   }
-  return new Date(date) // timestamp
+
+  return new Date(date) // everything else
 }
 
 class Dayjs {
--- a/src/index.js
+++ b/src/index.js
@@ -48,7 +48,7 @@ const parseDate = (date) => {
   if (date === null) return new Date(NaN) // null is invalid
   if (Utils.isUndefined(date)) return new Date() // today
   if (date instanceof Date) return date
-  if (typeof date === 'string' && !date.toLowerCase().endsWith('z')) {
+  if (typeof date === 'string' && !/Z$/i.test(date)) {
     const d = date.match(C.REGEX_PARSE)
     if (d) {
       return new Date(d[1], d[2] - 1, d[3] || 1, d[4] || 0, d[5] || 0, d[6] || 0, d[7] || 0)
--- a/test/parse.test.js
+++ b/test/parse.test.js
@@ -15,7 +15,7 @@ describe('Parse', () => {
     expect(dayjs().valueOf()).toBe(moment().valueOf())
   })
 
-  it('String 20130208', () => {
+  it('moment-js like formatted dates', () => {
     global.console.warn = jest.genMockFunction()// moment.js '2018-4-1 1:1:1:22' will throw warn
     let d = '20130108'
     expect(dayjs(d).valueOf()).toBe(moment(d).valueOf())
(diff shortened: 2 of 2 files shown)


src/index.js
```diff
@@ -45,21 +45,17 @@ Utils.isDayjs = isDayjs
...
-  // eslint-disable-next-line no-cond-assign
-  if ((typeof date === 'string')
-    && (/.*[^Z]$/i.test(date)) // looking for a better way
-    && (reg = date.match(C.REGEX_PARSE))) {
-    // 2018-08-08 or 20180808
-    return new Date(
-      reg[1], reg[2] - 1, reg[3] || 1,
-      reg[4] || 0, reg[5] || 0, reg[6] || 0, reg[7] || 0
-    )
+  if (typeof date === 'string' && !date.toLowerCase().endsWith('z')) {
```
iamkun: Just noticed that `endsWith` is a new added method in js, with cost some more bytes than `/.*[^Z]$/i.test` in the final es5 code.

test/parse.test.js
```diff
@@ -46,6 +46,40 @@ describe('Parse', () => {
     expect(dayjs(time).valueOf()).toBe(moment(time).valueOf())
   })
 
+  it('String RFC 2822, time and zone', () => {
```
iamkun: The added test suit here seems just test Date object rather dayjs here?

suspectpart: I just wanted to make that functionality explicit, because right now it is neither documented nor explicitly handled in the tests. Whenever we change our internal implementation we now get feedback whether our assumptions are still valid, right? Will tell you about breaking changes then.

src/index.js
suspectpart: Yes, but the concept is clearer. I also saved some letters on renaming `reg[...]` to `d[...]` multiple times and removing the `let reg`, wouldn't that even it out?

iamkun: `reg` will be minified in building I think, and `endsWith` becomes the added bytes.

suspectpart: Funnily, this was never needed at all, there is a bug in the regex.

If you look at it

export const REGEX_PARSE = /^(\d{4})-?(\d{1,2})-?(\d{0,2})[^0-9]*(\d{1,2})?:?(\d{1,2})?:?(\d{1,2})?.?(\d{1,3})?$/i`

you wonder: Why does it even match a `z` or `Z` it in the end? It is because the last dot in the regex is not escaped (the dot in part `?.?(\d{1,3})?`), so that matches *any character* and therefore matches the `z`

If you change it to

export const REGEX_PARSE = /^(\d{4})-?(\d{1,2})-?(\d{0,2})[^0-9]*(\d{1,2})?:?(\d{1,2})?:?(\d{1,2})?\.?(\d{1,3})?$/i`

(note the backlash to escape the dot), all tests are still green if we just do

```javascript
  if (typeof date === 'string') {
    const d = date.match(C.REGEX_PARSE)
    if (d) {
      return new Date(d[1], d[2] - 1, d[3] || 1, d[4] || 0, d[5] || 0, d[6] || 0, d[7] || 0)
    }
  }
```

So we even save a lot of bytes :-).

suspectpart: Pushed that fix to PR.

iamkun: Seems we've changed this regex a few days ago, maybe that's the reason?

suspectpart: But there are other issues with that check (no matter if we use the new or old regex): For example this date is not parsed by dayjs:

`2018-05-02T11Z` (missing minutes and seconds), but it is parsed by moment.js. Maybe we need a lot more test cases here...

suspectpart: Check my latest push - it will fix all of the issues I just mentioned automatically. All we want to do in our custom check is catch the yyyymmdd shorthand so you can parse `20180501`, the rest can be delegated to the `Date()` constructor just like moment.js does. All tests still green.

iamkun: Looks like magic, but could this just because we are lacking of test suits?

suspectpart: Yes, but some additional tests I wrote would fail in both cases. Pure Javascript allows for strings without a time separator `T` (so just having a space) or single-digit time or date bits (e.g. `2018-4-1`), there is no need to parse that explicitly - `new Date()` will handle it for us.

But still there is an incosistency with moment.js, as they allow partial time strings, e.g. `2018-01-01 13Z`, which is not ok with dayjs. I could add some (failing) tests to make it clear?

xxyuk: No you can't simplify the regex like this. The reason why we are using such long regex is the Date object varies among different browsers, expecially mobile browsers. Then you will facing a lot compatibility problem. 

@iamkun is correct, the magic is just because we are lacking some multi browser tests in a PR checking ( but will do in every CI release)

suspectpart: @xxyuk Ok, thanks for the info, I didn't know that this also covers cross-browser concerns. But maybe it would be good to capture the different browser behaviors in unit tests so the developers get a more instant feedback on doing changes?

suspectpart: And apart from that: The unescaped dot is still an issue, right?

xxyuk: Cross-browser tests cost too much time and resources, that why we only do this on every release.

> The unescaped dot is still an issue, right?

I'm not sure if it's a issue, and my suggestion is you could first revert to the original `regex(Z)`, and let this PR get merged. And open a new PR with this change, we are happy to make a manual multi browser test to see if it's ok

suspectpart: I will do that, @xxyuk , good ṕlan.

suspectpart: Ok, so I force pushed back to just the initial refactoring. Can we merge that as is? Next up I create a PR where I cut down the regex and escape the dot properly and we run through the CI?

iamkun: Although we are looking for a better way, I think `/.*[^Z]$/i.test(date)` is better here than `endsWith`

The rest looks fine.

suspectpart: Ok I did, and even shortened it as `/z$/i` does the same (asserting it ends with a `z`, case insensitive). `.*[^Z]$` was doing too much, no need to match any character 0-n times.

iamkun: `/z$/i.test('z')` doesn't make sense.

suspectpart: Yes, I fixed it. I could just use the web view as I do not have the solution checked out right now, will crush those commits, sorry.

suspectpart: Needs to be `/[^Z]$`.test(date)`, getting rid of `.*`.

test/parse.test.js
```diff
@@ -46,6 +46,40 @@ describe('Parse', () => {
...
+    expect(dayjs(Infinity).isValid()).toBe(false)
+    expect(dayjs(NaN).isValid()).toBe(false)
+    expect(dayjs([2018, 5, 1, 13, 52, 44]).isValid()).toBe(false) // Arrays with time part
+  })
+
+  it('parses Arrays with date part', () => {
+    const dateParts = [2018, 5, 1]
+    const expected = '2018-05-01T00:00:00.000Z'
+    const d = dayjs(dateParts)
+    const normalized = d.add(d.utcOffset(), 'minutes') // make test run in every timezone
```
iamkun: what's this test for?

src/index.js
suspectpart: Ok, I crushed those commits and no put the regex back. I tried to keep the Regex as short as possible, so I no went for

`!/Z$/.test(date)` rather then `/[^Z]$`i.test(date)`, as it is
- less bytes (not many)
- easier to comprehend the negation outside of the regex instead of with the [^Z] negation syntax. Is that ok now?

test/parse.test.js
suspectpart: Making implicit behavior explicit: Arrays fall through to the default `Date` constructor and should work according to ECMAScript standard (but only for the date part, time is not supported), so I wanted to assert that here explicitly. Testing this is a first step to also documenting that behavior or to change the behavior and be more strict than standard Javascript is. `moment.js` is way more opinionated about these cases than `dayjs` is, but it is hard to spot the differences.

suspectpart: Moment.js will handle that case differently than dayjs will:

```
const arr = [2018, 5, 1];
const dayjsDate = moment([2018, 5, 1]);
const momentDate = dayjs([2018, 5, 1]);
console.log(`dayjs: ${dayjsDate.toISOString()}`)
console.log(`momentjs: ${momentDate.toISOString()}`)
```

yields

```
dayjs: 2018-05-31T22:00:00.000Z
momentjs: 2018-04-30T22:00:00.000Z
```

So I wanted to assert that explicitly here.

iamkun: actually I'm asking why ` make test run in every timezone` 😬

suspectpart: If you parse a date without tz information (which can not be avoided with the Array), it will parse it as if it were your local timezone. So depending on in which timezone you run the test, you would get different results. To avoid that, I normalize the result by adding the UTC offset so I can expect against a fixed GMT timezone string.

iamkun: Got you, but a little bit hard to understand.

Can we just compare `format()` string or unix timestamp instead?

suspectpart: But it's a whole different date, isn't it? I am adding seconds to make it run, after all... not just reformatting it? 

If I take this example and run it locally, it actually assumes GMT+2 (because 2018-05-01 in Germany is summertime):

```
new Date([2018, 5, 1]).toUTCString()
"Mon, 30 Apr 2018 22:00:00 GMT"
```
So depending on where you run this code, it will assume the GMT time to be relative to the timezone that will be valid in that place at that point in time. Just correcting the timezone via its offset seems to be the easiest way? But maybe I am missing something?

iamkun: oh I see. You are right.

suspectpart: I struggled with this a lot during work on a project that made heavy use of dates (my task was to switch from momentjs to dayjs writing a million tests for everything). Those tests were all fine on my machine but failed on machines of colleagues in different timezones, so I wrote an abstraction that kind of does what my test here is doing (adjusting the timezone by offset). It is really not easy getting all of that right :/.

iamkun: @neemzy seems the test we made
```
expect dayjs() tobe moment()
```
just skipped this issue 😬

suspectpart:
Are we good to merge this and focus on the other PR (shortening the regex)? The other one is ready to be checked with the CI build and cross-browser testing.

PR #488, opened 2019-02-11
Update Swedish and Finnish locales

frozenzia:
Swedish locale was missing weekdaysShort and Min, and monthsShort.
Finnish locale's monthsShort was same as months. It is now updated to be similar to moment.


iamkun:
THX

Issue #497, opened 2019-02-15
Unix timestamp is incorrect?


jsumners:
`if (new Date().getTimezoneOffset() === 0)` then `dayjs.unix(1550251173).startOf('day').unix()` returns `1550188800`.

`if (new Date().getTimezoneOffset() === 300)` then `dayjs.unix(1550251173).startOf('day').unix()` returns `1550206800`.

A Unix timestamp is by definition a count of seconds from "00:00:00 Thursday, 1 January 1970, Coordinated Universal Time (UTC)" (http://pubs.opengroup.org/onlinepubs/9699919799/basedefs/V1_chap04.html#tag_04_16). Thus, the correct result should _always_ be `1550188800`.

xxyuk:
Because in `dayjs.unix(1550251173)`, dayjs parses and displays in local time.

jsumners:
Yes, that should parse into local time. But the final conversion _back_ to a Unix timestamp should generate a number at UTC.

xxyuk:
Yes, and that will be of course different. Cause `startOf('day')` is different.

jsumners:
What do you not understand? The number that results from `dayjsInstance.unix()` should be the number of seconds from 00:00:00 Thursday, 1 January 1970 _at the UTC timezone_.

jsumners:
I'm closing this until I can figure out where the actual bug lies. I'm not really sure what is happening here.

iamkun:
try UTC plugin

PR #499, opened 2019-02-16
Resolve issue #497

jsumners:
Here's a simple fix for #497. Note, there seems to be a failing test in the `dayOfYear` suite. As far as i can tell, the test code should not be hitting the path modified by this PR. But I don't know how to run your test suite through the debugger given all of the transpilation. So I can't be sure.

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -141,7 +141,7 @@ class Dayjs {
   }
 
   unix() {
-    return Math.floor(this.valueOf() / 1000)
+    return Math.floor(new Date(this.$d.toUTCString()) / 1000)
   }
 
   valueOf() {


src/index.js
```diff
@@ -141,7 +141,7 @@ class Dayjs {
   }
 
   unix() {
-    return Math.floor(this.valueOf() / 1000)
+    return Math.floor(new Date(this.$d.toUTCString()) / 1000)
```
xxyuk: I don't this is a proper fix. new date will always use local time zone

jsumners: Yes. I just realized this myself.

jsumners:
Nope. Try again.

PR #502, opened 2019-02-19
Fix docs about escaping with curly brackets

denis-sokolov:
The current behavior is treat curly brackets as regular characters, only the square brackets have special behavior.

Related code:

https://github.com/iamkun/dayjs/blob/8ede9b76974a1aa7a14f144c58c5e4ce63c393d6/src/index.js#L304-L307

https://github.com/iamkun/dayjs/blob/8ede9b76974a1aa7a14f144c58c5e4ce63c393d6/src/constant.js#L30


iamkun:
Yes, thanks.

PR #507, opened 2019-02-22
Add argument types for dayjs() methods

blankPen:
- support usage `dayjs('01-01-2019','DD-MM-YYYY')`  in typescript


iamkun:
Thanks, but this is not a good fix. I have updated the `indes.d.ts`.

PR #508, opened 2019-02-23
swahili version

0biWycliffe:


iamkun:
@0biWycliffe Thanks.
Why this commit has quite a lot of different compared with https://github.com/moment/moment/blob/develop/locale/sw.js?

PR #509, opened 2019-02-23
swahili version

Tracyterry:


iamkun:
Thanks. What' the relation with PR #508? 

Plus, according to https://en.wikipedia.org/wiki/List_of_ISO_639-2_codes, Swahili shoud be sw ?

PR #514, opened 2019-02-27
minor update, correcting semantics

pete-arif:


iamkun:
@pete-arif

Ref: https://github.com/moment/moment/blob/develop/locale/ms.js

iamkun:
@pete-arif Any comments here?

pete-arif:
it's to correcting semantics

iamkun:
@pete-arif
But according to 
https://github.com/moment/moment/blob/develop/locale/ms.js

This should be the right version?

pete-arif:
https://github.com/moment/moment/blob/develop/locale/ms.js is correct indeed. The one I am suggesting is more formally used. Perhaps need todo pull request on that one instead?

iamkun:
@pete-arif 
As you could imagine, this is a hard desition to me cause I'm not a native speaker of your language.
That's why I use moment's locale as a reference. 

Maybe we need some more reviewers from your language to make updates of both locales altogether.

PR #515, opened 2019-03-04
Added de-AT locale

mariomc:
This PR adds in the de-AT locale, based in the de locale, with differences in the months name and the seconds.


iamkun:
THANKS

PR #525, opened 2019-03-10
Bikram sambat

adhocore:
- adds weekdays short version for `ne` locale
- adds a `bikramSambat` (BS) plugin with method `toBS(formatStr: string): string`
   that converts date (year/month/date) from default `AD` to `BS` and also formats as per `formatStr`

BS calendar &/or date [system](https://en.wikipedia.org/wiki/Vikram_Samvat) is still widely used thus necessitating a plugin like this. 

the tests can be cross verified here: http://nepalicalendar.rat32.com/#date-conversion


iamkun:
First of all, thanks.

> adds weekdays short version for ne locale

Looks cool.

>adds a bikramSambat (BS) plugin with method toBS(formatStr: string): string
that converts date (year/month/date) from default AD to BS and also formats as per formatStr

Seems better release as a separate plugin by your self like https://github.com/alibaba-aero/jalaliday, cause it's hard for us to manage a unfamiliar calendar date format.

How do you think?

adhocore:
also good. i will do a new PR (in a day or two) with locale only and a ref to new plugin package

Issue #533, opened 2019-03-14
customParseFormat can't parse.


sxd1140:
**Describe the bug**
Got `Invalid Date` when use `customParseFormat` parse.
https://codepen.io/anon/pen/RdQrMe

**Expected behavior**
Get right date.

**Information**
 - Day.js Version: lastest `https://unpkg.com/dayjs`
 - OS: macOS
 - Browser: chrome 72

PR #536, opened 2019-03-17
fixed formatting issue

waseemahmad31:
Fixed issue #533 

Got Invalid Date when use customParseFormat parse.

### DIFF
--- a/src/plugin/customParseFormat/index.js
+++ b/src/plugin/customParseFormat/index.js
@@ -9,7 +9,7 @@ const matchUpperCaseAMPM = /[AP]M/
 const matchLowerCaseAMPM = /[ap]m/
 const matchSigned = /[+-]?\d+/ // -inf - inf
 const matchOffset = /[+-]\d\d:?\d\d/ // +00:00 -00:00 +0000 or -0000
-const matchWord = /\d*[^\s\d]+/ // Word
+const matchWord = /\d*[^\s\d-:/.()]+/ // Word
 
 let locale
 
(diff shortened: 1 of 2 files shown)


src/plugin/customParseFormat/index.js
```diff
@@ -9,7 +9,7 @@ const matchUpperCaseAMPM = /[AP]M/
 const matchLowerCaseAMPM = /[ap]m/
 const matchSigned = /[+-]?\d+/ // -inf - inf
 const matchOffset = /[+-]\d\d:?\d\d/ // +00:00 -00:00 +0000 or -0000
-const matchWord = /\d*[^\s\d]+/ // Word
+const matchWord = /\d*[^\s\d-:/.()]+/ // Word
```
iamkun: Good fix.

Do we need to list more possible divider (such as \ / _ etc, I'm not sure)?

waseemahmad31: Yes, we can include if we are considering them as the format separator, in this case, we also need to modify https://github.com/iamkun/dayjs/blob/9b3ba887fe1428e4a963daa9b691e2f2db36cb1d/src/plugin/customParseFormat/index.js#L1.

Should I include these dividers("_" and "\")?

iamkun: Not sure, to be honest.

Maybe we could add these later but not this time. 😬

waseemahmad31: Thanks @iamkun. :)

PR #541, opened 2019-03-22
Add 79 missing locales

xxyuk:
Hi,

I wrote a script and copied all the missing locales from moment.js.

Just the most important parts like month name and week name.


mariomc:
Thanks @xxyuk !
Could you extend (or make available) the script for the localized formats as well? [locale.format in dayjs](https://github.com/iamkun/dayjs/blob/dev/src/locale/en-gb.js#L26) and [locale.longDateFormat in moment](https://github.com/moment/moment/blob/develop/src/locale/en-gb.js#L13)
Pretty please 😅

src/locale/en-au.js
```diff
@@ -0,0 +1,17 @@
+import dayjs from 'dayjs'
+
+const locale = {
+  name: 'en-au',
+  weekdays: 'Sunday_Monday_Tuesday_Wednesday_Thursday_Friday_Saturday'.split('_'),
+  months: 'January_February_March_April_May_June_July_August_September_October_November_December'.split('_'),
+  weekStart: 1,
```
sstur, line 7: I do not believe this is correct (or else Moment has changed it since you made this PR). [According to Moment](https://github.com/moment/moment/blob/master/locale/en-au.js) Australia has a weekStart of 0 (Sunday). I can't find a canonical source to back up either choice, I wanted to point out that Day.js is deviating from Moment in this case. When tested on MacOS (System Preferences -> Languages & Region -> change Region to Australia) it appears that Australia starts their week on a Sunday.

PR #546, opened 2019-03-28
feat: add support for lowercase localizable formats (l, ll, lll, llll)

mariomc:
This PR tries to bridge the gap between dayjs and moment by implementing lowecase/short localizable formats, already present in moment, since 2.0.0

### DIFF
--- a/src/constant.js
+++ b/src/constant.js
@@ -27,4 +27,4 @@ export const INVALID_DATE_STRING = 'Invalid Date'
 
 // regex
 export const REGEX_PARSE = /^(\d{4})-?(\d{1,2})-?(\d{0,2})[^0-9]*(\d{1,2})?:?(\d{1,2})?:?(\d{1,2})?.?(\d{1,3})?$/
-export const REGEX_FORMAT = /\[.*?\]|Y{2,4}|M{1,4}|D{1,2}|d{1,4}|H{1,2}|h{1,2}|a|A|m{1,2}|s{1,2}|Z{1,2}|SSS/g
+export const REGEX_FORMAT = /(\[[^\]]+])|Y{2,4}|M{1,4}|D{1,2}|d{1,4}|H{1,2}|h{1,2}|a|A|m{1,2}|s{1,2}|Z{1,2}|SSS/g
--- a/src/constant.js
+++ b/src/constant.js
@@ -27,4 +27,4 @@ export const INVALID_DATE_STRING = 'Invalid Date'
 
 // regex
 export const REGEX_PARSE = /^(\d{4})-?(\d{1,2})-?(\d{0,2})[^0-9]*(\d{1,2})?:?(\d{1,2})?:?(\d{1,2})?.?(\d{1,3})?$/
-export const REGEX_FORMAT = /(?:\[([^\]]+)])|Y{2,4}|M{1,4}|D{1,2}|d{1,4}|H{1,2}|h{1,2}|a|A|m{1,2}|s{1,2}|Z{1,2}|SSS/g
+export const REGEX_FORMAT = /\[([^\]]+)]|Y{2,4}|M{1,4}|D{1,2}|d{1,4}|H{1,2}|h{1,2}|a|A|m{1,2}|s{1,2}|Z{1,2}|SSS/g
--- a/src/constant.js
+++ b/src/constant.js
@@ -27,4 +27,4 @@ export const INVALID_DATE_STRING = 'Invalid Date'
 
 // regex
 export const REGEX_PARSE = /^(\d{4})-?(\d{1,2})-?(\d{0,2})[^0-9]*(\d{1,2})?:?(\d{1,2})?:?(\d{1,2})?.?(\d{1,3})?$/
-export const REGEX_FORMAT = /(\[[^\]]+])|Y{2,4}|M{1,4}|D{1,2}|d{1,4}|H{1,2}|h{1,2}|a|A|m{1,2}|s{1,2}|Z{1,2}|SSS/g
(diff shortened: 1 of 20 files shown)


iamkun:
Thanks, will consider this PR.

mariomc:
Thanks!
If you wish to keep the localizableFormat plugin slim, I can write another plugin just for this.

iamkun:
It's not about the size of this plugin, but the size of all other locale files.

If we introduce some more localizable formats, all the locale files will be updated and added these format. That's a huge increase in the total size of our locale files.

That's why I'd like to keep this PR pending until more feature requests.

mariomc:
I see your worry. However, in the case of moment, there's only a few like [zh-cn](https://github.com/moment/moment/blob/develop/src/locale/zh-cn.js) and [sv](https://github.com/moment/moment/blob/develop/locale/sv.js) who set (some) of their own short localized formats.
The big majority of locales don't even define short localized formats and fallback to the default behavior/configuration.

dayjs is well designed and this funcionality can be implemented easily with a plugin (and, for my case, I did just that). I just thought that, by having the current locale plugin with this capacity (hence the PR), it could ease the transition, for those, like me, who are migrating from moment to dayjs.

In any case, thank you for the reply and awesome work on dayjs!

iamkun:
You are right, only a few have this. Is there a full list of locales that have lowercase localizable formats instead of checking every locale files on GitHub one by one? 

I'm willing to merge this if it's not that much. 😬

mariomc:
On moment/locale, doing a `grep -e ' l :' -e ' ll : ' -e ' lll :' -e ' llll :' *.js | awk '{print $1}' | sed "s/.js://" | uniq -ci` outputs the locale list that use a short-hand localised formats, with the number of different short-hand locales they use (l, ll, lll, llll):

```
   3 ca
   1 cs
   4 eu
   4 fi
   4 he
   4 ja
   4 ko
   4 lt
   2 sv
   4 vi
   4 zh-cn
   4 zh-hk
   4 zh-tw
```

It's a bit more than 10% (13/127), but still not too many. That should represent 46 new lines total.

iamkun:
Great!

xxyuk:
We should update the docs accordinglly

mariomc:
Will do. Adding current locales as well as fixing current tests to contain those key additions. Noticed an issue with localizedFormats. Will keep in touch and open accordingly.

src/plugin/localizedFormat/index.js
```diff
@@ -9,15 +9,19 @@ export default (o, c, d) => {
...
+    llll: 'ddd, MMM D, YYYY h:mm A'
   }
   d.en.formats = englishFormats
   proto.format = function (formatStr) {
     const locale = this.$locale()
     const formats = locale.formats || {}
     const str = formatStr || FORMAT_DEFAULT
-    const result = str.replace(/LTS|LT|L{1,4}/g, match =>
-      formats[match] || englishFormats[match])
+    const result = str.replace(/(\[[^\]]+])|(LTS?|l{1,4}|L{1,4})/g, (_, a, b) =>
```
mariomc: Props to @Anmo on the regex magic

mariomc:
Added fix. Will review and add the missing localized formats for fi, he, lt.
Didn't add them previously because the interpolation issue.

xxyuk:
`/(\[[^\]]+])|` looks nice compare to the original version in core `index.js` format method.

Maybe we could update the `index.js` file to make it size smaller?

mariomc:
Well thought @xxyuk, thanks! We managed to bring the package size down a notch 😄 
Should we consider adding final bundle size metrics alongside codecov?

xxyuk:
Not sure, can codecov do this?

mariomc:
I don't know about codecov. The only one I do know is [bundlesize](https://github.com/siddharthkp/bundlesize) and it looks great!

xxyuk:
Oh I see, we already have such size limit tool during our CI build. You could find them in `package.json`.

xxyuk:
Maybe update doc here as well ? 
https://github.com/iamkun/dayjs/blob/dev/docs/en/Plugin.md#localizedformat

mariomc:
Done. I also took a dabble at the regex previously mentioned and managed to shave off a few more bytes off of it.
Thanks for the heads-up @xxyuk !

src/constant.js
```diff
@@ -27,4 +27,4 @@ export const INVALID_DATE_STRING = 'Invalid Date'
 
 // regex
 export const REGEX_PARSE = /^(\d{4})-?(\d{1,2})-?(\d{0,2})[^0-9]*(\d{1,2})?:?(\d{1,2})?:?(\d{1,2})?.?(\d{1,3})?$/
-export const REGEX_FORMAT = /\[.*?\]|Y{2,4}|M{1,4}|D{1,2}|d{1,4}|H{1,2}|h{1,2}|a|A|m{1,2}|s{1,2}|Z{1,2}|SSS/g
+export const REGEX_FORMAT = /(?:\[([^\]]+)])|Y{2,4}|M{1,4}|D{1,2}|d{1,4}|H{1,2}|h{1,2}|a|A|m{1,2}|s{1,2}|Z{1,2}|SSS/g
```
iamkun: Can we simplify here as `\[([^\]]+)]` instead of `(?:\[([^\]]+)])` ?

test/parse.test.js
```diff
@@ -18,6 +18,7 @@ describe('Parse', () => {
   it('moment-js like formatted dates', () => {
     global.console.warn = jest.genMockFunction()// moment.js '2018-4-1 1:1:1:22' will throw warn
     let d = '20130108'
+    let f = '[interpolated] mm dd'
```
iamkun: I think we have such tests in `display.test.js` L#140

mariomc: You are totally right. Will fix accordingly. Still finding my way around the tests 😅

PR #548, opened 2019-03-28
fix: add support for year before 100

dmitriytat:
This PR brings ability to use years between 0 and 99 while parse date string

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -50,14 +50,22 @@ const parseDate = (cfg) => {
   if (date === null) return new Date(NaN) // null is invalid
   if (Utils.u(date)) return new Date() // today
   if (date instanceof Date) return new Date(date)
+
   if (typeof date === 'string' && !/Z$/i.test(date)) {
     const d = date.match(C.REGEX_PARSE)
     if (d) {
+      let dt
+
       if (utc) {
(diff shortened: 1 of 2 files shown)


iamkun:
Thanks to this fix. Just not sure if there a real use case or not for the year before 100?

dmitriytat:
I In my case it was validation of birthdate in text input. Date restrictions were from 1900 to now - 13 years. But if user selects 0010-04-05, it will be converted to 1910-04-05 and input pass validation.

xxyuk:
`0010-04-05` is a strange date string either. Can we just update the date picker to get the correct date string?

dmitriytat:
I agree, it’s weird, but it’s format of default html input type=date. And yes, we can use min and max with this type, if it is supported.

I think it’s strange that I can't use dayjs just like I used to use moment or vanilla new Date(‘0010-04-04’).

iamkun:
Yes, it's a small bug with Day.js of parsing `0010-04-04` format. But such date string is too wired that we could just ignore this. 

Until we could find some other need for parsing this kind of string, I'm afraid I may not want to merge this.

iamkun:
I'll close this pr since it's been a while since it's been opened. Feel free to reopen if you have updates on this

PR #551, opened 2019-03-31
Adds Occitan locale file

Quenty31:
In its general / central dialect


src/locale/oc.js
```diff
@@ -0,0 +1,38 @@
+import dayjs from 'dayjs'
+
+const locale = {
+  name: 'ca',
```
mariomc:
```suggestion
  name: 'oc',
```

iamkun:
https://github.com/moment/moment/blob/develop/src/locale/oc-lnc.js

what's the difference between `oc` and `oc-lnc`, please?

src/locale/oc.js
Quenty31: Done! Thanks!

Quenty31:
There is no difference, if I put only «oc» it means «lengadocian» occitan. It's almost the standard, but we don’t have a standard form, the lengadocian is the most central dialect.

iamkun:
@Quenty31 should we keep the same with moment.js and change to oc-lnc as well?

Quenty31:
It's up to up, both are correct, but if you want to use the same as moment.js, feel free to rename it :)

iamkun:
Please keep the same with moment.js then. 😁

iamkun:
and update `name: 'oc',`, please.

PR #557, opened 2019-04-02
feat: make lowercase localized formats fallback to slightly modified …

mariomc:
…uppercase formats

This feature, present in moment, is the reason why most locales don't need to define a lowercase locale.


iamkun:
oops, I renamed `localizableFormat.test.js` and caused a conflict. How could we resolve this?

mariomc:
No worries. Let me take a look later today and I'll resolve the conflict.

xxyuk:
check the current code, just rename to localizedFormat.test.js will be fine

PR #563, opened 2019-04-08
Typo fix in API-reference.md

ledleds:
Just noticed this tiny typo in your docs!


xxyuk:
@ledleds I think we should do a global search / replace of other language such as `es-es` and `pt-br`

ledleds:
> @ledleds I think we should do a global search / replace of other language such as `es-es` and `pt-br`

Good shout, will do it now

iamkun:
Thanks

PR #566, opened 2019-04-14
Update Japanese API Reference

oti:
I intended to make up for the lack of translation, but the outline of the document was different from the original (English version), so I followed it and retranslated.


iamkun:
That's awesome! Thanks.

Issue #555, opened 2019-04-02
YYYY-MM-DD Format Not Working


GoPro16:
**Describe the bug**
When passing in a parse format of `YYYY-MM-DD` with an invalid date string of `10/12/2014`, should result in and invalid date but instead resolves to `08-14-2015`

**Expected behavior**
Should result in `date.isValid() === false`

**Information**
 - Day.js Version [e.g. v1.8.11]
 - OS: [e.g. MacOS]
 - Browser [e.g. chrome 73]

[**CodeSandbox Link**](https://codesandbox.io/s/km684x1qpo)

xxyuk:
With `customParseFormat` plugin
`dayjs('2014-12-01', 'YYYY-MM-DD').toISOString()` returns 2014-11-30T16:00:00.000Z

Looks good.

GoPro16:
@xxyuk Right sorry, I updated my issue comments. If you check the codesandbox it should be throwing an error when trying to parse the invalid date format but instead parses to a random date.

iamkun:
will fix soon

IssuehuntBot:
@issuehunt has funded $50.00 to this issue.

---
- Submit pull request via [IssueHunt](https://issuehunt.io/repos/128907699/issues/555) to receive this reward.
- Want to contribute? Chip in to this issue via [IssueHunt](https://issuehunt.io/repos/128907699/issues/555).
- Checkout the [IssueHunt Issue Explorer](https://issuehunt.io/issues) to see more funded issues.
- Need help from developers? [Add your repository](https://issuehunt.io/r/new) on IssueHunt to raise funds.

adil-chbada:
https://day.js.org/docs/en/plugin/custom-parse-format#docsNav

PR #568, opened 2019-04-14
fixed parsing issue

waseemahmad31:
fix issue #555


iamkun:
Thanks. This is a clever fix. I'll merge it first.

Still, a refactor is working in progress in order to avoid such unexpected behavior and make the logic more clear and throw an error on (month greater than 13, date greater than 31, ... )

Issue #571, opened 2019-04-16
CustomParseFormat escape unknown lenght string


vendelin8:
Hi,

I kind of like this project, I'd like to use if for parsing different date formats.

There are some, that contain to be ignored strings, that are unknown length, eg.: `Monday, 22 April, 18:00`. In this case I could cut it at the first space, but generally it would be awesome to have something simpler for this. Is it possible to add something like that? Thanks!

xxyuk:
A reproduction online demo or code, please?

vendelin8:
```
import customParseFormat from 'dayjs/plugin/customParseFormat';
dayjs.extend(customParseFormat);

var fmt = '[someday] DD MMMM HH:mm';

var d = dayjs('Tuesday 16 January 18:00', fmt);
// works fine: d.isValid() is true

d = dayjs('Monday 15 January 18:00', fmt);
// fails: d.isValid() is false, because [someday] uses the space at the end as well

d = dayjs('Saturday 20 January 18:00', fmt);
// fails too: d.isValid() is false, because [someday] lets the y hanging at the end
```

xxyuk:
I don't think @waseemahmad31 's #576 is a proper fix for this issue.

The main problem is that Day.js can't deal with unknown length of strings like 'spaces'.

waseemahmad31:
The solution I have tried to implement is providing a way to parse the weekday names using customParseFormat plugin. Also by looking at the problem that @vendelin8 faced, it seems like he was trying to find a way to parse the weekday names. @xxyuk  I too agree its not the solution by which we can parse variable length of string for a token like `[someday]`

PR #576, opened 2019-04-20
customParseFormat: added logic to parse the week day

waseemahmad31:
Resolves issue #571 

import customParseFormat from 'dayjs/plugin/customParseFormat';

var fmt = 'dddd DD MMMM HH:mm';

var d = dayjs('Tuesday 16 January 18:00', fmt);
// works fine: d.isValid() is true

d = dayjs('Monday 15 January 18:00', fmt);
// works fine: d.isValid() is true

d = dayjs('Saturday 20 January 18:00', fmt);
// works fine: d.isValid() is true


waseemahmad31:
test case failing with moment 2.24.0. 

I have checked with moment 2.5.1. It's working.
https://jsfiddle.net/pkje2y6z/

BePo65:
What would be the use of parsing the weekday?

Wouldn't it be better to parse variable length text (e.g. the name of a weekday) like in issue #571 ?

PR #580, opened 2019-04-25
fix: Allow customizing "am" / "pm" strings with locale meridiem function

### LINKED ISSUE #578: Allow customizing "am" / "pm" strings with locale
Some organizations specifically require the use of "a.m." and "p.m." instead of "am" and "pm". It would be very useful if the two strings could be customized using a custom locale -- they appear to be hard-coded right now.

iamkun:
fix #578

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -279,7 +279,7 @@ class Dayjs {
     const zoneStr = Utils.z(this)
     const locale = this.$locale()
     const {
-      weekdays, months
+      weekdays, months, meridiem
     } = locale
     const getShort = (arr, index, full, length) => (
       (arr && arr[index]) || full[index].substr(0, length)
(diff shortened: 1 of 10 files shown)


src/locale/ja.js
```diff
@@ -19,6 +19,7 @@ const locale = {
     lll: 'YYYY年M月D日 HH:mm',
     llll: 'YYYY年M月D日(ddd) HH:mm'
   },
```
muhammadridho: is it evey locale have this method ?

src/index.js
```diff
@@ -305,8 +310,8 @@ class Dayjs {
       HH: Utils.s(this.$H, 2, '0'),
       h: get$H(1),
       hh: get$H(2),
-      a: this.$H < 12 ? 'am' : 'pm',
-      A: this.$H < 12 ? 'AM' : 'PM',
+      a: meridiemFunc(this.$H, this.$m, true),
```
muhammadridho: what if you change the name `a` into `am` thats readable object or key, wdyt ?

xxyuk: I think no. This lib is designed to keep the same with moment.js

src/locale/ja.js
xxyuk: if locale need other than am/pm might need this. moment.js'locale file is a good reference

src/index.js
muhammadridho: got it

src/locale/ja.js
muhammadridho: lgtm, if we have updated documentation too 👍

gka:
yay, thanks!

PR #584, opened 2019-05-02
Fix error in relativeTime plugin when the interval is 1 second

jansons:
Fixes bug introduced in #558


iamkun:
@jansons thanks, and what's this fix for?

jansons:
@iamkun There was a runtime TypeError when the calculated interval is 1 second.
You can try to run updated tests on v1.8.13 to get the error

iamkun:
Nice fix, thanks

eamodio:
@iamkun will there be a new release with this fix soon?

iamkun:
of course 😁

Issue #171, opened 2018-05-19
Locale and Documentation Contributions Wanted 🙋🙏


iamkun:
Release 1.6.2 was just published to the NPM with I18n support. Thanks again, everyone, for helping with this effort.

Day.js users come from all around the world, so I will happily accept translations from other languages for this project. And your work will affect thousands of daily users of Day.js.

Feel free to open a pull request if you've finished one of the three translations. There's no need to finish all of the three in one time.

## Locale (I18n)
Locale is the first translation work we need to do, and it will be released with our main code.
[Locale template](https://github.com/iamkun/dayjs/blob/master/docs/en/I18n.md#customize)

## Readme file
Readme file let user get a glance of our project. And tells them what we are doing.
If there is no sub pages' translation, please keep the links in readme linked to English sub pages.
Add a link to your translation in the default english readme file.
[Default English Readme](https://github.com/iamkun/dayjs/blob/master/README.md)

## Sub pages of the document
Sub pages is the detailed documents of our project.
[Sub pages folder](https://github.com/iamkun/dayjs/blob/master/docs)

## Directory structure
Please place the translation files in a separate folder, and named with ISO 639-1 area code. Directory structure could be found below.
```
.
├── docs
│   └── es
│        ├── API-reference.md
│        ├── I18n.md
│        ├── Installation.md
│        ├── Plugin.md
│        └── README-es.md
├── src
│   └── locale
│        └── es.js
└── README.md  // default index english readme
```

Finally, thanks again for your contribution.

iamkun:
ping @ohbarye, and seems you've already done it. Bravo!

ludvon:
I'm working on the Brazilian Portuguese translation!

iamkun:
@ludvon 

> Is it OK to slightly modify some of the docs' wording

Of course. It's up to you.

But better keep the document's structure the same, so that I could update them easily if there is some small changes like API name or agreements change.

KorbinianKuhn:
We could convert the momentjs locales to the dayjs format. I tried it automatically and it worked for most of them (there are some languages which require additional information for month name handling and the ordinal function), but we could use all others as a basis. @iamkun what do you think?

iamkun:
@KorbinianKuhn Thanks, this should work, but little copycat I think.

aalises:
@iamkun for the fromNow plugin to work, I think the **relativeTime** variable should also be added to the locale template? Right now it is implemented on the spanish locale and the default english one, e.g
```
[...]

PR #594, opened 2019-05-13
Add fil locale (same as tl-ph)

aadityataparia:
Filipino language (`fil`) is synonym to tagalog language (`tl-ph`)

https://en.wikipedia.org/wiki/Filipino_language#Filipino_vs._Tagalog

### DIFF
--- a/src/locale/fil.js
+++ b/src/locale/fil.js
@@ -1,0 +1,7 @@
+import dayjs from 'dayjs'
+import tlLocale from './tl-ph'
+
+const locale = Object.assign({}, tlLocale, { name: 'fil' })
+dayjs.locale(locale, null, true)
+
+export default locale


iamkun:
Thanks. 

One more question, what's the difference between Philippine languages (phi), fil and tl-ph, and what's tl-ph stand for (is it two combined language)

aadityataparia:
They are basically same, and `tl-ph` is Tagalog (Philippines), `tl` alias can also be added if we want that too. They are just different ISOs, `tl` -> two letter language code, `tl-ph` langauge + country, `fil` -> three letter language code. 

I wanted to add `fil` because I am using it in i18n.

Screen Shot 2019-05-13 at 12 17 19

iamkun:
As mentioned in #171, we use https://en.wikipedia.org/wiki/List_of_ISO_639-2_codes as a reference. 
We use tl-ph to keep the same with moment.js, surprisingly, it's not listed on that table (but fil does)

aadityataparia:
> As mentioned in #171, we use https://en.wikipedia.org/wiki/List_of_ISO_639-2_codes as a reference.
We use tl-ph to keep the same with moment.js, surprisingly, it's not listed on that table (but fil does)

Rename `tl-ph` to `tl` maybe. Or create alias to keep compatible with moment.

iamkun:
Found two related issue / pr in moment.js 

https://github.com/moment/moment/issues/3054

https://github.com/moment/moment/pull/4239

aadityataparia:
Tried this today, even this works 
```js
dayjs.locale('fil', dayjs.Ls['tl-ph'])
```

PR #597, opened 2019-05-13
fix: Pad formatted year string with zeros

at-soft:
This pull request fixes formatting of date with pattern "YYYY" for years less than 1000

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -295,7 +295,7 @@ class Dayjs {
 
     const matches = {
       YY: String(this.$y).slice(-2),
-      YYYY: this.$y,
+      YYYY: Utils.s(this.$y, 4, '0'),
       M: $M + 1,
       MM: Utils.s($M + 1, 2, '0'),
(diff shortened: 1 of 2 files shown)


hc-12:
I just stumbled into this issue. It would be nice to get this merged in

cokkto:
Not really good. You can pad number on demand in your particular project but internationally years before 1000 are used unpadded

at-soft:
> Not really good. You can pad number on demand in your particular project but internationally years before 1000 are used unpadded

My intention is to keep compatibility with moment.js API, which has `YYYY` format for 4-digit years and `Y` to display unpadded year. Moreover, current formatting leads to some errors like

```js
const initial = new Date('0001-01-01');
const initial_dayjs = dayjs(initial);
const formated_initial = initial_dayjs.format('YYYY-MM-DD'); // produces 1-01-01
const parsed = dayjs(formated_initial);
const formated_result = initial_dayjs.format('YYYY-MM-DD'); // produces 2001-01-01
```

Issue #492, opened 2019-02-12
require doesn't support `default` exports in ES Modules in Webpack


iamkun:
Day.js added `module` in `package.json` in v1.8.1+ , in order to support ES Modules.

However, there's some unexpected behavior while using commonJS  `require` with `webpack`

And there a discussion from `webpack` here, https://github.com/webpack/webpack/issues/4742

```
const dayjs = require('dayjs') // works with 1.8.0 - 


const dayjs = require('dayjs') // broken with 1.8.1 +
const dayjs = require('dayjs').default // works with 1.8.1 +
```

Note: add `.default` is just a temporary workaround, and it's still a bug with webpack we think. Since `default` is an artifact of ES Modules and should not appear in CJS requires

mweststrate:
@iamkun no, this is what was specified in your source code, [here](https://github.com/iamkun/dayjs/blob/f14208244b97f2f950a740e48c1fd8ec74758c4a/src/index.js#L416).
If you don't want a `.default` when using common JS, the exports should be written as `module.exports = dayJS` (which has no ESM equivalent). The reason why it worked before, is that because if you were exporting commonJS, babel was _assuming_ that with `require('dayjs')` you were meaning `require('dayjs').default` (it generates some feature detecting code for that), which so far made seemed that require without `default` was the commonJS  equivalent, but it really isn't

iamkun:
@mweststrate Yes, that's why we have two bundles `main: /dayjs.min.js` is the umd bundle for commonjs, while `module: /esm/index.js` is for the ESM.

And webpack just load the `module: /esm/index.js` no matter it is `require` or `import`.

mweststrate:
That is what one would expect right? Otherwise your lib might end up twice
in the bundle. The problem seems that your minified bundle does _not_ have
a `.default` export, so effectively you are shipping a different API in
your commonJS version than in your ESM version

Op di 12 feb. 2019 om 16:17 schreef iamkun <notifications@github.com>:

githoniel:
> For ESM the "namespace" object is returned by `require()` (and for `import()`). There is no spec for this, but there is a spec for `import()`. Babel also transpiles ESM this way.
> Exposing only the default export would make all other export inaccessible and most packages un-usable from CJS.
> 
> As a guideline: When packaging CJS and ESM in a package, expose the same API from both of them. You can't use a function assigned to `module.exports`.
> 
> ```js
> exports default "something"
> export var named = "other";
> ```
> 
> ```js
[...]

PR #612, opened 2019-06-07
With module you can define the path to ES module

kevinsimper:
The module attribute is too define ES module so that loaders can find the unpackaged code.

https://github.com/rollup/rollup/wiki/pkg.module

https://pikapkg.com is looking for ES modules in that field

https://unpkg.com serves that file if you define that it should load with `?module`

### DIFF
--- a/package.json
+++ b/package.json
@@ -4,7 +4,7 @@
   "description": "2KB immutable date time library alternative to Moment.js with the same modern API ",
   "main": "dayjs.min.js",
   "types": "index.d.ts",
-  "module": "dayjs.min.js",
+  "module": "esm/index.js",
   "scripts": {
     "test": "TZ=Pacific/Auckland npm run test-tz && TZ=Europe/London npm run test-tz && TZ=America/Whitehorse npm run test-tz && npm run test-tz && jest",
     "test-tz": "date && jest test/timezone.test --coverage=false",


iamkun:
#492

kevinsimper:
@iamkun I read it now but it makes bundler that looks for ES Modules code find the wrong code that doesn't use the correct syntax. That bundlers can figure out is not a great solution.

The day.min.js doesn't expose any of the ES Module export syntax, I don't see how this is a solution in #492 😅

iamkun:
I am afraid there's no way to fix this issue until webpack fix that related issue first. Let me know if I'm wrong.

kevinsimper:
The solution can also be to publish two packages, as @Tvrqvoise says, I don't think you can solve both, but that is why you have two entrypoints, one for node.js and one for ES modules :)

kevinsimper:
I published a new npm package that people can use easily in the browser
https://unpkg.com/dayjs-es?module

Static example can be seen here https://codesandbox.io/s/static-dcus0

kevinsimper:
Here is also a discussion about how to tackle it :) https://twitter.com/_developit/status/1138115151318704128

iamkun:
I'll close this since it's been a while since it's been opened. Feel free to reopen if you have updates on this

kevinsimper:
@iamkun Isn't the three previous comments enough? 😅

PR #617, opened 2019-06-18
modify types of diff

pochi-sato:
If passing only one argument to a dayjs diff method, it is said that the type is not correct.
So I fixed it.

-  document
image

- code
image


iamkun:
THX

PR #626, opened 2019-07-08
Add optional properties to DA locale

DerJacques:
This PR adds the optional properties `weekdaysShort`, `weekdaysMin` and `monthsShort` to the Danish locale.

All changes have been crosschecked using a Danish dictionary:
https://ordnet.dk/ddo/ordbog

### DIFF
--- a/src/locale/da.js
+++ b/src/locale/da.js
@@ -3,7 +3,10 @@ import dayjs from 'dayjs'
 const locale = {
   name: 'da',
   weekdays: 'søndag_mandag_tirsdag_onsdag_torsdag_fredag_lørdag'.split('_'),
+  weekdaysShort: 'søn._man._tirs._ons._tors._fre._lør.'.split('_'),
+  weekdaysMin: 'sø._ma._ti._on._to._fr._lø.'.split('_'),
   months: 'januar_februar_marts_april_maj_juni_juli_august_september_oktober_november_december'.split('_'),
+  monthsShort: 'jan._feb._mar._apr._maj_juni_juli_aug._sept._okt._nov._dec.'.split('_'),
   weekStart: 1,
   ordinal: n => `${n}.`,
   formats: {


src/locale/da.js
```diff
@@ -3,7 +3,10 @@ import dayjs from 'dayjs'
 const locale = {
   name: 'da',
   weekdays: 'søndag_mandag_tirsdag_onsdag_torsdag_fredag_lørdag'.split('_'),
+  weekdaysShort: 'søn._man._tirs._ons._tors._fre._lør.'.split('_'),
+  weekdaysMin: 'sø._ma._ti._on._to._fr._lø.'.split('_'),
   months: 'januar_februar_marts_april_maj_juni_juli_august_september_oktober_november_december'.split('_'),
+  monthsShort: 'jan._feb._mar._apr._maj_juni_juli_aug._sept._okt._nov._dec.'.split('_'),
```
iamkun: thanks, how about removing all the ending dot to keep the same with https://github.com/moment/moment/blob/develop/locale/da.js ?

DerJacques: Thanks for the comment!
I would argue that the dots are part of the way short versions of months are to be displayed in Danish. Similar to the way it's handled in Norwegian (Bokmål):
https://github.com/iamkun/dayjs/blob/dev/src/locale/nb.js

You can also see the format used in the Wikipedia article about date formatting in Denmark:
https://en.m.wikipedia.org/wiki/Date_and_time_notation_in_Denmark

iamkun: According to Wikipedia, seems not dot as well?:
```
lørdag, den 1. januar 2000 
tirsdag, den 25. juni 2019
lørdag 1. januar 2000 
tirsdag 25. juni 2019
```

DerJacques: That's the examples with full months :) I only changed the short versions that use abbreviations. From the Wikipedia article:

1. jan. 2000 (versus 1. januar 2000)
25. jun. 2019 (versus 25. juni 2019)

iamkun: Oh I see.

PR #636, opened 2019-07-18
feat: Improving '.set' applied multiply changes

afranioce:


iamkun:
Added vie ObjectSupport Plugin https://day.js.org/docs/en/plugin/object-support

PR #642, opened 2019-07-26
fix: add meridiem locale for ko

cheolheon:
Hey, at the first, thank you for great and modern date library! 

This PR is really simple - just adding meridiem locale for 'ko'.

Korean has its unique meridiem notation '오전'(AM) and '오후'(PM) but dayjs doesn't provide it yet so I add it!

Have a nice day.


iamkun:
THX

PR #643, opened 2019-07-26
Add missing weekdaysShort in translation files

vigosan:
This PR adds weekdaysShort to those locale files that they do not have this variable defined.


iamkun:
THX

PR #649, opened 2019-07-31
Fixes issue#646

### LINKED ISSUE #646: fromNow() gives inconsistent results between UTC & local mode
**Describe the bug**
I've found a situation where `fromNow()` gives different results from MomentJS, and it gives different results depending on if you're in UTC mode or not.

Given the following:

```javascript
let dutc = dayjs.utc(currentTimestamp37hrsAgo); // UNIX time for 37 hours before now
let mutc = moment.utc(currentTimestamp37hrsAgo);

console.log(`37 hours ago - DayJS (utc):     ${dutc.toISOString()}, fromNow(): ${dutc.fromNow()}`);
console.log(`37 hours ago - MomentJS (utc):  ${mutc.toISOString()}, fromNow(): ${mutc.fromNow()}`);
```

Output of the above:

```
37 hours ago - DayJS (utc):     2019-07-28T09:28:35.873Z, fromNow(): a day ago
37 hours ago - MomentJS (utc):  2019-07-28T09:28:35.873Z, fromNow(): 2 days ago
```

Here's a [more extensive reproduction on Runkit](https://runkit.com/accelerate/fromnow)

If it helps, my local timezone is PST (-07:00)

**Expected behavior**
It is my understanding that `fromNow()` should not "care" about timezones, and it's based strictly off of the difference in absolute time. `fromNow()` should give the same results, regardless of if we're using `.utc()` or not. I didn't check to see if `toNow()` also has the same problem, but I'd assume so.

**Information**
 - Day.js Version: `1.8.15`
 - OS: `MacOS 10.14.6`
 - Browser: `Chrome 75.0.3770.142`
 - Time zone: `-07:00 DST (Pacific Daylight Time)`

g1eny0ung:
close #646

### DIFF
--- a/package.json
+++ b/package.json
@@ -8,7 +8,6 @@
   "scripts": {
     "test": "TZ=Pacific/Auckland npm run test-tz && TZ=Europe/London npm run test-tz && TZ=America/Whitehorse npm run test-tz && npm run test-tz && jest",
     "test-tz": "date && jest test/timezone.test --coverage=false",
-    "test:jest": "jest",
     "lint": "./node_modules/.bin/eslint src/* test/* build/*",
     "prettier": "prettier --write \"docs/**/*.md\"",
     "babel": "cross-env BABEL_ENV=build babel src --out-dir esm --copy-files && node build/esm",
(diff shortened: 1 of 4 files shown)


package.json
```diff
@@ -8,6 +8,7 @@
   "scripts": {
     "test": "TZ=Pacific/Auckland npm run test-tz && TZ=Europe/London npm run test-tz && TZ=America/Whitehorse npm run test-tz && npm run test-tz && jest",
     "test-tz": "date && jest test/timezone.test --coverage=false",
+    "test:jest": "jest",
```
iamkun: why adding this?

PR #653, opened 2019-08-02
Fix TS2339: Property 'week' does not exist on type 'Dayjs'.

kingller:
Fix `TS2339: Property 'week' does not exist on type 'Dayjs'` when in `TypeScript`

### DIFF
--- a/types/index.d.ts
+++ b/types/index.d.ts
@@ -51,6 +51,8 @@ declare namespace dayjs {
 
     millisecond(value: number): Dayjs
 
+    week(): number
+
     set(unit: UnitType, value: number): Dayjs
 
     get(unit: UnitType): number


iamkun:
Please use weekOfYear plugin to get .week() API

kingller:
@iamkun OK. The `weekOfYear` need to be added as the same error
```
Property 'weekOfYear' does not exist on type 'Dayjs'
```

iamkun:
@kingller https://github.com/iamkun/dayjs/blob/dev/docs/en/Plugin.md#weekofyear

Please follow the doc here.

kingller:
Oh, sorry, it's my mistake. It's OK. Thanks very much.

Issue #610, opened 2019-06-06
Please fix declension in relative time


VBagmut:
**Describe the bug**
Thanks for fixing https://github.com/iamkun/dayjs/issues/577. But I found another bug. This time with relativeTime plugin.

**Expected behavior**
It is incorrectly using declension. For example: "2 dayjs ago" needs to be "2 дня назад" but instead shows "2 дней назад". How it needs to be described here http://www.unicode.org/cldr/charts/28/supplemental/language_plural_rules.html#ru
I hope you will fix this and then I'll be able to throw away heavyweight momentjs. Because now I need to use moment for this cases.

**Information**
 - Day.js: 1.8.14
 - OS: Ubuntu 18.04.2 LTS
 - Browser: Google Chrome 74.0.3729.169 (Official Build) (64-bit)
 - Time zone: GMT+03:00

moosemanf:
Same in German...

altough it's elegant in english, it is not possible in german:

```
d.en.relativeTime = {
    future: 'in %s',
    past: '%s ago',
    y: 'a year',
    yy: '%d years'
  }
```
```
  if (withoutSuffix) return out
    return ((result > 0) ? loc.future : loc.past).replace('%s', out)
```
we would need (for example) a `yy` for `withoutSuffix == true` and another `yy` for `withoutSuffix == false`

### withoutSuffix
3 Jahre

### withSuffix
vor 3 Jahren 

#304 there is a good starting solution, but unfortunately no progress...

stalkerg:
I am also interested in that feature. It's completely unnatural for Slavic languages (and rule a little bit more complicated than in English).

Issue #416, opened 2018-11-27
Non-Arabic Numerals


bcmn:
Our specific use case requires for conditionally displaying "[eastern arabic numerals](https://en.wikipedia.org/wiki/Eastern_Arabic_numerals)" `۰۱۲۳۴۵۶۷۸۹` instead of "[arabic numerals](https://en.wikipedia.org/wiki/Arabic_numerals)" `0123456789`. Presumably any implementation would be intended to be easily extensible for any other commonly-used base 10 numeral systems.

This would result in DayJS strings like:
```
۲۸ آوریل ۲۰۱۸
پیش ساعت ۴
```
instead of (the current behaviour):
```
28 آوریل 2018
پیش ساعت 4
```

This issue is primarily to ascertain whether this is something that other people would feel is useful or necessary to have as a plugin for DayJS, though there are also some implementation issues that might be useful to discuss.

Obviously the alternative is that this could remain a bespoke implementation detail, & our project just treats this as a generic string formatting operation, completely unaware of DayJS.

--- 

### Implementation Thoughts

After a quick investigation, I'm not sure DayJS currently has a way of applying this sort of library-wide formatting operation to all strings returned? 

It is very easy to add the behaviour to any one function, e.g. `.format` or the overridden `.format` function of the `advancedFormat` plugin, but this would very noticeably miss applying the logic to other plugins, such as `relativeTime`. Is there a better solution than this?

I'm aware `advancedFormat` currently uses binding to achieve some conditional formatting. There _may_ be a way to do this for all plugins that might return a string containing a numeral, but it doesn't seem ideal. 🤔

iamkun:
You may could extend `advancedFormat ` by adding your own tokens, like `DEMO`.

```
import advancedFormat from 'dayjs/plugin/advancedFormat' 
import youOwnFormat from 'dayjs/plugin/youOwnFormat' 
dayjs.extend(advancedFormat) 
dayjs.extend(youOwnFormat) 
dayjs().format('DEMO YYYY - MM') 
```

kalyantm:
@iamkun  Can u please elaborate? I think I'm in need of a similar solution

xxyuk:
@bcmn You might could simply do a global replace('1234', '۰۱۲۳۴۵۶۷۸۹') on the final `format` result .

This might be easy and without a dayjs plugin.

VehpuS:
@bcmn - If still relevant, I believe #1255 solves this issue (at least that's how it's done in moment.js).

[...]

PR #659, opened 2019-08-13
Feature: Relative time for fusional languages

tomas-ki:
Hi,

Currently there are several issues regarding relative time on fusional languages, this PR would address these issues (I hope I didn't miss any):
https://github.com/iamkun/dayjs/issues/610 https://github.com/iamkun/dayjs/issues/416 https://github.com/iamkun/dayjs/issues/415 https://github.com/iamkun/dayjs/issues/302

I have looked through moment implementation of relative time and adapted it to fit dayjs. I am aware that a lot of the this code is taken from moment, but I don't think that reinventing the wheel in this place is a good approach. I think it is okay to have moment as a point of truth for proper localization results, because translation is a very long process and this makes it easy to have correct translations even without knowing the language.

There are a couple of fixes in language translations (polish for example was completely off, since translating relative time to the past translates it as a future term, see "po 1 godziny" for example.).

I have also added preparsers/postparsers when needed. In this case Arabic locale numerals.

I've added tests and some helpers to reduce code duplication and add coverage to localized formats. Current implementation makes it easy to compare dayjs localization results with moment. There are some cases that are difficult to cover, so I'm leaving it not as 100% for now.

I've also added handling for relative time formats as functions, which addresses plural forms in most cases.

Thanks in advance for taking a moment (no pun intended) to look into this and please voice any concerns that you may have.


iamkun:
Thanks, but here's simple one. #767

it-xtech-dev:
Polish seems to be not handled in #767.

iamkun:
> Polish seems to be not handled in #767.

fix in d3724758bb27d5b17587b995ba14e7e80dcd1151

it-xtech-dev:
@iamkun how to know when / whether this fix is available in release (and which version)?

iamkun:
next coming version

PR #661, opened 2019-08-15
remove ( ) from arrow functions which have a single pram

catastrop:


iamkun:
this does not follow our es-lint rule

PR #665, opened 2019-08-23
Update zh-cn.js

Jackong:
The weekStart is 1, but the weekdays weekdaysShort and weekdaysMin is not


iamkun:
the dayjs.format function will deal with this

PR #672, opened 2019-09-01
Fix SL abbreviation for month may

janezkranjc:
The SL abbreviation for the month May is wrong. Slovenian word for May is `maj` meaning that it is already only 3 characters long so it doesn't need the trailing `.` which denotes an abbreviation.

### DIFF
--- a/src/locale/sl.js
+++ b/src/locale/sl.js
@@ -6,7 +6,7 @@ const locale = {
   months: 'januar_februar_marec_april_maj_junij_julij_avgust_september_oktober_november_december'.split('_'),
   weekStart: 1,
   weekdaysShort: 'ned._pon._tor._sre._čet._pet._sob.'.split('_'),
-  monthsShort: 'jan._feb._mar._apr._maj._jun._jul._avg._sep._okt._nov._dec.'.split('_'),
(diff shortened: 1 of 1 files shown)


iamkun:
Thanks for this PR. Day.js is intent to keep the same with Moment.js. And according to https://github.com/moment/moment/blob/develop/locale/sl.js, there's also a trailing `.`.

It will be great if you could send a PR to them and update both.

janezkranjc:
Yeah, I did - on 28th of July but they are very slow to respond.

https://github.com/moment/moment/pull/5180

iamkun:
OK. In this way, we don't mind merge this PR first. Would you please add some relevant links like wikipidia or online dictionary.

janezkranjc:
So I've been doing a bit of research and apparently the abbreviation is ok because when a calendar date is abbreviated in such a way the month is in the second declination which would make the word: `maja` and gets abbreviated to `maj.`

So, maybe it's best to just close this and not merge it.

Here's a source (in slovenian)
https://www.lektorsko-drustvo.si/jezikovni-pogovori/pisanje-datumov-v-slovenscini

iamkun:
THX

PR #705, opened 2019-10-25
Fall back relative time to english if cant find loaded locale

lkostrowski:
Currently dayjs is not actually "safe" in terms of usage.

I know there is a goal of keeping size limit, however maybe there is a reasonable way to improve code quality and move code compiling to some super-minifier like Closure compiler, without omitting code security and overall readability (variable names etc)?

In this PR I do fallback to english locale in relativeTime plugin, which throws without reasonable message if locale isn't loaded properly. I also show error in console. I have expected behavior like this since I failed loading language in my project and it didn't inform me about it.

Don't know if this is something expected in this library or it's very "raw" deliberately

### DIFF
--- a/src/plugin/relativeTime/index.js
+++ b/src/plugin/relativeTime/index.js
@@ -18,7 +18,15 @@ export default (o, c, d) => {
     yy: '%d years'
   }
   const fromTo = (input, withoutSuffix, instance, isFrom) => {
-    const loc = instance.$locale().relativeTime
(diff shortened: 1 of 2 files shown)


iamkun:
Sorry for the late reply, I just noticed this PR.

The better way I think, is to make sure each locale has it's `relativeTime` locale config or simply copy from English locale if they are sharing the same result.

Issue #707, opened 2019-10-28
Web-based Documentation


denisinvader:
It's quite uncomfortable to access docs through GitHub files. It would be nice to have a web version of docs with search and sections. 

I suggest using [Vuepress](https://vuepress.vuejs.org) and Github pages or an external server and domain name.

As far as I can see, the migration of current docs to Vuepress will be rather simple because of its structure.

I could integrate Vuepress in this

denisinvader:
#709

iamkun:
Yes we have one ! https://day.js.org/

PR #709, opened 2019-10-29
Add Vuepress for docs

denisinvader:
#707


iamkun:
Thanks very much. Actually we have a test repo to build a doc website with i18n here https://github.com/dayjs/website. 

It will be great if you could join us.

iamkun:
https://day.js.org/

PR #710, opened 2019-10-30
Minor localisation change

Kashkovsky:
relativeTime.past is defined as '%s назад' which is Russian translation. Correct Ukrainian localisation is '%s тому'.


iamkun:
Thanks

PR #713, opened 2019-10-31
Fix polish locale

ar4s:
`d` and `dd` is responsible for day and days. In polish day means dzień, week means tydzień.


iamkun:
Thanks

PR #727, opened 2019-11-10
Italian locale relative time translation fix

guidolodetti:
"3 hours ago"

Currently it is translated to "da 3 ore". It is grammatically correct, but it has a different meaning (it means that something has been occurring for 3 hours, not that occurred 3 hours ago).

The correct translation would be "3 ore fa". MomentJS translates it in the same way.


iamkun:
Looks good! Can you cherry-pick and submit your commit only, please?

guidolodetti:
Hi @iamkun! Can you please suggest me how to do it? I have never done it before. Thanks!

iamkun:
@guidolodetti  You can try `merge origin/dev` first.

PR #756, opened 2019-12-20
If match is null, match[0] will throw an exception error and it will …

reyramos:
…not proceed.

"TypeError: Cannot read property '0' of null
    at node_modules\dayjs\plugin\customParseFormat.js:99:70


iamkun:
Test failed. Can you provide a test case, please?  Or the related incorrect parse format.

reyramos:
The test fails regardless of my change of not, The change is a validation that match[0] is not undefined.

PR #776, opened 2020-01-15
add unit value: date support for `.add()` & `.diff()` function and perfect reference

TomIsion:
fix: Add unit quarter explain to API reference
fix: Add unit date for `.add()` & `.diff()` function

### DIFF
--- a/docs/en/API-reference.md
+++ b/docs/en/API-reference.md
@@ -208,6 +208,7 @@ dayjs().get('day')
 | `minute`      | `m`       | Minute                                   |
 | `second`      | `s`       | Second                                   |
 | `millisecond` | `ms`      | Millisecond                              |
+| `quarter`     | `Q`       | Quarter (Only support in `.diff()`)      |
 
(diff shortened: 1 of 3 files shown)


iamkun:
Why should we support `add date` API, please?

TomIsion:
Ehh... No need.

But from current API reference: as [units](https://github.com/iamkun/dayjs/blob/dev/docs/en/API-reference.md#list-of-all-available-units) define, I think `date` is **available** unit for `.add()`

when I use, I find `.add(1, 'date')` no use...

---

Append: `.diff()` support 'quarter' unit but not mention in API reference.

iamkun:
> .diff() support quarter unit but not mention in API reference.
Thanks, it's better to add this in another pull request.

TomIsion:
> Thanks, it's better to add this in another pull request.

Yep.

No need support add date API -> Maybe you should clear every api unit list *like moment api does*.

iamkun:
Let's pending this PR and a better doc website is on the way

PR #780, opened 2020-01-18
Add Bislama Locale

michaeltoohig:
Bislama is the national language of Vanuatu. 

Perhaps there may be a rule about `weekdaysMin` being 2 characters but I needed 3 characters for Sunday/Saturday. Same for my choice of using a 4 character `weekdaysShort` for Friday. The `ae` in `Fraede` is a diphthong and would be wrong I believe if the vowels are not kept together.

### DIFF
--- a/src/locale/bi.js
+++ b/src/locale/bi.js
@@ -1,0 +1,39 @@
+import dayjs from 'dayjs'
+
+const locale = {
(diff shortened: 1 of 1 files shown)


iamkun:
Thanks

PR #786, opened 2020-01-24
Fix Russian locale

AdvCake:
This is a little fix in the short versions of the months names. Now's everything according to the proper Russian language grammar :)


AdvCake:
Honestly, I don't really know why the Travis CI build failed... I just tried to build the project with my changes locally and everything seemed to work

iamkun:
This ru locale is keep updated with moment.js https://github.com/moment/moment/blob/develop/locale/ru.js

any problem so far? a demo code please

AdvCake:
@iamkun Hey! Thanks for your response!

The thing is it's not about the code here, it's about the grammar. I am a native Russian speaker and I know for a fact that when you use short version of the month name you never end it with a dot, as well as short version of "November" is "ноя", not "нояб.". 

Would you like me to send the pictures of Russian calendars or smth to prove to you that apparently moment.js has wrong localization?

iamkun:
@AdvCake  I see.
It will be great if you could open a pull request on moment.js as well. 
So that we could merge it together.

PR #787, opened 2020-01-27
optimize isSame method if units is not provided

cokkto:
This will have about x5 performance boost https://jsfiddle.net/cocteau/wyqrpgxv/36/

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -103,7 +103,9 @@ class Dayjs {
 
   isSame(that, units) {
     const other = dayjs(that)
-    return this.startOf(units) <= other && other <= this.endOf(units)
+    return units !== undefined ?
+      this.startOf(units) <= other && other <= this.endOf(units) : 
+      this.unix() === (other || {}).unix();
   }
 
   isAfter(that, units) {


iamkun:
But will also increase the bundle size 😿

cokkto:
Few bytes in a one of the most used method are worth this, aren't they?

src/index.js
```diff
@@ -103,7 +103,9 @@ class Dayjs {
 
   isSame(that, units) {
     const other = dayjs(that)
-    return this.startOf(units) <= other && other <= this.endOf(units)
+    return units !== undefined ?
+      this.startOf(units) <= other && other <= this.endOf(units) :
+      this.valueOf() === (other || {}).valueOf()
```
iamkun: what's this {} for, pls?

cokkto: ` (other || {}) ` this is a shorthand to avoid null reference exception calling `.valueOf` if `other` is null/undefined

cokkto: basically can be easily removed if by design dayjs doesn't care about excpetions thrown by incorrect input, resulting in just `this.valueOf() === other.valueOf()`

iamkun: other will always be a dayjs obejct with no exception I think

cokkto: not really, at least it can be null value, or, in loosely-typed javascript world, basically whatsoever user could only send. `this` is an dayjs object always and need no extra check, but input parameter is absolutely everything. And null check, as I mentioned, is up to you do you think is error here expected on wrong input or not.

iamkun: dayjs(null).valueOf() -> NaN (no exception)

cokkto: example should be: `dayjs().isSame(undefined)`, because we are testing `other`.

iamkun: yes, that why we have `const other = dayjs(that)` that is `dayjs(undefined)`

cokkto: i see, updated

iamkun:
should we consider unit 'ms' as well?

cokkto:
> should we consider unit 'ms' as well?

I wouldn't. 1 - extra check adds extra bytes with no value (user who is interested in ms comparison should rather leave unit parameter empty) 2 - for unification purpose lets all calls with unit specified follow the same chain

addisonElliott:
Sorry to revive this old PR, but I think the `ms` units should be considered. Tested with latest DayJS and `.isSame` takes into account the ms. I don't think it's wise to revert that logic. Could have unexpected consequences for people.

Updated the logic to include `&& this.$ms === other.$ms` and there was a slight hit in performance, but not much. (unix tests with ms and unix2 doesn't).

```
"isSame() 35.94000008888543 ms"
"unix() 10.174999944865704 ms"
"unix2() 8.84499994572252 ms"
```

Those were my results, so we're talking about a 0.6us difference in time, pretty minimal to me. In addition, if performance were an absolute must, then the function could be updated in the future to do the above logic if units are seconds.

This update will take more space, but that's not important to me. This library is much smaller than moment and I'm mainly using it in Node.js where I want performance over size.

PR #790, opened 2020-01-30
feat: Add correct Estonian relativeTime locale for past and future tense

HarlesPilter:
Since https://github.com/iamkun/dayjs/pull/767 got merged, I added correct future tense for Estonian locale.

### DIFF
--- a/src/locale/et.js
+++ b/src/locale/et.js
@@ -1,6 +1,39 @@
 // Estonian [et]
 import dayjs from 'dayjs'
 
+function relativeTimeWithTense(number, withoutSuffix, key, isFuture) {
+  const pastTense = {
+    s: 'mõni sekund',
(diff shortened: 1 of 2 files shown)


src/locale/et.js
```diff
@@ -1,6 +1,26 @@
 // Estonian [et]
 import dayjs from 'dayjs'
 
+function relativeTimeWithTense(number, withoutSuffix, key, isFuture) {
+  const format = {
+    s: ['mõne sekundi', 'mõni sekund', 'paar sekundit'],
+    ss: ['%d sekundi', '%d sekundit'],
```
iamkun: Think we don't need this line?

HarlesPilter: Yeah, I wondered why that was there since I don't think moment uses it either.

PR #797, opened 2020-02-15
fix: Relative times in Finnish locale

ttsirkia:
This fixes problems in Finnish translation (#351) which are now possible to solve after merging #767. Also, the output is now the same as in Moment.js.

### DIFF
--- a/test/locale/fi.test.js
+++ b/test/locale/fi.test.js
@@ -1,4 +1,5 @@
 import MockDate from 'mockdate'
+import moment from 'moment'
 import dayjs from '../../src'
 import relativeTime from '../../src/plugin/relativeTime'
 import '../../src/locale/fi'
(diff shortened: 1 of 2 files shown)


test/locale/fi.test.js
```diff
@@ -0,0 +1,27 @@
...
+  MockDate.set(new Date())
+})
+
+afterEach(() => {
+  MockDate.reset()
+})
+
+it('Finnish locale relative time in past and future', () => {
+  expect(dayjs().add(1, 'd').locale('fi').fromNow())
+    .toBe('päivän kuluttua')
```
iamkun: it will be great if you could add moment.js result in test like: 

```
expect(dayjs().add(1, 'd').locale('fi').fromNow())
    .toBe('päivän kuluttua')
expect(dayjs().add(1, 'd').locale('fi').fromNow())
    .toBe(moment().add(1, 'd').locale('fi').fromNow())
```

ttsirkia: Those are actually not equal because the original translation here uses word ”kuluttua” and Moment ”päästä”. They are synonyms.

ttsirkia: The word used here is actually better but should they be the same?

iamkun: Yes, please. Day.js trying to keep everything the same with moment.js at the present. 

You can check other locales' test file as a reference.

ttsirkia: There seems to be another difference, too. Dayjs tells "2 päivää sitten" and Moment "kaksi päivää sitten". So for some reason, Moment uses words for numbers less than ten. Should this be replicated as well? This seems to be specific for Finnish locale although using integers would be fine.

ttsirkia: This difference has already been present here when the original Finnish translation is merged.

ttsirkia: I changed these.

PR #799, opened 2020-02-18
Handle value setting in dayOfYear call signature

tpict:
Minor bug fix. The current version doesn't handle the case where the date is being set. Thanks!


iamkun:
THX

PR #810, opened 2020-03-02
fix german weekdays

flo-l:
the week starts with monday by convention in German speaking countries.

### DIFF
--- a/src/locale/de-at.js
+++ b/src/locale/de-at.js
@@ -3,9 +3,9 @@ import dayjs from 'dayjs'
 
 const locale = {
   name: 'de-at',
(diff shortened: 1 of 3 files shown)


iamkun:
Thanks. 

The thing you should update is 
```
weekStart: 1,
```

not the array of week name, I think.

flo-l:
Ah I see. It is already set to 1, so everything should be correct on your side.

I'm using https://github.com/y0c/react-datepicker, which does apparently ignore weekStart. I'm going to close this PR and try to fix the issue there.

Thank you for your fast response!

Issue #827, opened 2020-03-09
Error using new isoWeek plugin: "error TS2304: Cannot find name 'UnitType'"


Toub:
**Describe the bug**

After installing last version (1.8.22) to be able to use isoWeek plugin by @iamkun, I get a typescript compilation error:

```bash
/node_modules/dayjs/plugin/isoWeek.d.ts(6,20): error TS2304: Cannot find name 'UnitType'.
```
In the regarding file, `UnitType` is not imported:

```ts
import { PluginFunc, QUnitType, ConfigType } from 'dayjs'

declare const plugin: PluginFunc
export = plugin

type ISOUnitType = UnitType | 'isoWeek'; // TS2304: Cannot find name 'UnitType'.
```

**Expected behavior**

Typescript build should be OK.

**Information**
 - Day.js Version 1.8.22
- Typescript: 3.5.3
 - OS: linux/docker

**Related issues**

-  https://github.com/iamkun/dayjs/issues/724

iamkun:
will do thanks.

iamkun:
And it will be great if you could send us a pull request to fix this type definition error. 😬

Toub:
@iamkun are you going to release a patch to npm with this one soon?

iamkun:
in this week

PR #828, opened 2020-03-09
Fix isoWeek plugin typescript definition #827

Toub:
Fix isoWeek plugin typescript definition.

https://github.com/iamkun/dayjs/issues/827

fix: #827


Toub:
@iamkun here is the import fix

iamkun:
Thanks!

PR #836, opened 2020-03-15
Fix Arabic relative time

elbourki:
ar-tn, ar-sa, ar-ma, etc... are surprisingly following the correct formatting for relative time. This one should too.


iamkun:
THX

Issue #806, opened 2020-02-27
Does Calendar can use function in format as possible value


scorsi:
Hello,

It could be a good idea to add callbacks as possible values to Calendar format.
Like __momentjs__ does since `2.14.0` : [look here](https://momentjs.com/docs/#/displaying/calendar-time/).

Exemple:
With arrow function:
```javascript
dayjs("2020-02-26").calendar(null, {
  sameElse: (now, self) => {
    const d = self.diff(now);
    if (d > 0) return "After";
    else return "Before";
  }
});
```
With classic function:
```js
dayjs("2020-02-26").calendar(null, {
  sameElse: function (now) {
    const d = this.diff(now);
    if (d > 0) return "After";
    else return "Before";
  }
});
```

We can pass a function which take as parameter the date given to calendar call (here `null` so `now`) and to be fully compatible with __momentjs__ we can use `this` instead of the second parameter to access the base instance (here `dayjs("2020-02-26")`).

Thanks ;)

iamkun:
will do

KarthikSiddarth:
I am interested in working on this.

iamkun:
@KarthikSiddarth   that would be nicem thx

iamkun:
added now 😬

PR #839, opened 2020-03-17
Add option callback to calendar plugin

edumoreira1506:
This PR resolves issue #806 

Now the second param of `dayjs().calendar(...)` can be an object of callbacks, and the old behavior with a simples object of strings is working too!

I added some tests too for test this new feature. Hope that helps!

### DIFF
--- a/src/plugin/calendar/index.js
+++ b/src/plugin/calendar/index.js
@@ -10,8 +10,8 @@ export default (o, c, d) => {
     sameElse: L
   }
   const proto = c.prototype
-  proto.calendar = function (referenceTime, formats) {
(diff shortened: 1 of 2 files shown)


iamkun:
Thanks. 

Plus, I'd like to keep the callback the same as moment.js.  Could you please add more test to test it's `callback is a function with the scope of the current moment and first argument a moment that depicts now`

Cheers.

iamkun:
test failed
Screen Shot 2020-03-20 at 12 25 22 AM

edumoreira1506:
Fixed @iamkun !

iamkun:
THX

PR #846, opened 2020-03-20
Update km.js

sinh117801:
update Khmer translate


iamkun:
we now keep the same with https://github.com/moment/moment/blob/develop/locale/km.js

PR #850, opened 2020-03-24
Eslint update

sinh117801:
This PR includes:
- update eslint, eslint-plugin, and eslint-config-airbnb-base
- add new npm run script to auto fix lint rule.
- remove rule semi from .eslintrc.json
- fix the files that has lint error


iamkun:
Don't think we should change our es-lint rule

sinh117801:
I see. The eslint latest version not support node 6 either.

sinh117801:
this PR should be closed then.

PR #851, opened 2020-03-26
feat: add strict and custom thresholds support for relativeTime plugin

### LINKED ISSUE #830: How to get precise second(s) in `relativeTime` plugin?
**Describe the bug**
https://github.com/iamkun/dayjs/blob/dev/src/plugin/relativeTime/index.js#L8
All `seconds` displays `a few seconds`

**Expected behavior**
I want to display `1 second, 2 seconds...`

**Information**
 - Day.js Version: 1.8.22

JounQin:
close #830

### DIFF
--- a/src/plugin/relativeTime/index.js
+++ b/src/plugin/relativeTime/index.js
@@ -1,6 +1,36 @@
 import * as C from '../../constant'
 
+const LOOSE_THRESHOLDS = [
+  { l: 's', r: 44, d: C.S },
+  { l: 'm', r: 89 },
+  { l: 'mm', r: 44, d: C.MIN },
+  { l: 'h', r: 89 },
+  { l: 'hh', r: 21, d: C.H },
+  { l: 'd', r: 35 },
+  { l: 'dd', r: 25, d: C.D },
+  { l: 'M', r: 45 },
+  { l: 'MM', r: 10, d: C.M },
+  { l: 'y', r: 17 },
+  { l: 'yy', d: C.Y }
+]
+
+const STRICT_THRESHOLDS = [
+  { l: 's', r: 1 },
+  { l: 'ss', r: 59, d: C.S },
+  { l: 'm', r: 1 },
+  { l: 'mm', r: 59, d: C.MIN },
+  { l: 'h', r: 1 },
+  { l: 'hh', r: 23, d: C.H },
+  { l: 'd', r: 1 },
+  { l: 'dd', r: 29, d: C.D },
+  { l: 'M', r: 1 },
+  { l: 'MM', r: 11, d: C.M },
+  { l: 'y' },
+  { l: 'yy', d: C.Y }
+]
+
 export default (o, c, d) => {
+  o = o || {}
   const proto = c.prototype
   d.en.relativeTime = {
     future: 'in %s',
@@ -17,21 +47,23 @@ export default (o, c, d) => {
     y: 'a year',
     yy: '%d years'
   }
+  if (o.strict) {
+    Object.assign(d.en.relativeTime, {
+      s: '%d second', // 0 or 1
+      ss: '%d seconds',
+      m: '1 minute',
+      h: '1 hour',
+      d: '1 day',
+      M: '1 month',
+      y: '1 year'
+    })
+  }
   const fromTo = (input, withoutSuffix, instance, isFrom) => {
     const loc = instance.$locale().relativeTime
-    const T = [
-      { l: 's', r: 44, d: C.S },
-      { l: 'm', r: 89 },
-      { l: 'mm', r: 44, d: C.MIN },
-      { l: 'h', r: 89 },
-      { l: 'hh', r: 21, d: C.H },
-      { l: 'd', r: 35 },
-      { l: 'dd', r: 25, d: C.D },
-      { l: 'M', r: 45 },
-      { l: 'MM', r: 10, d: C.M },
-      { l: 'y', r: 17 },
-      { l: 'yy', d: C.Y }
-    ]
+    const T = o.thresholds || (o.strict ? STRICT_THRESHOLDS : LOOSE_THRESHOLDS)
+    if (!loc.ss) {
+      loc.ss = loc.s // locale like Chinese
+    }
     const Tl = T.length
     let result
     let out
@@ -44,10 +76,10 @@ export default (o, c, d) => {
           ? d(input).diff(instance, t.d, true)
           : instance.diff(input, t.d, true)
       }
-      const abs = Math.round(Math.abs(result))
+      const abs = Math[o.strict ? 'floor' : 'round'](Math.abs(result))
       isFuture = result > 0
       if (abs <= t.r || !t.r) {
-        if (abs === 1 && i > 0) t = T[i - 1] // 1 minutes -> a minute
+        if (abs <= 1 && i > 0) t = T[i - 1] // 1 minutes -> a minute, 0 seconds -> 0 second
         const format = loc[t.l]
         if (typeof format === 'string') {
           out = format.replace('%d', abs)
--- a/src/plugin/relativeTime/index.js
+++ b/src/plugin/relativeTime/index.js
@@ -29,35 +29,43 @@ const STRICT_THRESHOLDS = [
   { l: 'yy', d: C.Y }
 ]
 
+const LOOSE_EN_LOCALE = {
+  future: 'in %s',
+  past: '%s ago',
+  s: 'a few seconds',
+  m: 'a minute',
+  mm: '%d minutes',
+  h: 'an hour',
+  hh: '%d hours',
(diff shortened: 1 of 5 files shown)


src/plugin/relativeTime/index.js
```diff
@@ -1,6 +1,36 @@
...
+
+const STRICT_THRESHOLDS = [
+  { l: 's', r: 1 },
+  { l: 'ss', r: 59, d: C.S },
+  { l: 'm', r: 1 },
+  { l: 'mm', r: 59, d: C.MIN },
+  { l: 'h', r: 1 },
+  { l: 'hh', r: 23, d: C.H },
+  { l: 'd', r: 1 },
+  { l: 'dd', r: 30, d: C.D },
```
JounQin: Should it be `29` or `30`?

src/utils.js
```diff
@@ -45,6 +45,7 @@ const prettyUnit = (u) => {
 const isUndefined = s => s === undefined
 
 export default {
+  C, // for custom plugin
```
iamkun: what's this export for?

src/plugin/relativeTime/index.js
```diff
@@ -17,21 +47,23 @@ export default (o, c, d) => {
     y: 'a year',
     yy: '%d years'
   }
+  if (o.strict) {
+    Object.assign(d.en.relativeTime, {
```
iamkun: please avoid using es6 syntax `Object.assign`

src/utils.js
JounQin: Tests or user custom plugin.

https://github.com/iamkun/dayjs/pull/851/files#diff-f870cbe6ffaae01f79bb9242311d1f1aR133

If I'm going to write a custom plugin, I have to copy these constants manually before.

src/plugin/relativeTime/index.js
JounQin: But `Object.keys` is used in [updateLocale](https://github.com/iamkun/dayjs/blob/dev/src/plugin/updateLocale/index.js#L6).

iamkun: `Object.keys` is not introduced in es6

JounQin: Ah, sorry about that, how about this `extend` util?

src/plugin/relativeTime/index.js
```diff
@@ -17,21 +48,23 @@ export default (o, c, d) => {
     y: 'a year',
     yy: '%d years'
   }
+  if (o.strict) {
+    U.x(d.en.relativeTime, {
```
iamkun: can we just do 
```
d.en.relativeTime = {
  // and copy all the relevant keys
}
```

JounQin: I don't think it's a good pattern personally, but you'd like to do this, that's fine to me.

iamkun: Or, do we really have to update this? Why not just leave it to our user themself?

JounQin: Would it be much easier to use for users?

iamkun:
After carefully thinking about this PR, it seems it is enough for us to just add a `thresholds` option. And leave the rest to our end user. 

How do you think?

JounQin:
I think a `strict` preset would be helpful commonly, actually I don't understand why `moment` choose the loose way by default which is counterintuitive to me personally.

iamkun:
Here's my concern, If we added `strict` mode, should we update all other locales `relativeTime` part as well? Otherwise, this might be meanless for some locale.

iamkun:
However, if we just add a custom threshold, and leave the configuration to our user, it will be much better. All our user have to do is to use `UpdateLocale` to update `relativeTime` locale, to keep the sync with their custom thresholds.

JounQin:
> Otherwise, this might be meanless for some locale.

Yes, users have to update locales except built-in English.
But using `updateLocale` seems always required for other locales.
With `strict` option, we keep built-in locale simple.

Another idea would be supporting `strict` `relativeTime` option in locale file, but maybe it's not worth to do it.

```js
module.exports = {
  relativeTime: {
    s: 'a few seconds',
    _s: '%d second', // strict text
    ss: '%d seconds' // fallback to `ss` if `_ss` is not present in `strict` mode
  }
}
```

src/plugin/relativeTime/index.js
```diff
@@ -44,10 +84,10 @@ export default (o, c, d) => {
           ? d(input).diff(instance, t.d, true)
           : instance.diff(input, t.d, true)
       }
-      const abs = Math.round(Math.abs(result))
+      const abs = Math[o.strict ? 'floor' : 'round'](Math.abs(result))
```
JounQin: Another difference between two modes is here, `0.1s` would display `0 second` in `strict` mode.

iamkun: we can leave this to Additional token processing https://day.js.org/docs/en/customization/relative-time#additional-token-processing

iamkun:
As for now, how about adding the `thresholds` option in this coming release only. And waiting for issues reply to decide whether to introduce `strict` mode or not.

types/index.d.ts
```diff
@@ -96,9 +96,9 @@ declare namespace dayjs {
     locale(preset: string | ILocale, object?: Partial<ILocale>): Dayjs
   }
 
-  export type PluginFunc = (option: any, c: typeof Dayjs, d: typeof dayjs) => void
+  export type PluginFunc<T = unknown> = (option: T, c: typeof Dayjs, d: typeof dayjs) => void
 
-  export function extend(plugin: PluginFunc, option?: any): Dayjs
+  export function extend<T = unknown>(plugin: PluginFunc<T>, option?: T): Dayjs
```
iamkun: I don't know much about TypeScript. What's this change about, please?

JounQin: Stricter plugin options.

```ts
dayjs.extend(relativeTime, {
  Strict: true // error
})

dayjs.extend(relativeTime, {
  strict: true // pass
})
```

src/plugin/relativeTime/index.js
JounQin: Then we have to write a few methods for `ss`, `mm`, `hh`, `dd`, etc.?
That seems terrible to me.

JounQin:
> As for now, how about adding the `thresholds` option in this coming release only. And waiting for issues reply to decide whether to introduce `strict` mode or not.

I don't think it is feasible about "waiting for issues reply" personally.
There is only a few users care about it, maybe. 😂

iamkun:
https://momentjs.com/docs/#/customization/relative-time-threshold/
https://momentjs.com/docs/#/customization/relative-time-rounding/

moment has these two options.  We can do this too to suit your needs.
```js
{
threshold: {},
rounding: Math.floor
}
```

JounQin:
@iamkun Updated according to your last comment.

src/plugin/relativeTime/index.js
```diff
@@ -1,11 +1,13 @@
 import * as C from '../../constant'
 
 export default (o, c, d) => {
+  o = o || {}
   const proto = c.prototype
   d.en.relativeTime = {
     future: 'in %s',
     past: '%s ago',
     s: 'a few seconds',
+    ss: '%d seconds',
```
iamkun: seems we don't need this key.

src/utils.js
iamkun: We better not doing this. The constant object will increase a huge bundle size.

src/plugin/relativeTime/index.js
JounQin: `ss` is built-in supported by `moment`, and if we don't set it here, if user passes a custom `ss` threshold, its default format would be `undefined`.

iamkun: we can leave this to our user, I think. Same reason, we can't update the English locale only.

src/utils.js
JounQin: OK, I found `constant.js` is preserved in `esm` folder, but unavailable in root dir.

iamkun:
That's great! Thanks.

types/index.d.ts
```diff
@@ -96,9 +96,9 @@ declare namespace dayjs {
     locale(preset: string | ILocale, object?: Partial<ILocale>): Dayjs
   }
 
-  export type PluginFunc = (option: any, c: typeof Dayjs, d: typeof dayjs) => void
+  export type PluginFunc<T = unknown> = (option: T, c: typeof Dayjs, d: typeof dayjs) => void
```
iamkun: One last question, what's this <T = unknown> for?

JounQin: For compatibility and `dayjs.extend<T>()` as above, if a plugin ts definition doesn't pass a custom type for generic `T`, it fallbacks to `unknown`.

iamkun:
THX

PR #855, opened 2020-04-02
Add en-in and en-tt locales

ElliotAshby:
Addition of en-in and en-tt locales as extensions of the en-gb locale so that the correct date formats are used for these locales rather than using the default en formatting.


iamkun:
Seems not quite necessary?

Why not just parse the locale string before passing into `dayjs.locale`?

ElliotAshby:
Well, we support A LOT of locales (or at least try to), so it wouldn't really be reasonable to have to check each and every locale for what should be used in it's stead. We currently default to the base language's locale if the specific country's locale is not supported but I imagine for a lot of locales this is not correct. I figured it'd be a good thing for dayjs to support as many locales as possible out of the box, especially this way, as it is rather lightweight.
A lot of the currently supported english locales I noticed are almost exactly the same, perhaps a similar approach could be adopted with those to make them more lightweight?

iamkun:
Can you provide a list on Wikipedia or some alike that has these two locale strings, please?

Besides, please avoided using es6 syntax. Cause it will enlarge the final bundle size.

ElliotAshby:
Not entirely sure what you mean but I looked here at these links when confirming what was required.
http://www.localeplanet.com/icu/en-IN/index.html & http://www.localeplanet.com/icu/en-TT/index.html
So you'd rather have the locales defined in full to reduce bundle size?

iamkun:
> So you'd rather have the locales defined in full to reduce bundle size?

Yes, please. 

And any difference between en-IN and en-in, or which one we should choose?

ElliotAshby:
I've dropped the ES6 syntax, although I saw no difference in build size locally.

As for en-IN vs. en-in, the capitalised country code is the standard as far as I'm aware. But the locales in dayjs use lowercase country codes (except en-SG, not sure why). Perhaps it would be good to make things consistent?

iamkun:
Thanks, let's get this merged first. 😬

ElliotAshby:
thanks

PR #866, opened 2020-04-15
Consider using Intl api and update customParseFormat

iamkun:


janat08:
So does this simply improve performance building on my code, and if so can I get a share of bounty for helping.

iamkun:
@janat08 That would be nice.

This PR is only a test at the moment, still, I did not find a good way of implementing day time saving with timezone. e.g. dayjs('now', 'time-zone').add(1, 'month') , this might have a DTS rule change.

janat08:
ermon.jas@gmail.com is my PayPal.

I don't remember, but presumably you're aware of a function for detecting dst changes (should output a diff in tz with dst in play) in my code.

iamkun:
What I really mean is if we update DST in initialization only, that's not a quite right way to implement time zone. Cause this will cause errors after manipulation.

As a result, I may close the PR and look for another way to deal with DST after manipulation.

janat08:
I think I thought the utc offset and target locale variable should be used along with dst check once output of dayjs date is required. if you require comp perf the old method should be used. It should be outlined in my PR.

iamkun:
Besides, APIs like startOf, endOf will also cause a DST failure, if we only store a fixed offset like this PR. 

I guess, this might not the right way to implement tz.

janat08:
Looking at my code again, I belief the proto.tz method is for correcting date after day boundary pass, I assume it will integrate with any API so long as it runs the tz function.

janat08:
Did no other library like moment for example account for dst using browser API?

janat08:
As for the startof method there would have to be trickery involved, which would be the reason why I didn't jump to integrating tz with dayjs apis.

I think integration would look like running tz method as very first thing, and etc.

janat08:
I don't have a personal computer to help, nor a public one with quarantine, and so I would clearly appreciate a share of the bounty.

PR #869, opened 2020-04-15
upgrade dependencies

sinh117801:
- upgrade jest and fixed som error on test after upgrade. REF: [#2533](https://github.com/facebook/jest/pull/2533)
- upgrade size-limit reduce the size from 2.55KB to 2.11KB after build.


iamkun:
The new version would cause an `npm install error` on node 6.

Besides, I don't think `upgrade size-limit reduce the size from 2.55KB to 2.11KB after the build.` is really reduce the bundle size. Instead, it may just adjust its calculation algorithm?

sinh117801:
@iamkun, just a question. When will you plan to stop support node 6?

iamkun:
there's no need to stop support node 6, make be just a time to skip unit test on node 6

sinh117801:
Ok, I see. I will close this PR.

PR #871, opened 2020-04-15
fixed incorrect definition of locale in typescript

### LINKED ISSUE #870: dayjs.locale() defined incorrectly in typescript
**Describe the bug**
The [docs page for checking local](https://day.js.org/docs/en/i18n/getting-locale) says that you can call dayjs.locale() with no arguments to get the currently loaded locale. However, the type definition does not allow this. The current definition says that a parameter called preset must be present. This causes typescript to forbid calling locale with no arguments.  

**Expected behavior**
I expected to be able to call locale from typescript like this:
```
dayjs.locale();
```
but that is a syntax error

I would have expected the locale definition to be one of the following:
```
  export function locale(preset?: string | ILocale, object?: Partial<ILocale>, isLocal?: boolean): string
```

**Information**
 - Day.js Version: 1.8.24
 - OS: windows
 - Browser: compile time problem w/ TS
 - Time zone: EST

gregveres:
the first parameter of locale is optional so that locale() can be called with no arguments to return the current locale.
I would have added a test case to prove this works, but everything is in javascript and I expect you would not appreciate adding typescript to the unit tests.

This fixes #870

### DIFF
--- a/types/index.d.ts
+++ b/types/index.d.ts
@@ -100,7 +100,7 @@ declare namespace dayjs {
 
   export function extend<T = unknown>(plugin: PluginFunc<T>, option?: T): Dayjs
 
-  export function locale(preset: string | ILocale, object?: Partial<ILocale>, isLocal?: boolean): string
+  export function locale(preset?: string | ILocale, object?: Partial<ILocale>, isLocal?: boolean): string
 
   export function isDayjs(d: any): d is Dayjs
 


iamkun:
@gregveres Thanks.

iamkun:
It will be great if we could test d.ts in the unit test as well. Is it hard to implement this?

PR #883, opened 2020-04-26
Fix: Update RelativeTime plugin default config

zerooverture:
if require('dayjs/locale/en')
instance.$locale().relativeTime is undefined

Loc in fromTo is undefined
The loc[t.l] throws an error: Cannot read property 'xx' of undefined


iamkun:
THX

PR #886, opened 2020-04-27
feat:add/set/subtract support Object parameter

zerooverture:
add/set/subtract support Object parameter
and constructor support Array/Object parameter


iamkun:
Thanks.

1. It will be better to support these features as separate plugins.

2. Please add the related unit test.

zerooverture:
commit a standalone plugin[#887]

PR #887, opened 2020-04-27
feat:add plugin objectSupport

zerooverture:
add/set/subtract support Object parameter
and constructor support Object parameter

The related unit tests have been added.

### DIFF
--- a/src/plugin/objectSupport/index.js
+++ b/src/plugin/objectSupport/index.js
@@ -1,41 +1,27 @@
+import U from '../../utils'
+
 export default (o, c) => {
   const proto = c.prototype
-
-  /**
-   * Array to date
-   * @param d
-   * @param utc
-   * @returns {Date}
-   */
-  const parseArrayArgument = (d, utc) => {
-    if (utc) {
-      return new Date(Date.UTC(d[1], d[2] - 1, d[3]
-        || 1, d[4] || 0, d[5] || 0, d[6] || 0, d[7] || 0))
-    }
-    return new Date(d[1], d[2] - 1, d[3] || 1, d[4] || 0, d[5] || 0, d[6] || 0, d[7] || 0)
-  }
-
-  /**
-   * Converts the object to a date array
-   * @param d Object { years, months, date, hours, minutes, seconds, milliseconds}
-   * @param utc
-   */
-  const parseObjectArgument = (d, utc) => parseArrayArgument([
-    0,
-    d.y || d.year || d.years || 1970,
-    d.M || d.month || d.months || 1,
-    d.d || d.day || d.days || d.date || 1,
-    d.h || d.hour || d.hours || 0,
-    d.m || d.minute || d.minutes || 0,
-    d.s || d.second || d.seconds || 0,
-    d.ms || d.millisecond || d.milliseconds || 0
-  ], utc)
-
   const isObject = obj => !(obj instanceof Date) && !(obj instanceof Array) && obj instanceof Object
   const parseDate = (cfg) => {
     const { date, utc } = cfg
-    // if (Array.isArray(date)) return parseArrayArgument([0, ...date], utc)
-    if (isObject(date)) return parseObjectArgument(date, utc)
+    const $d = {}
+    if (isObject(date)) {
+      Object.keys(date).forEach((k) => {
+        $d[U.p(k)] = date[k]
+      })
+      const arr = [
+        $d.year || 1970,
+        $d.month - 1 || 0,
+        $d.day || 1,
+        $d.hour || 0,
+        $d.minute || 0,
+        $d.second || 0,
+        $d.millisecond || 0
+      ]
+      if (utc) return new Date(Date.UTC(...arr))
+      return new Date(...arr)
+    }
     return date
   }
 
--- a/src/plugin/objectSupport/index.js
+++ b/src/plugin/objectSupport/index.js
@@ -29,56 +29,35 @@ export default (o, c) => {
 
   const oldParse = proto.parse
   proto.parse = function (cfg) {
-    // console.log(cfg)
     cfg.date = parseDate.bind(this)(cfg)
     oldParse.bind(this)(cfg)
   }
 
-  const setObject = function (argument) {
-    const keys = Object.keys(argument)
-    let chain = this.clone()
-    keys.forEach((key) => {
-      chain = chain.$set(key, argument[key])
-    })
-    return chain
-  }
-  const addObject = function (argument) {
-    const keys = Object.keys(argument)
-    let chain = this
-    keys.forEach((key) => {
-      chain = chain.add(argument[key], key)
-    })
-    return chain
-  }
-
-  const subtractObject = function (argument) {
-    const keys = Object.keys(argument)
-    let chain = this
-    keys.forEach((key) => {
-      chain = chain.subtract(argument[key], key)
-    })
-    return chain
-  }
-
-  const oldSet = proto.set
-  proto.set = function (string, int) {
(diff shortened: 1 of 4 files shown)


src/plugin/objectSupport/index.js
```diff
@@ -0,0 +1,93 @@
...
+    return date
+  }
+
+  const oldParse = proto.parse
+  proto.parse = function (cfg) {
+    cfg.date = parseDate(cfg)
+    oldParse.bind(this)(cfg)
+  }
+
+  proto.setObject = function (argument) {
```
iamkun, line 47: seems no need to put it on the `proto`, just a simple function will be better.

zerooverture: you are right.

iamkun:
Help us keeping 100% test coverage https://github.com/iamkun/dayjs/blob/dev/CONTRIBUTING.md

zerooverture:
> Help us keeping 100% test coverage https://github.com/iamkun/dayjs/blob/dev/CONTRIBUTING.md
have already accomplished

types/plugin/objectSupport.d.ts
```diff
@@ -0,0 +1,12 @@
+import { PluginFunc } from 'dayjs'
+
+declare const plugin: PluginFunc
+export = plugin
+
+declare module 'dayjs' {
+    interface Dayjs {
+        set(argument: object): Dayjs
+        add(argument?: object): Dayjs
```
iamkun, line 9: why a `?` here?

zerooverture: sorry, I was careless.processed.

src/plugin/objectSupport/index.js
```diff
@@ -0,0 +1,96 @@
...
+  }
+
+  /**
+   * Converts the object to a date array
+   * @param d Object { years, months, date, hours, minutes, seconds, milliseconds}
+   * @param utc
+   */
+  const parseObjectArgument = (d, utc) => parseArrayArgument([
+    0,
+    d.y || d.year || d.years || 1970,
```
iamkun, line 25: check our `this.$utils().p()` prettyUnit to better handle this.

zerooverture: Their role is different, `d` is input by the user
`d => { years: 2010, months: 1, days: 14, hours: 15, minutes: 25, seconds: 50, milliseconds: 125 }`
or
`d => { y: 2010, M: 1, d: 14, h: 15, m: 25, s: 50, ms: 125 }`
I don't have to judge units, just need to get the value of some units, just like the REGEX_PARSE in constant.js.
This code is similar to the part in index.js where parseDate handles strings.
` const d = date.match(C.REGEX_PARSE)`  In the index.js 60 lines

iamkun: https://github.com/iamkun/dayjs/blob/dev/src/plugin/duration/index.js#L42

Something like this.

zerooverture: Is that right? But it doesn't support  date-> `dayjs({date:1})`

iamkun: 1 you can get `prettyUnit` from `this.$utils().p` instead of important it directlly.

2 you may have to extend a new `prettyUnit` function to treat `day and date key both mean day-of-the-month.` like https://github.com/iamkun/dayjs/blob/dev/src/plugin/duration/index.js#L26

zerooverture: Is that all right?

iamkun: cool

zerooverture: Thank you, seniors, let me learn a lot.
This is my first PR. Thanks♪(･ω･)ﾉ

iamkun: You are most welcome, cheers.

src/plugin/objectSupport/index.js
```diff
@@ -0,0 +1,84 @@
...
+        $d.year || 1970,
+        $d.month - 1 || 0,
+        $d.day || 1,
+        $d.hour || 0,
+        $d.minute || 0,
+        $d.second || 0,
+        $d.millisecond || 0
+      ]
+      if (utc) return new Date(Date.UTC(...arr))
+      return new Date(...arr)
```
iamkun, line 25: please avoid using this `...arr` advanced es6 syntax, this will hugely increase the bundle size

zerooverture: It seems much smaller now in packed

src/plugin/objectSupport/index.js
```diff
@@ -0,0 +1,84 @@
...
+        return new Date(Date.UTC(y, M, d, h, m, s, ms))
+      }
+      return new Date(y, M, d, h, m, s, ms)
+    }
+    return date
+  }
+
+  const oldParse = proto.parse
+  proto.parse = function (cfg) {
+    // console.log(cfg)
```
iamkun, line 32: no test code, pls.

src/plugin/objectSupport/index.js
```diff
@@ -0,0 +1,84 @@
...
+  const addObject = function (argument) {
+    const keys = Object.keys(argument)
+    let chain = this
+    keys.forEach((key) => {
+      chain = chain.add(argument[key], key)
+    })
+    return chain
+  }
+
+  const subtractObject = function (argument) {
```
iamkun, line 54: these three functions could be combined into one function with different arguments

src/plugin/objectSupport/index.js
```diff
@@ -0,0 +1,84 @@
...
+  }
+  const oldAdd = proto.add
+  proto.add = function (number, string) {
+    if (number instanceof Object) {
+      return addObject.bind(this)(number)
+    }
+    return oldAdd.bind(this)(number, string)
+  }
+  const oldSubtract = proto.subtract
+  proto.subtract = function (number, string) {
```
iamkun, line 78: `substract` could be implemented as negative `add`, subtract { day: 1 } => add { day: 1 * -1 }, so that we could only care `add` and `set` logic

zerooverture: the three methods have been merged into one.

src/plugin/objectSupport/index.js
```diff
@@ -0,0 +1,63 @@
...
+  }
+
+  const oldParse = proto.parse
+  proto.parse = function (cfg) {
+    cfg.date = parseDate.bind(this)(cfg)
+    oldParse.bind(this)(cfg)
+  }
+
+  const oldSet = function (int, string) {
+    return this.clone().$set(string, int)
```
iamkun, line 37: why a copy-paste instead of `proto.set`

zerooverture: here I've changed the `set ` to have the same parameter location as `add`, otherwise you have to judge in `callObject`,  which is inefficient and increases the package size.
or set and add can't share a `callObject`.
---------------------------
also, my attempt to use `proto.set` in `oldSet` will cause a dead loop because `proto.set` is redirected.
Using 'proto.$set' has the problem of the 'this' scope.

iamkun: Oh I see, why useing `return proto.set`  will cause a dead loop?

zerooverture:
```javasctip
 const oldSet = function (int, string) {
    return proto.set(string, int)
  }
...
proto.set = function (string, int) {
...
}
```
his code causes `oldSet` to enter the newly defined `proto.set` each time, creating A dead loop.

In the new code, I direct used a callback to solve `this` pointing problem.

Issue #792, opened 2020-02-04
Errors when updating. Webpack related


developdeez:
I haven't had this error until I updated. What loader do I use with dayjs now?

WARNING in ./node_modules/dayjs/locale/types.d.ts 1:8
Module parse failed: Unexpected token (1:8)
You may need an appropriate loader to handle this file type, currently no loaders are configured to process this file. See https://webpack.js.org/concepts#loaders
> declare interface ILocale {
|   name: string
|   weekdays?: string[]
 @ ./node_modules/dayjs/locale lazy ^\.\/.*$ namespace object ./types.d.ts      
 @ ./src/index.js

iamkun:
Any reproduction repo, please?

mariusa:
It happens when using 
```
require('dayjs/locale/' + lang)
```
That is, with a variable.

Example:
```
vue create hello-world
cd hello-world/
npm install dayjs --save
npm run serve
```

then change main.js to
```
import Vue from 'vue'
import App from './App.vue'

Vue.config.productionTip = false

import day from 'dayjs'
let lang = 'ro'
require('dayjs/locale/' + lang)
console.log(day().format())

new Vue({
  render: h => h(App),
}).$mount('#app')
```

munkhorgil:
Any update on this? I got this error too when dynamically importing locales.

iamkun:
@munkhorgil  Any reproduction repo, please?

munkhorgil:
> @munkhorgil Any reproduction repo, please?

https://github.com/erxes/erxes in widgets we use the dayjs

iamkun:
thanks, @munkhorgil, nice project.

However, it is hard for me to debug this issue in this huge repo. Can you please provide an issue reproduction repo containing this issue only, if you have time. 

Cheers

munkhorgil:
> thanks, @munkhorgil, nice project.
> 
> However, it is hard for me to debug this issue in this huge repo. Can you please provide an issue reproduction repo containing this issue only, if you have time.
> 
> Cheers

Thanks, in order to reproduce you don’t need to install the whole project.
1. Clone project
2. Navigate to widgets folder
3. cp .env.sample .env
4. yarn install
5. yarn dev
<img width="485" alt="Screen Shot 2020-05-29 at 4 28 52 PM" src="https://user-images.githubusercontent.com/978987/83238777-8334de80-a1c9-11ea-9af1-9d8f4506d01d.png">

JounQin:
[...]

PR #920, opened 2020-05-29
[Need Review] fix: Fix typescript type reference error

iamkun:
#792

Tring to fix typescript type error in #792

> /// \<reference path="./locale/index.d.ts" />

It may be caused by `webpack` treats `Triple-Slash Directives` as an imported resource.

I don't know much about typescript, and not sure if this is the correct way to fix it.

A code review is welcomed.

### DIFF
--- a/types/index.d.ts
+++ b/types/index.d.ts
@@ -1,4 +1,4 @@
-/// <reference path="./locale/index.d.ts" />
+import { ILocale } from './locale/types'
 
 export = dayjs;
 declare function dayjs (date?: dayjs.ConfigType, option?: dayjs.OptionType, locale?: string): dayjs.Dayjs
--- a/types/locale/index.d.ts
+++ b/types/locale/index.d.ts
@@ -1,4 +1,4 @@
-/// <reference path="./types.d.ts" />
+import { ILocale } from './types'
 
 declare module 'dayjs/locale/*' {
   namespace locale {
--- a/types/locale/types.d.ts
+++ b/types/locale/types.d.ts
@@ -1,4 +1,4 @@
-declare interface ILocale {
+export declare interface ILocale {
   name: string
   weekdays?: string[]
   months?: string[]


iamkun:
@JounQin Can you review this, please?

JounQin:
See https://github.com/iamkun/dayjs/issues/792#issuecomment-639961997

iamkun:
I see. Is this PR a proper fix to #921?

JounQin:
Not either. The issue author has clarified that there are two copies of `dayjs` which caused the problem. There's no issue as far as I can tell.

iamkun:
cool

iamkun:
@JounQin  One more question, in #924 I want to over-ride customParseFormat's type definition to support strict mode like `dayjs('1970-01-01', 'YYYY-MM-DD', true)`. However, I don't know how to fix it, can you please help if you have time? 
Cheers.

JounQin:
Sorry it seems not possible to implement it in a plugin definition AFAIK, function overloads must be defined at a time.

```ts
// index.d.ts
declare function dayjs (date?: dayjs.ConfigType, option?: dayjs.OptionType, locale?: string): dayjs.Dayjs

/**
 * @see {@link ./plugin/customParseFormat.d.ts}
 */
declare function dayjs (date?: dayjs.ConfigType, option?: dayjs.OptionType, strict?: boolean): dayjs.Dayjs
declare function dayjs (date?: dayjs.ConfigType, option?: dayjs.OptionType, locale?: string, strict?: boolean): dayjs.Dayjs
```

iamkun:
I see, thanks.  
That is to say, we have to put everything above in `index.d.ts`
Then, what's this line used for?
```
/**
 * @see {@link ./plugin/customParseFormat.d.ts}
 */
```

JounQin:
It a ts user calls the last two methods without enabling the related plugin, maybe the comment could help him.

iamkun:
I see, thanks

PR #943, opened 2020-06-27
feat: add arabic meridiem localization

ziadalzarka:
I added the localization for PM and AM in the Arabic locale


iamkun:
Thanks.

Please add the related unit test.

Besides, in https://github.com/moment/moment/blob/develop/locale/ar.js#L147, there's no difference between lower case or not? Should we follow this as well?

PR #955, opened 2020-07-09
Fix: Locale presets be typesafe

dyoshikawa:
Thanks for this great library.

TypeScript union type looks better than "string".
I hope for dayjs users understand expected `locale` args easily.


iamkun:
THX

PR #984, opened 2020-08-04
Update weekStart for pt-br

tiagozc:
In Brazil the first day of the week is Sunday, not Monday.


iamkun:
Thanks, we can simply remove this line then.

PR #985, opened 2020-08-05
Added typedefs in esm build

### LINKED ISSUE #971: Include type defs. in esm build
**Describe the bug**
Types are not included in the esm folder so they are not detected when using standard imports. 

```import dayjs from 'dayjs/esm';```

### PR DESCRIPTION
**Pull request by stoicsleuth**
Fixes #971

### DIFF
--- a/build/esm.js
+++ b/build/esm.js
@@ -1,6 +1,7 @@
 const fs = require('fs')
 const path = require('path')
 const util = require('util')
+const { ncp } = require('ncp')
 
 const { promisify } = util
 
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

**Comment by iamkun** [comment]
If we implement it in this way, there might be a type error while using plugins.

e.g.
```
import dayjs from 'dayjs/esm/index'
import utc from 'dayjs/esm/plugin/utc'

dayjs.extend(utc)

dayjs.utc // still type error because the TS compiler will search type file in './index.d.ts' not './esm/index.d.ts'
```

iamkun:
thanks

PR #996, opened 2020-08-08
feat: adds plugin for plural getters/setters

### LINKED ISSUE #692: Getter/setter plural aliases missing (e.g. hour() works but hours() does not)
https://github.com/iamkun/dayjs/issues/58#issuecomment-540910296

Spence-S:
fixes #692


iamkun:
Thanks. Please add the relevant type `d.ts` file.

And maybe we use this syntax `proto.map` to reduce our main bundle size as well?

PR #1009, opened 2020-08-17
[feature] Add parse duration from string

jamalroger:
Hello everyone,
This is a new feature can be adding to  moment duration
now we can create duration just from a string like  "1 day 57 minutes 34 seconds" will be  parse into normal object and create duration object
how work??
```js
console.log(dayjs.duration().fromString("1 day 57 minutes 34 seconds" )) // duration object
```

### DIFF
--- a/src/plugin/duration/index.js
+++ b/src/plugin/duration/index.js
@@ -1,4 +1,10 @@
-import { MILLISECONDS_A_WEEK, MILLISECONDS_A_DAY, MILLISECONDS_A_HOUR, MILLISECONDS_A_MINUTE, MILLISECONDS_A_SECOND } from '../../constant'
+import {
+  MILLISECONDS_A_WEEK,
+  MILLISECONDS_A_DAY,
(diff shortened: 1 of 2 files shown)


iamkun:
why not passing an object instead?

jamalroger:
I worked in project where the back-end return response  duration  like 1' day 57 minutes 34 seconds'
and  we want to parse it into  dayjs.duration to make some calculation 
For that reason I want this feature to be add  to dayjs

iamkun:
I think this might not be the best way.

```
dayjs.duration({
  seconds: 34,
  minutes: 57,
  days: 1,
});
```

This should be enough? All you need is a custom converter to do this.
```
1' day 57 minutes 34 seconds' ->  {
  seconds: 34,
  minutes: 57,
  days: 1,
}
```

jamalroger:
I already do that, I want to be function in dayjs but if not the best way I will close the pull request

PR #1032, opened 2020-09-04
fix: timezone plugin should have the same behavior in latest ICU v…

csvwolf:
Node 14 use new ICU rules: https://github.com/nodejs/node/issues/33027 which will cause the timezone plugin testing failed.

The behavior should be the same in different node version.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,6 +4,7 @@ cache:
     - ~/.npm
     - node_modules
 node_js:
+  - '14'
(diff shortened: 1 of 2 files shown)


iamkun:
THX

jeiea:
Should not it need a test for future regression?
[it](https://codesandbox.io/s/cocky-bash-t3qwm?file=/src/index.ts) might be a case...

Issue #1044, opened 2020-09-11
Localization issues


tarasbobak:
First of all thank you for an awesome library! 😃
This is were I stopped looking for an alternative to momentjs.

While testing dayjs with the locales we use in our project I've found several issues with a library localization.

1) For Poland (pl), Croatia (hr) and Lithuania (lt) names of month are not in the correct case of the word. Here is a difference between moment and dayjs localization. This are the results of dayjs().format('D MMMM, YYYY');
![image](https://user-images.githubusercontent.com/12515715/92928978-838b7980-f448-11ea-8c80-1cc65863a6ee.png)

2) For Bulgaria (bg), Spain (es), Croatia (hr), Poland (pl), Portugal (pt), Romania (ro), Serbia (sr) there are casing issues - month names are written capitalized while they should be lowercase. Example:
![image](https://user-images.githubusercontent.com/12515715/92929239-e54be380-f448-11ea-94f5-e88966ff2a32.png)

3) For wekday names dayjs().localeData().weekdays() list there are issues with casing for Bulgaria (bg), Criatia (hr), Poland (pl), Romania (ro) and Serbia (sr). Example:
![image](https://user-images.githubusercontent.com/12515715/92929951-0a8d2180-f44a-11ea-8994-8994c16d3200.png)

iamkun:
Hi, feel free to send us a pr if some locales are not correct.

tarasbobak:
Created PR with fixes.

iamkun:
> Created PR with fixes.

Cool thanks

iamkun:
I'll close this issue since it's been a while since it's been opened. Feel free to reopen if you have updates on this

PR #1045, opened 2020-09-11
Fix Polish, Hungarian and Lithuanian localization.

tarasbobak:
Fixing months names for Polish, Hungarian and Lithuanian locales:
- added formatted month names
- added function which determines which format to use

Fixing weekday names casing.

#1044


iamkun:
Thanks, please fix the CI error first.

iamkun (approved):

PR #1048, opened 2020-09-14
Fixes casing for en-SG locale name.

tarasbobak:
In order to simplify the way consumer can work with locales programmatically it is better to unify all the locale names to single format.

All other locales are lower case, only Singapore is en-SG


iamkun:
we also have to update the filename accordingly.

iamkun:
This is somehow a breaking change. Let's make it to 1.9.0

PR #1053, opened 2020-09-17
fix: fix DST bug in utc plugin

### LINKED ISSUE #961: Add Latin locale
For [romcal](https://github.com/romcal/romcal), I need to add Latin locale to `dayjs`. Therefore I have some questions regarding this.

---

1. As Latin weekday names are composed mostly of two words (like `feria secunda` for Monday), should I use a command as follows?

```js
weekdays: 'Dominica-feria secunda-feria tertia-feria quarta-feria quinta-feria sexta-Sabbato'.split('-'),
```

---

2. There’s a subissue: in Latin, there are [two different ways to name the weekdays](https://la.wikipedia.org/wiki/Hebdomas), and a third one (e.g. used in the Roman Catholic Church) which modifies one of the former two:

A. using `dies` (used in the Ancient Rome):
   - `dies Solis`;
   - `dies Lunae`;
   - `dies Martis`;
   - `dies Mercurii`;
   - `dies Iovis`;
   - `dies Veneris`;
   - `dies Saturni`;

B. using `ferias` (numbering days, starting with 1 for Sunday):
   - `feria prima`;
   - `feria secunda`;
   - `feria tertia`;
   - `feria quarta`;
   - `feria quinta`;
   - `feria sexta`;
   - `feria septima`;

C. variation of the `feria` names, used in the RC Church (uses `Dominica` for Sunday and `Sabbato` for Saturday):
   - `Dominica`;
   - `feria secunda`;
   - `feria tertia`;
   - `feria quarta`;
   - `feria quinta`;
   - `feria sexta`;
   - `Sabbato`.

Now, I need the C one only, but some other people _might_ like some other names. If I implement this, I’d like to be as thorough as possible, so my question is: should I create three separate regions (like `la-va` for Vatican, i.e. the 3rd type of weekday names)? Or how else should I fix this?

---

3. Weekday names: when the `feria` names are used, one may shorten it by using Roman numerals (e.g. `feria sexta` → `fera VI`). I presume I should use these in `weekdaysShort`, but I need a confirmation. Seldom, simply the Roman numerals are used (like `IV`), for which I could use `weekdaysMin`.

---

4. Another thing is that in Latin, they don’t use Arabic numeral, instead they _always_ use Roman numerals (for days of month, year, weekday numbers). I suggest to use Arabic numerals in time format (I don’t think that in Ancient Rome, they used time format at all and nowadays, I believe they use Arabic numerals for time instead). But how do I output Roman numerals? Does `dayjs` include such a converter? Should I create a custom function within `la.js`?

---

5. As I know that you compare the output of `dayjs` with the output of Moment in `dayjs` tests, is it required to create `la.js` locale in Moment first?

iamkun:


tukusejssirs:
This PR fixes the DST bug we discussed in #961, however it does not fix that issue (the issue deals with adding a Latin locale, which will be fixed by #966), therefore you should remove `fix #961` from the PR description.

As a sidenote: I have also rebased my `latin` branch (in a separate `l2` branch) on top of this PR just to try if it passes all tests in there too, and all tests are passed in there too.

iamkun:
Correct, I'll remove the description.

noppaklaow (approved):

noppaklaow (commented):
no

tukusejssirs:
@noppaklaow, may I ask what do you mean by `no`? :smiley:

PR #1057, opened 2020-09-18
fix: Add `setDefault` typing to timezone.d.ts

Qrymy:
Timezone plugin set default timezone at v1.8.36, however `timezone.d.ts` does not have type definition of `setDefault` method.

### DIFF
--- a/types/plugin/timezone.d.ts
+++ b/types/plugin/timezone.d.ts
@@ -11,7 +11,7 @@ declare module 'dayjs' {
   interface DayjsTimezone {
     (date: ConfigType, timezone: string): Dayjs
     guess(): string
-    setDefault(timezone?: string): Dayjs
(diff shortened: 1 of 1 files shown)


types/plugin/timezone.d.ts
```diff
@@ -11,6 +11,7 @@ declare module 'dayjs' {
   interface DayjsTimezone {
     (date: ConfigType, timezone: string): Dayjs
     guess(): string
+    setDefault(timezone?: string): Dayjs
```
iamkun, line 14: setDefault function returns nothing

Qrymy: I got it, I'm replacing `Dayjs` to `void`, Thanks!

iamkun:
Thanks

PR #1067, opened 2020-09-22
fix(locale): fixed typo for "monday" in arabic

AlaaZorkane:
It's `الإثنين` instead of `الإتنين` 

Sources: 
- https://en.wikipedia.org/wiki/Names_of_the_days_of_the_week
- https://shorturl.at/bnxQ2
- I am arab (Moroccan).

**UPDATE:** It seems like other arabic countries got this problem as well, converting this into draft until I fix it in all of them.


iamkun:
Hi, if you have finished this PR, you can remove the draft label so that it could get reviewed.

AlaaZorkane:
> Hi, if you have finished this PR, you can remove the draft label so that it could get reviewed.

Hey! Sorry for the delay, I had some work going on, I will work on it as soon as possible, which will be in the next 2 days or so ~ It's not really a big deal it's just a typo, but for middle-eastern clients might look a bit weird.

src/locale/ar-kw.js
```diff
@@ -3,7 +3,7 @@ import dayjs from 'dayjs'
 
 const locale = {
   name: 'ar-kw',
-  weekdays: 'الأحد_الإتنين_الثلاثاء_الأربعاء_الخميس_الجمعة_السبت'.split('_'),
+  weekdays: 'الأحد_الإثنين_الثلاثاء_الأربعاء_الخميس_الجمعة_السبت'.split('_'),
   months: 'يناير_فبراير_مارس_أبريل_ماي_يونيو_يوليوز_غشت_شتنبر_أكتوبر_نونبر_دجنبر'.split('_'),
   weekdaysShort: 'احد_اتنين_ثلاثاء_اربعاء_خميس_جمعة_سبت'.split('_'),
```
iamkun, line 8: I don't know Arabic, just a simple question, should we update this as well?  Or to keep the same with https://github.com/moment/moment/blob/develop/locale/ar-kw.js

AlaaZorkane: Yeah, `الإتنين` should be `الإثنين`, that's how it's spelled in Arabic, I honestly don't know where moment got these locales but if you check other Arabic countries `ar_*` they're all spelling it `الإثنين`, it's the same in all Arabic countries.

Here is is in wikipedia: https://cutt.ly/Wf9BSuk

iamkun: What I really ask is should we update `weekdaysShort ` in `ar-kw` as well?

AlaaZorkane: Yeah yes, I updated it ^^

iamkun: Cool

PR #1069, opened 2020-09-23
fix: update timezone plugin to support getting offset name e.g. EST

zazzaz:


test/plugin/timezone.test.js
```diff
@@ -252,3 +252,17 @@ describe('set Default', () => {
...
+
+describe('Get offsetName', () => {
+  const dtz = dayjs.tz('2012-03-11 01:59:59', NY)
+  it('short', () => {
+    const d = dtz.offsetName('short')
+    const m = moment.tz('2012-03-11 01:59:59', NY).format('z')
+    expect(d).toBe(m)
+    expect(d).toBe('EST')
+  })
+  it('short', () => {
```
akoskm, line 264: This should say 'long'.

zazzaz: right

PR #1089, opened 2020-09-28
fix: fix objectSupport plugin to get the correct result (zero-based month)

iamkun:
`dayjs({month: 1})` should return `2020-02-01` instead if `1970-01-01`

1. if no year provide, use the current year
2. month should be 0-based


iamkun:
This should be a breaking change, let make it from 1.8.x to 1.9.1

PR #1094, opened 2020-09-29
fix: export type of duration plugin

CatsMiaow:
All plugins have the following format, but only the duration plugin is `export default` in types
https://github.com/iamkun/dayjs/blob/f95ac15a4577ae5a3d1ce353872a2cd9fc454bc2/src/plugin/isMoment/index.js#L1
https://github.com/iamkun/dayjs/blob/f95ac15a4577ae5a3d1ce353872a2cd9fc454bc2/types/plugin/isMoment.d.ts#L4

When dayjs is actually built, it doesn't export the `default` object, so typescript runtime error occurs on `duration.default`.
https://cdn.jsdelivr.net/npm/dayjs@1.9.1/plugin/duration.js

This is a fix that keeps the same exports as other plugins.

### DIFF
--- a/types/plugin/duration.d.ts
+++ b/types/plugin/duration.d.ts
@@ -1,58 +1,61 @@
 import { PluginFunc } from 'dayjs'
 
 declare const plugin: PluginFunc
+export as namespace plugin;
 export = plugin
 
-type DurationInputType = string | number | object
-type DurationAddType = number | object | Duration
+declare namespace plugin {
+  type DurationInputType = string | number | object
+  type DurationAddType = number | object | Duration
 
-declare class Duration {
-  constructor (input: DurationInputType, unit?: string, locale?: string)
+  interface Duration {
+    new (input: DurationInputType, unit?: string, locale?: string): Duration
 
-  clone(): Duration
+    clone(): Duration
 
-  humanize(withSuffix?: boolean): string
+    humanize(withSuffix?: boolean): string
 
-  milliseconds(): number
-  asMilliseconds(): number
+    milliseconds(): number
+    asMilliseconds(): number
 
-  seconds(): number
-  asSeconds(): number
+    seconds(): number
+    asSeconds(): number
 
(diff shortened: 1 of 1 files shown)


iamkun:
Thanks. But this would lead to a new issue, how can we get the `Duration` type?

```
// currently we could import the type like this
import { Duration } from 'dayjs/pulgin/duration'
```

CatsMiaow:
PR fixed.
Referred to [@types/mathjs](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/mathjs/index.d.ts#L14-L16)

`esModuleInterop: false`
```ts
import * as duration from 'dayjs/plugin/duration';
const time: duration.Duration = dayjs.duration(1234, 's');
const [hour, minute, second] = [time.hours(), time.minutes(), time.seconds()];
```

`esModuleInterop: true`
```ts
import duration, { Duration } from 'dayjs/plugin/duration';
const time: Duration = dayjs.duration(1234, 's');
const [hour, minute, second] = [time.hours(), time.minutes(), time.seconds()];
```

iamkun:
Cool

types/plugin/duration.d.ts
```diff
@@ -1,58 +1,61 @@
...
 
-type DurationInputType = string | number | object
-type DurationAddType = number | object | Duration
+declare namespace plugin {
+  type DurationInputType = string | number | object
+  type DurationAddType = number | object | Duration
 
-export declare class Duration {
-  constructor (input: DurationInputType, unit?: string, locale?: string)
+  interface Duration {
```
iamkun, line 11: One small question, can we still use class Duration here, or is it necessary?

CatsMiaow: An error occurred when using a class in a namespace, so the class was implemented as an interface.
It probably conflicts with the `const plugin` object. https://stackoverflow.com/a/51716899

PR #1099, opened 2020-10-01
Allows to add durations to dayjs instance

### LINKED ISSUE #1093: dayjs().add(Duration) API
Hello, I'm replacing moment.js with `dayjs` in a project.

I'm not sure if the APIs are not compatible at this point or I'm making a mistake here:

```
// the old code basically had `moment` where you see dayjs now

const date1 = dayjs('2018-04-04T16:00:00.000Z')
const date2 = dayjs();
const duration = dayjs.duration(dayjs().diff(date1));
const date3 = date2.add(duration);
console.log(date3);
```

the output of the above code is:
```
    d {
      '$L': 'en',
      '$u': undefined,
      '$d': Invalid Date,
      '$x': {},
      '$y': NaN,
      '$M': NaN,
      '$D': NaN,
      '$W': NaN,
      '$H': NaN,
      '$m': NaN,
      '$s': NaN,
      '$ms': NaN
    }
```
Looking at the implementation of `add`:

https://github.com/iamkun/dayjs/blob/f95ac15a4577ae5a3d1ce353872a2cd9fc454bc2/src/index.js#L220-L221

it always expects the first parameter to be a number so that could one reason why this isn't working.

In moment.js we could add a `Duration` to a Moment object: https://momentjs.com/docs/#/manipulating/add/

One workaround would be to convert the duration to milliseconds and add it to the `date2`:

```
const date1 = dayjs('2018-04-04T16:00:00.000Z')
const date2 = dayjs();
const duration = dayjs.duration(dayjs().diff(date1)).asMilliseconds();;
const date3 = date2.add(duration, 'ms');
```

stebogit:
Fixes #1093

Now, once you've added the `Duration` plugin, you can 
```js
const d = dayjs('2020-10-01');
const days = dayjs.duration(2, 'days');
d.add(days).format('YYYY-MM-DD')); // 2020-10-03
```

### DIFF
--- a/src/plugin/duration/index.js
+++ b/src/plugin/duration/index.js
@@ -180,8 +180,9 @@ export default (option, Dayjs, dayjs) => {
   const oldAdd = Dayjs.prototype.add
   Dayjs.prototype.add = function (addition, units) {
     if (isDuration(addition)) {
-      return oldAdd.bind(this)(addition.asMilliseconds(), 'ms')
+      addition = addition.asMilliseconds();
+      units = 'ms';
     }
-    return oldAdd.bind(this)(addition, units)
+    return oldAdd.bind(this)(addition, units);
   }
 }
--- a/src/plugin/duration/index.js
+++ b/src/plugin/duration/index.js
@@ -180,9 +180,9 @@ export default (option, Dayjs, dayjs) => {
   const oldAdd = Dayjs.prototype.add
   Dayjs.prototype.add = function (addition, units) {
     if (isDuration(addition)) {
-      addition = addition.asMilliseconds();
-      units = 'ms';
+      addition = addition.asMilliseconds()
+      units = 'ms'
     }
-    return oldAdd.bind(this)(addition, units);
+    return oldAdd.bind(this)(addition, units)
   }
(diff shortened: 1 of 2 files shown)


stebogit:
Not sure why the two CI jobs have not been completed... 😞

iamkun:
@stebogit I don't know either, I'll restart it manually.

src/plugin/duration/index.js
```diff
@@ -176,4 +176,12 @@ export default (option, Dayjs, dayjs) => {
     return wrapper(input, { $l }, unit)
   }
   dayjs.isDuration = isDuration
+
+  const oldAdd = Dayjs.prototype.add
+  Dayjs.prototype.add = function (addition, units) {
+    if (isDuration(addition)) {
+      return oldAdd.bind(this)(addition.asMilliseconds(), 'ms')
```
iamkun, line 183: better DRY here
```
let a 
return oldAdd.bind(this)(a, 'ms')
```

stebogit:
here we go @iamkun

iamkun:
Cool

PR #1101, opened 2020-10-01
fix lowercase in LocaleData.longDateFormat()

xvaara:
LocaleData.longDateFormat() doesn't return anything for lowercase localized formats if they aren't defined in the locale files. this fixes that the same way that it's done in the localizedFormats plugin.

### DIFF
--- a/src/plugin/localeData/index.js
+++ b/src/plugin/localeData/index.js
@@ -13,6 +13,10 @@ export default (o, c, dayjs) => { // locale needed later
     return result.map((_, index) => (result[(index + (weekStart || 0)) % 7]))
   }
   const getDayjsLocaleObject = () => dayjs.Ls[dayjs.locale()]
+  const getLongDateFormat = format =>
+    getDayjsLocaleObject().formats[format] ||
+    t(getDayjsLocaleObject().formats[format.toUpperCase()])
+
   const localeData = function () {
     return {
       months: instance =>
@@ -25,8 +29,8 @@ export default (o, c, dayjs) => { // locale needed later
         (instance ? instance.format('dd') : getShort(this, 'weekdaysMin', 'weekdays', 2)),
       weekdaysShort: instance =>
         (instance ? instance.format('ddd') : getShort(this, 'weekdaysShort', 'weekdays', 3)),
-      longDateFormat: format =>
-        this.$locale().formats[format] || t(this.$locale().formats[format.toUpperCase()])
+      longDateFormat: format => getLongDateFormat(format)
+
     }
   }
   proto.localeData = function () {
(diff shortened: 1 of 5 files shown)


src/plugin/localeData/index.js
```diff
@@ -12,6 +12,7 @@ export default (o, c, dayjs) => { // locale needed later
   }
   const getDayjsLocaleObject = () => dayjs.Ls[dayjs.locale()]
   const localeData = function () {
+    const t = format => format.replace(/(\[[^\]]+])|(MMMM|MM|DD|dddd)/g, (_, a, b) => a || b.slice(1))
```
iamkun, line 15: can we export this method from localizedFormat plugin for better code management?

xvaara: utils.js? what should I call it? makeLocalSmall again? :D

iamkun: No, what I think is simply export this function from  localizedFormat.js. This should not affect the plugin to work but also simply the logic here.

iamkun:
Should we take this https://github.com/iamkun/dayjs/pull/1106 into consideration as well?

xvaara:
I added the same to global objecct. I think I merged my branch on the wrong commit, sorry about that...

src/plugin/localeData/index.js
```diff
@@ -23,7 +25,8 @@ export default (o, c, dayjs) => { // locale needed later
         (instance ? instance.format('dd') : getShort(this, 'weekdaysMin', 'weekdays', 2)),
       weekdaysShort: instance =>
         (instance ? instance.format('ddd') : getShort(this, 'weekdaysShort', 'weekdays', 3)),
-      longDateFormat: format => this.$locale().formats[format]
+      longDateFormat: format =>
+        this.$locale().formats[format] || t(this.$locale().formats[format.toUpperCase()])
```
iamkun, line 29: can we make this a separate function like
```
const getLongDateFormat = (locale) => locale.formats[format] || t(locale.formats[format.toUpperCase()])
```

xvaara: I did
```javascript
const getLongDateFormat = format =>
  getDayjsLocaleObject().formats[format] ||
  t(getDayjsLocaleObject().formats[format.toUpperCase()])
```

src/plugin/localeData/index.js
```diff
@@ -23,7 +29,8 @@ export default (o, c, dayjs) => { // locale needed later
         (instance ? instance.format('dd') : getShort(this, 'weekdaysMin', 'weekdays', 2)),
       weekdaysShort: instance =>
         (instance ? instance.format('ddd') : getShort(this, 'weekdaysShort', 'weekdays', 3)),
-      longDateFormat: format => this.$locale().formats[format]
+      longDateFormat: format => getLongDateFormat(format)
```
iamkun, line 32: we should pass instance locale, right?  this.$locale()

xvaara: I added those to the function.

PR #1108, opened 2020-10-02
fix: Serbian locale grammar (sr, sr-cyrl)

dusansimic:
This fixes the Serbian locale grammar since current logic doesn't account for different cases in different tenses (past and future) and no tenses (without suffix) alongside long date formats that have already been fixed in Moment.js.

The same logic has been sent as a PR for Moment.js https://github.com/moment/moment/pull/5742 where fixes are described a little bit more.


iamkun:
Thanks, and waiting for native speaker review ( 0 / 2 )

iamkun:
LGTM

PR #1109, opened 2020-10-03
fix: de-locale for relativeTime, and added unit-tests

froks:
I've fixed the relativeTime output for the 'de'-locale when withoutSuffix is set. Also added a unit-test that checks the results in absolute and against moment.


iamkun:
Thanks

PR #1110, opened 2020-10-04
add localizedFormats to customParseFormat

xvaara:
It's not working on all the localizedFormats ( LLL, LLLL and llll ) but that would be something in the main parser since I think parsing a and A fails in non "en" locales.


xvaara:
@iamkun should I create an issue about the bug in A and a formats in localizations?

src/plugin/customParseFormat/index.js
```diff
@@ -118,6 +120,9 @@ function correctHours(time) {
 }
 
 function makeParser(format) {
+  if (format.match(/l/i)) {
+    format = format.replace(/(\[[^\]]+])|(LTS|LT|l+|L+)/g, (_, a, b) => a || locale.formats[b] || t(locale.formats[b.toUpperCase()]))
```
iamkun, line 124: just thinking should we export this logic from localizedFormat as well?

xvaara: it's now exported from there.

xvaara:
should the u function be in utils.js also? Like #1132 cased error? Then we'd need to move englishFormats there also?

iamkun:
Thanks, not sure if we need this feature or not. Since `new Date()` already has the ability to parse such format ?

xvaara:
> Thanks, not sure if we need this feature or not. Since `new Date()` already has the ability to parse such format ?

I don't follow. but my knowledge of localized parsing using pure javascript Date functionality is non existent. I'm just converting my app from moment to dayjs. I'm using this to validate (and create dayjs instance) user input in localised format "l LT", but if this isn't in the scope of the project I can continue using my functionality to extract the localised formats to the inputed format string... But in those I'm importing stuff from dayjs internals, so this would make it easier and future proof.

iamkun:
Agreed. Let's see if someone needs this feature as well.

xvaara:
What I'm reading from docs, new Date() doesn't parse any locales... or am I missing something? I noticed that this didn't work when converting from moment, and hacked something to get by. But since it's a hack that depends on importing locales to my code and inlining en locales (since they aren't exported anywhere) I looked at how to implement it directly in dayjs. Since dayjs is supposed to be same api as moment, shouldn't this also be implemented? For this pull request I looked how moment does this, and It's pretty much the same.

iamkun:
What do you mean by `new Date() doesn't parse any locales`? Day.js supports parsing in a specific locale  like `dayjs('2018 Enero 15', 'YYYY MMMM DD', 'es')`

xvaara:
> What do you mean by `new Date() doesn't parse any locales`? 

Well you said that new Date() parses such formats 6 days ago... but I can't seem to find anything about parsing different formats, only hints to use moment.js. Since localizedFormats is moment.js invention, I really think no other parses them (except luxon). 

> Day.js supports parsing in a specific locale like `dayjs('2018 Enero 15', 'YYYY MMMM DD', 'es')`

Yes, but doesn't parse the localizedFormats, that's the point of this pull request.

iamkun:
Oh, I see, what I mean above, 'LLLL' string like `Thursday, September 4, 1986 8:30 PM` is a valid input to `new Date()`, that is to say, even if we don't implement this, `dayjs('Thursday, September 4, 1986 8:30 PM')` still outputs an expected result.

xvaara:
that leaves every other variation of formats and locales not functioning...

iamkun:
Will this is just an example, other formats in localizedFormats should all have be supported.

xvaara:
they are not. As the test case I included. It fails without this patch.

Here the same test with one locale on codesandbox: https://codesandbox.io/s/lingering-dream-tm4tk?file=/src/index.js

iamkun:
Well, my bad. then this pr makes sense.

PR #1116, opened 2020-10-06
Export function declarations from LocaleData Plugin

Mocha--:


types/plugin/localeData.d.ts
```diff
@@ -6,5 +6,14 @@ export = plugin
 declare module 'dayjs' {
   interface Dayjs {
     localeData(): any
```
iamkun, line 8: should we add it here?

Mocha--: You mean under interface Dayjs?

If so, I think the exported functions should not sit under the interface since those functions are not instance methods of Dayjs. 

Instead, those functions are like static, not binding to a class.

iamkun: How can typescript knows that weekdaysMin is available on `dayjs` like this dayjs.weekdaysMin()

Mocha--: because those functions are exported from module 'dayjs', which has nothing to do with interface Dayjs.

Screen Shot 2020-10-08 at 10 38 42 am

just like MinMax plugin

iamkun: True, should we take dayjs().localeData().weekdays() into consideration as well?

Mocha--: right. I've added localeData types to both dayjs and Dayjs instances.

Now (dayjs or dayjs()).localeData() both have types.

iamkun:
Looks cool.

Two minor suggestions:

1. we can pass an instance to  `dayjs().localDate().months(instance)`, the same as  monthsShort, firstDayOfWeek, weekdays, weekdaysMin, weekdaysShort
2. we can pass a format to `dayjs().localDate().longDateFormat(format)` and `dayjs.localDate().longDateFormat(format)`

Ref:  https://github.com/iamkun/dayjs/blob/dev/src/plugin/localeData/index.js#L21

Issue #1118, opened 2020-10-08
UpdateLocale Plugin does not allow partial updates to locale


stefanbinder:
**Describe the enhancement**
While using the UpdateLocale Plugin, it isn't possible to just update one single value in a nested value.
For example I just want to update the `format.L` value:

```
dayjs.extend(UpdateLocale);
dayjs.updateLocale('de', {
  formats: {
    L: 'D.M.YYYY',
  },
});
```

Like this other formats `LL, LLL, ...` are getting lost.

**Expected behavior**
I want to pass in an object, which updates the locale information partially.

**Information**
 - Day.js Version [e.g. v1.9.1]

PS:  
If the issue is accepted as a bug/enhancement, I would like to contribute and create a PR for that issue 😃

iamkun:
Will update this feature later.

ajisulist:
Trying to fix here https://github.com/iamkun/dayjs/pull/2075 @iamkun

PR #1122, opened 2020-10-11
#1118 - Partial Locale update added

VigneshMurugan:
In case of Object we would be using Object.assign to enable partial update of the locale property

### DIFF
--- a/src/plugin/updateLocale/index.js
+++ b/src/plugin/updateLocale/index.js
@@ -5,7 +5,7 @@ export default (option, Dayjs, dayjs) => {
     if (!localeConfig) return
     const customConfigKeys = customConfig ? Object.keys(customConfig) : []
     customConfigKeys.forEach((c) => {
-      localeConfig[c] = customConfig[c]
+      localeConfig[c] = typeof localeConfig[c] === 'object' ? Object.assign({} , localeConfig[c], customConfig[c]): customConfig[c]
     })
     return localeConfig // eslint-disable-line consistent-return
   }


iamkun:
Thanks. It will be better not using `Object.assign`, cause babel will bundle a huge polyfill to make it work on old browsers.

xvaara:
does spread add something in babel? eg. 
```
{...localeConfig[c], ...customConfig[c]}
```

or do we need a helper function like:
```javascript
function m(o1,o2){
    var o = {};
    for (var attrname in o1) { o[attrname] = o1[attrname]; }
    for (var attrname in o2) { o[attrname] = o2[attrname]; }
    return o;
}
```

or do we need those as recursive functions?

iamkun:
It's not about the syntax itself, but the whole bundle size while compiling it to es5.

VigneshMurugan:
@iamkun looking for review. I have replaced Object.assign with custom extend method

iamkun:
Thanks, would you please move the utils function to the plugin file?

VigneshMurugan:
> Thanks, would you please move the utils function to the plugin file?

Done

iamkun:
sorry that I didn't make it clear, what I mean is to move the extend function to `src/plugin/updateLocale/index.js` or maybe `src/plugin/updateLocale/utils.js`

VigneshMurugan:
> sorry that I didn't make it clear, what I mean is to move the extend function to `src/plugin/updateLocale/index.js` or maybe `src/plugin/updateLocale/utils.js`

Done @iamkun

iamkun:
LGTM, would you like to add some unit tests for this feature?

VigneshMurugan:
@iamkun Not sure why this got away in my checklist. I have updated the unit test cases now.

PR #1126, opened 2020-10-12
Update da.js

bjerh:
Danish date formats are using hyphens rather than dots...


bjerh:
You can do either, it seems. Sorry for the inconvenience.

PR #1156, opened 2020-10-20
Fix issues with adding/subtracting durations from Dayjs objects

mfogel:
In #1099 support was added for passing in a duration object to `add()` or `subtract()`, which is awesome. I ran across a few issues when using it. This PR:

- adds support for `subtract(duration)` (changelog thinks this was part of #1099, but it wasn't)
- adds typescript type signatures for `add(duration)` and `subtract(duration)`

Also missing are docs for this feature, though that looks like that should be done in the [website repo](https://github.com/dayjs/dayjs-website) so not included in this PR.

Thanks a ton for dayjs!

### DIFF
--- a/src/plugin/duration/index.js
+++ b/src/plugin/duration/index.js
@@ -180,17 +180,12 @@ export default (option, Dayjs, dayjs) => {
 
   const oldAdd = Dayjs.prototype.add
   const oldSubtract = Dayjs.prototype.subtract
-  const transformArgs = (value, units) => {
-    if (isDuration(value)) {
-      value = value.asMilliseconds()
-      units = 'ms'
-    }
-    return [value, units]
+  Dayjs.prototype.add = function (value, unit) {
+    if (isDuration(value)) value = value.asMilliseconds()
+    return oldAdd.bind(this)(value, unit)
   }
-  Dayjs.prototype.add = function (value, units) {
-    return oldAdd.apply(this, transformArgs(value, units))
-  }
-  Dayjs.prototype.subtract = function (value, units) {
-    return oldSubtract.apply(this, transformArgs(value, units))
+  Dayjs.prototype.subtract = function (value, unit) {
+    if (isDuration(value)) value = value.asMilliseconds()
+    return oldSubtract.bind(this)(value, unit)
   }
 }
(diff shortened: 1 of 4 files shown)


src/plugin/duration/index.js
```diff
@@ -179,11 +179,18 @@ export default (option, Dayjs, dayjs) => {
...
 
   const oldAdd = Dayjs.prototype.add
-  Dayjs.prototype.add = function (addition, units) {
-    if (isDuration(addition)) {
-      addition = addition.asMilliseconds()
+  const oldSubtract = Dayjs.prototype.subtract
+  const transformArgs = (value, units) => {
+    if (isDuration(value)) {
+      value = value.asMilliseconds()
       units = 'ms'
```
iamkun, line 186: Cool, seems we don't need this line, cause 'ms' is the default unit.

mfogel:
@iamkun thanks for catching that

I also pushed [another commit](https://github.com/iamkun/dayjs/pull/1156/commits/379f3484fca94057f2f39d26e503aa83b031fbc2) to this PR: the typescript typings for `add()` and `subtract()` show the `unit` arg as required, which as you pointed out, it isn't.

That commit introduced some merge conflicts with work that's been merged to `dev` recently, so I also rebased on the latest on `dev`.

Let me know if you see anything else to fix/change or whatnot, thank you.

types/index.d.ts
```diff
@@ -222,21 +222,21 @@ declare namespace dayjs {
...
      */
-    add(value: number, unit: OpUnitType): Dayjs
+    add(value: number, unit?: OpUnitType): Dayjs
     /**
      * Returns a cloned Day.js object with a specified amount of time subtracted.
      * ```
      * dayjs().subtract(7, 'year')// => Dayjs
      * ```
-     * Units are case insensitive, and support plural and short forms.
+     * Units are case insensitive, support plural and short forms, and default to 'ms' if not supplied.
```
iamkun, line 235: IMO it's better not to add this note, this will make our user confusing, and the code is hard to read.

mfogel: sounds good. I pushed a commit just now to revert those changes to those doc strings.

PR #1166, opened 2020-10-23
Update index.js: Changed assignment of "this.$L" in "Dayjs" constructor

florianmarkusse:
Hello,

In the constructor of the Dayjs class, I changed the line:
`this.$L = this.$L || parseLocale(cfg.locale, null, true)` to
`    this.$L = parseLocale(cfg.locale, null, true)`
because `this.$L` is undefined at the time of assignment and the right operand of the OR-operation will then always set `this.$L`.

Hope it can save you some bytes!

Kind regards,
Florian

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -76,7 +76,7 @@ const parseDate = (cfg) => {
 
 class Dayjs {
   constructor(cfg) {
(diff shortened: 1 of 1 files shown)


iamkun:
True, thanks.

PR #1192, opened 2020-11-08
Add relative times to lv (Latvian) locale

edgarsn:
Latvian language is missing `relativeTimes` translation object so here you go :)


iamkun:
THX

PR #1214, opened 2020-11-13
feat: avoid install installed plugin

ibufu:


iamkun:
Any use case, please?

ibufu:
For example, UTC plugin will set a new format which call the old format
function. If we repeatedly call extend(UTC), it will repeatedly call
format. It is very dangerous for SSR because leaking of memory.

iamkun <notifications@github.com>于2020年11月14日 周六上午12:15写道：

ibufu:
Like use function in Vue.js, it also detects duplicate of plugins.

cx L <chenxuanincn@gmail.com>于2020年11月14日 周六上午12:27写道：

iamkun:
What if we want to install two different plugins at the same time?

ibufu:
It's ok. It only prevents same plugin.

derekhu:
This MR is greate to fix the problem in SSR.

Because of re-install for plugin, the function overwriten ref the the old one, so in SSR , the function call will increase each time the module loaded.  Cause to stack over-flow finally.

PR #1229, opened 2020-11-16
Fix startOf/endOf behavior with timezone

### LINKED ISSUE #1212: Getting wrong startOf('day') in a different timezone
**Describe the bug**
On calculating startOf('day') in a different timezone, I am getting the wrong result. Note that I tried this at 6 PM on November 12 in Asia/Kolkata timezone.
Code - `dayjs.tz(new Date()).tz('America/New_York').startOf('day').toString()`
Result - `"Thu, 12 Nov 2020 15:30:00 GMT"`

**Expected behavior**
It should be `Thu, 12 Nov 2020 05:00:00 GMT`

**Information**
 - Day.js Version - Tried in browser console on day.js.org website
 - OS - MacOS Mojave
 - Browser - Chrome 86.0.4240.193 
 - Time zone: Asia/Kolkata

nbudin:
This fixes #1212.  Because the timezone plugin stores a `$d` value that's relative to the timezone of the dayjs instance, `startOf` and `endOf` need to ignore the absolute `$d` value and instead use the logical value in that time zone.

This overrides `startOf` to ignore the timezone of the instance, calculate the `startOf` value, then reapply the timezone while keeping local time.  This should also fix the behavior for `endOf` since it is actually calling `startOf` internally.

### DIFF
--- a/src/plugin/timezone/index.js
+++ b/src/plugin/timezone/index.js
@@ -114,6 +114,17 @@ export default (o, c, d) => {
     return result && result.value
   }
 
+  const oldStartOf = proto.startOf
+  proto.startOf = function (units, startOf) {
+    if (!this.$x || !this.$x.$timezone) {
+      return oldStartOf.call(this, units, startOf)
+    }
+
+    const withoutTz = d(this.format('YYYY-MM-DD HH:mm:ss:SSS'))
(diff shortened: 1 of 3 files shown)


nikhilag:
Thanks @nbudin for fixing this.
Is it possible to ping a maintainer to look into this PR?

iamkun:
nice catch

PR #1230, opened 2020-11-16
fix: better nil check on utils

pedronauck:
When you try to pass an empty string for the constructor an obtrusive `Invalid date` error is showing.
Something like that doesn't work: `dayjs('').format(YYYY-MM-DD')` and breaks the system.

What I did is just check if a string is an empty string or is undefined or null.

### DIFF
--- a/src/utils.js
+++ b/src/utils.js
@@ -43,7 +43,9 @@ const prettyUnit = (u) => {
   return special[u] || String(u || '').toLowerCase().replace(/s$/, '')
 }
 
-const isUndefined = s => s === undefined
+const isNil = val => {
+  return (typeof === 'string' && !Boolean(val.length) || val === undefined || val === null;
+}
 
 export default {
   s: padStart,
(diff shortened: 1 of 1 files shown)


iamkun:
Thanks. But I don't think `''` is a valid date in js.

pedronauck:
I know that `''` isn't a valid date, but sometimes you can't control all data in some systems, this is a common case (mainly in legacy systems), and update to dayjs can be something hard since if you have some like `''` inside the constructor it will break the entire system. Handle properly with that could be something to prevent system break-outs and turn dayjs able for a lot of developers.

I talking to you about that because of some system of one of my customers is a little legacy and this happened when I try to change of moment to dayjs (was thousand of files).

iamkun:
Alright, I do think adding a custom plugin to parse `''` is a better choice. Cause if we accept this PR, it will also affect a lot of projects that depend on dayjs that treat `''` as invalid

PR #1232, opened 2020-11-18
fa-AF

NotNotMarshall:
fa-AF is Afghani dialect of the Persian language (also known as Farsi Afghani).
In Afghanistan the Gregorian month names are a bit different than how they sound in Iran.

### DIFF
--- a/src/locale/fa-AF
+++ b/src/locale/fa-AF
@@ -1,0 +1,40 @@
+// Persian [fa]
+import dayjs from 'dayjs'
+
+const locale = {
+  name: 'fa',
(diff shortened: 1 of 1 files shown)


iamkun:
Thanks

Some small error:

1 file name in lower cases.

2 file name add .js extension

3 add come unit test

NotNotMarshall:
> Thanks
> 
> Some small error:
> 
> 1 file name in lower cases.
> 
> 2 file name add .js extension
> 
> 3 add come unit test

Is it stupid to say I have never did Unit Testing in my life and always tried to avoid it?😆

iamkun:
AHA, maybe it's a good time to have a try. 

It's not that hard, you can check here for some reference,  https://github.com/iamkun/dayjs/tree/dev/test/locale

PR #1233, opened 2020-11-18
Added Pashto language support

NotNotMarshall:
Pashto is one of the spoken languages in Afghanistan and Pakistan which has over 60 million speaker worldwide.
ps is the international ISO code for this locale.

### DIFF
--- a/src/locale/ps.js
+++ b/src/locale/ps.js
@@ -1,0 +1,40 @@
+// Pashto [ps]
+import dayjs from 'dayjs'
+
+const locale = {
+  name: 'pashto',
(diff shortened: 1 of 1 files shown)


src/locale/ps.js
```diff
@@ -0,0 +1,40 @@
+// Pashto [ps]
+import dayjs from 'dayjs'
+
+const locale = {
+  name: 'pashto',
```
iamkun, line 5: should be 'ps' here

NotNotMarshall: Yep, my mistake. Thanks!

PR #1255, opened 2020-12-06
Preparse postformat + implementation in ar locale + tests

VehpuS:
Added a plugin to implement the preparse / postformat functions for locales such as "ar", based on the moment implementation. I added tests for both the plugin and the locale changes based on the moment tests.

This should solve #416 

Do note that the relative time implementation for the ar locale does not match the moment one, so the tests for the locale are different (to focus on coverage and not correctness tests - since I am not an arabic speaker and could not fix the relative time implementation in the scope of this change).

### DIFF
--- a/src/plugin/preParsePostFormat/index.js
+++ b/src/plugin/preParsePostFormat/index.js
@@ -1,0 +1,53 @@
+import localeData from '../localeData'
+
+// Plugin template from https://day.js.org/docs/en/plugin/plugin
+export default (option, dayjsClass, dayjsFactory) => {
+  // This plugin depends on other plugins - so I will import them here
+  // equivalent to dayjsClass.extend(duration)
+  localeData(option, dayjsClass, dayjsFactory)
+
+  const oldParse = dayjsClass.prototype.parse
+  dayjsClass.prototype.parse = function (cfg) {
+    if (typeof cfg.date === 'string') {
+      const locale = this.$locale()
+      cfg.date =
+        locale && locale.preparse ? locale.preparse(cfg.date) : cfg.date
+    }
+    // original parse result
+    return oldParse.bind(this)(cfg)
+  }
+
+  // // overriding existing API
+  // // e.g. extend dayjs().format()
+  const oldFormat = dayjsClass.prototype.format
+  dayjsClass.prototype.format = function (...args) {
+    // original format result
+    const result = oldFormat.call(this, ...args)
+    // return modified result
+    const locale = this.$locale()
+    return locale.postformat ? locale.postformat(result) : result
+  }
+
+  const oldFromTo = dayjsClass.prototype.fromToBase
+
+  if (oldFromTo) {
+    dayjsClass.prototype.fromToBase = function (
+      input,
+      withoutSuffix,
+      instance,
+      isFrom
+    ) {
+      const locale = this.$locale() || instance.$locale()
+
+      // original format result
+      return oldFromTo.call(
+        this,
+        input,
+        withoutSuffix,
+        instance,
+        isFrom,
+        locale.postformat
+      )
+    }
+  }
+}
--- a/src/plugin/preParsePostFormat/index.js
+++ b/src/plugin/preParsePostFormat/index.js
@@ -1,11 +1,5 @@
-import localeData from '../localeData'
-
 // Plugin template from https://day.js.org/docs/en/plugin/plugin
-export default (option, dayjsClass, dayjsFactory) => {
-  // This plugin depends on other plugins - so I will import them here
-  // equivalent to dayjsClass.extend(duration)
-  localeData(option, dayjsClass, dayjsFactory)
-
+export default (option, dayjsClass) => {
   const oldParse = dayjsClass.prototype.parse
   dayjsClass.prototype.parse = function (cfg) {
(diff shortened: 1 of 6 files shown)


VehpuS:
@iamkun just making sure: is there something that needs fixing / changing before this can be committed?

iamkun:
First of all, thanks for this PR.

One thing I did not get clearly is, seems it changed the original code too much to support this feature.

And I'm afraid I did not know this issue much.

Any more detail you can give us, please? Or is it better for us to leave it as a separate plugin package?

VehpuS:
Hi @iamkun - the changes to the original code are because I did a merge commit to my branch following a change you've made to the dev branch (commit **acd8736**). I did it to ensure my code will merge well with the branch. I can try to create a new pull request without the merge commit if it's confusing - I was only trying to ensure tests pass with the most recent code. I don't think it should mess anything up frankly...

The only manual change I made to existing (plugin) code was to the relative time plugin, where I made an inner logic overridable through the prototype chain to support postFormatting (see fromToBase in the relativeTime changes).

Let me know ho you want me to proceed :)

iamkun:
Ok, I see. 

How should we use this plugin?

As far as I can see, if I want to get the correct `ar` locale in relative time, should we enable preParsePostFormat plugin first?

It's a little bit wired to me, to be honest.

VehpuS:
> Ok, I see.
> 
> How should we use this plugin?
> 
> As far as I can see, if I want to get the correct `ar` locale in relative time, should we enable preParsePostFormat plugin first?
> 
> It's a little bit wired to me, to be honest.

The plugin works fine with or without the relative time component. It is not a requirement.

iamkun:
Then what's this plugin used for? I've checked the code and think it might be related to `ar` locale relative time format string.

Correct me if I am wrong.

VehpuS:
Now I understand the confusion.

Preparse / postformat is a feature in moment that lets you process the input before the parser and process the string output after the formatter. 

In the AR locale specifically, it is used to support Arabic numerals (see the last two commits). I needed to include support for the relative time plugin because in moment it is also processed by this logic.

VehpuS:
It is also used in other locales an I personally used it in a custom locale for work on moment

iamkun:
Thanks for this background information.

As I am not familiar with the concept, could you please give some reference of moment.js in this topic we've talked about?

This could give me a more detailed  understanding of this PR.

VehpuS:
In the docs: https://momentjs.com/docs/#/i18n/locale-data/

In the code: https://github.com/moment/moment/blob/b7ec8e2ec068e03de4f832f28362675bb9e02261/moment.js
(search for preparse / postformat - there aren't many uses in the code)

Moment.js test (on which I based the dayjs tests): https://github.com/moment/moment/blob/develop/src/test/moment/preparse_postformat.js

iamkun:
THX. I'll take a look at the doc you've provied.

VehpuS:
> THX. I'll take a look at the doc you've provied.

Did you get a chance to look at the references?

src/plugin/preParsePostFormat/index.js
```diff
@@ -0,0 +1,53 @@
+import localeData from '../localeData'
+
+// Plugin template from https://day.js.org/docs/en/plugin/plugin
+export default (option, dayjsClass, dayjsFactory) => {
+  // This plugin depends on other plugins - so I will import them here
+  // equivalent to dayjsClass.extend(duration)
+  localeData(option, dayjsClass, dayjsFactory)
```
iamkun, line 7: we don't have to import other plugins like this.

Just listing the dependent plugin in the doc.

iamkun:
Yes, it looks good to me. If I want to get the correct ar locale result, should I use it in this way
```
dayjs.extend(localeData)
dayjs.extend(preParsePostFormat)
dayjs.loclae('ar')

dayjs().formNow()
```

VehpuS:
> Yes, it looks good to me. If I want to get the correct ar locale result, should I use it in this way
> 
> ```
> dayjs.extend(localeData)
> dayjs.extend(preParsePostFormat)
> dayjs.loclae('ar')
> 
> dayjs().formNow()
> ```

You actually don't need to extend localeData, the plugin takes care of it for you.

The ar tests I've added demonstrate the usage of the plugin (https://github.com/iamkun/dayjs/pull/1255/commits/a5b920060bc7f29efe08403a8c176e4e8ed2c36f).

iamkun:
Yes, you can see the comment above, I don't think it's a good idea to import localeData plugin into this plugin

iamkun:
Cause by design, a plugin could be dependent on another one.

https://day.js.org/docs/en/plugin/advanced-format

Check format Week of year as an example here.

VehpuS:
> Cause by design, a plugin could be dependent on another one.
> 
> https://day.js.org/docs/en/plugin/advanced-format
> 
> Check format Week of year as an example here.

As I said - it's a pretty simple fix. I've made it and tests seem to still pass. Let me know if you need me to make any additional changes.

iamkun:
LGTM. Let get this version released and see if there's any improvement that we could do.

iamkun (approved):

ahmadconnectedmotion (commented):
LGTM

PR #1266, opened 2020-12-10
feat: add ordinal to localeData plugin

Xenonym:
Moment.js exposes the ordinal function in [localeData][1].

Let's add ordinal to the localeData plugin as well.

[1]: https://momentjs.com/docs/#/i18n/locale-data/


iamkun:
cool

PR #1287, opened 2020-12-23
fix: update he locale for double units

zvizvi:
Better Hebrew localization


iamkun:
Thanks.

Would you like to add some test,s please? like https://github.com/iamkun/dayjs/blob/dev/test/locale/hu.test.js#L29

zvizvi:
Done.

PR #1292, opened 2020-12-28
fix: small errors in docs

0xflotus:
Shouldn't it be `dayjs()` instead of `dayjs`?

### DIFF
--- a/docs/en/API-reference.md
+++ b/docs/en/API-reference.md
@@ -108,8 +108,8 @@ dayjs(1318781876406)
 Returns a `Dayjs` from a Unix timestamp (seconds since the Unix Epoch)
 
 ```js
-dayjs.unix(1318781876)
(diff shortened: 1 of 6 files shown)


iamkun:
First of all, it is correct `dayjs.unix`

Besides, please check the latest document at https://day.js.org/

0xflotus:
> First of all, it is correct `dayjs.unix`
> 
> Besides, please check the latest document at https://day.js.org/

Ok. But one question: isn't it the latest documentation that I tried to correct?^^

iamkun:
The documentation is correct both on github you've seen and https://day.js.org/.

Just we only update new doc to https://day.js.org/, which is mentioned here https://github.com/iamkun/dayjs/blob/dev/docs/en/API-reference.md#notice

Issue #1281, opened 2020-12-19
🎁 2.0 Milestones【Breaking Change】


iamkun:
We want to make some changes to make Day.js modern and in a standard JS way.

And this would be a ***breaking change*** because of these changes below:

###  1. Mark the correct ES6 Module entry
Update package.json "module" to "esm/index"

###  2. Make TypeScript defenition the same behavior as JS code 
Update `types/index.d.ts` to `export default dayjs;`

iamkun:
I currently have some problem with `2. update TS definition`.

In the current Day.js, while we import some plugin like isLeapYear, it will update the TS definition to add a new method `.isLeapYear`, so that TS could know that `dayjs().isLeapYear` is corrent.

However, after I changing `types/index.d.ts` to `export default dayjs`, the above code does not work at all.

I don't know TS much and need some help, pls.✋🏼

g1eny0ung:
> I currently have some problem with `2. update TS definition`.
> 
> In the current Day.js, while we import some plugin like isLeapYear, it will update the TS definition to add a new method `.isLeapYear`, so that TS could know that `dayjs().isLeapYear` is corrent.
> 
> checking this: https://github.com/iamkun/dayjs/blob/dev/types/plugin/isLeapYear.d.ts
> 
> However, after I changing `types/index.d.ts` to `export default dayjs`, the above code does not work at all.
> 
> I don't know TS much and need some help, pls.✋🏼

Because the original type definitions use `Named Exports` (via `export =`) so everything in the `dayjs` namespace can be imported.

For `Default Exports`, it's needed to import by `import default from` because there only have one export.

TheThing:
I wanna point out, the recent typescript changes done in 1.10 actually broke in my project with the way that it was documented online on how to install/use dayjs in angular

```javascript
import * as updateLocale from 'dayjs/plugin/updateLocale'
import * as relativeTime from 'dayjs/plugin/relativeTime'
import * as LocalizedFormat from 'dayjs/plugin/localizedFormat'
import * as dayjs from 'dayjs'
dayjs.extend(LocalizedFormat)
dayjs.extend(relativeTime)
dayjs.extend(updateLocale)
dayjs.locale('is')
```

Error: `"export 'extend' (imported as 'dayjs') was not found in 'dayjs'`

So there already has been breaking changes, especially with usage on typescript so might as well just push these changes over since people are gonna have to update their projects

iamkun:
[...]

PR #1298, opened 2021-01-03
feat: add ES6 Module Support, package.json module point to "esm/index.js"

### LINKED ISSUE #313: ES Module Support?
I am trying to use this library in a polymer 3 typescript project.  Would it be possible to add support for es modules?

Code
```javascript
import dayjs from 'dayjs';
```

Result
```
my-app.ts:73 One or more components failed to load SyntaxError: The requested module '../node_modules/dayjs/dayjs.min.js' does not provide an export named 'default'
```

Workaround
```
<script src="node_modules/dayjs/dayjs.min.js" async></script>
```

### LINKED ISSUE #598: `module` field in package.json doesn't point to an esm module.
**Describe the bug**
module field in `package.json` points to minified version of the library.

Also causes issues when using `npm link` across modules for development (minified file cannot be found / imported)

**Expected behavior**
module field should point to the esm distribution so that it can be imported by consuming projects and tree-shaken / imported correctly.

**Information**
 - Day.js Version 1.8.14

iamkun:
fix #598, fix #313

### DIFF
--- a/package.json
+++ b/package.json
@@ -4,7 +4,7 @@
   "description": "2KB immutable date time library alternative to Moment.js with the same modern API ",
   "main": "dayjs.min.js",
   "types": "index.d.ts",
-  "module": "dayjs.min.js",
+  "module": "esm/index.js",
   "scripts": {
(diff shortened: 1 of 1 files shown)


iamkun:
relate to #1281

PR #1309, opened 2021-01-04
fix: Add format options for iso week and weekyear

ChiefORZ:
Like moment.js - day.js should allow the format function to be compatible with iso weeks. and since isoweeks are a plugin already, we can just call it.

Input | Example | Description
-- | -- | --
gggg | 2014 | Locale week year
w ww | 1..53 | Locale week of year
GGGG | 2014 | ISO week year
W WW | 1..53 | ISO week of year

### DIFF
--- a/docs/en/Plugin.md
+++ b/docs/en/Plugin.md
@@ -123,18 +123,21 @@ dayjs().format('Q Do k kk X x')
 
 List of added formats:
 
-| Format | Output                | Description                                           |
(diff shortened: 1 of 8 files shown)


iamkun:
This LGTM.

Would you like to update the doc here https://github.com/dayjs/dayjs-website?

The doc in the readme is no longer maintained anymore.

iamkun (approved):

PR #1317, opened 2021-01-06
Fix negative duration

kamontat:
# Problem

When I create Duration with negative number, the default value of each unit is -1 since you round everything down (Math.floor).

## Solution

I add if case when number is negative it should run Math.ceil() instead.

## Example

For positive number
```js
const duration = dayjs.duration(1, "milliseconds")
console.log(duration) // { year: 0, month: 0, day: 0, hour: 0, minute: 0, second: 0, millisecond: 1 }
```

For negative number (current)
```js
const duration = dayjs.duration(-1, "milliseconds")
console.log(duration) // { year: -1, month: -1, day: -1, hour: -1, minute: -1, second: -1, millisecond: -1 }
```

For negative number (expected)
```js
const duration = dayjs.duration(-1, "milliseconds")
console.log(duration) // { year: 0, month: 0, day: 0, hour: 0, minute: 0, second: 0, millisecond: -1 }
```

### DIFF
--- a/src/plugin/duration/index.js
+++ b/src/plugin/duration/index.js
@@ -25,6 +25,7 @@ const wrapper = (input, instance, unit) =>
   new Duration(input, unit, instance.$l) // eslint-disable-line no-use-before-define
 
 const prettyUnit = unit => `${$u.p(unit)}s`
+const roundNumber = number => number < 0 ? Math.ceil(number) : Math.floor(number)
 
 class Duration {
   constructor(input, unit, locale) {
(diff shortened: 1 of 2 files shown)


test/plugin/duration.test.js
```diff
@@ -27,6 +27,11 @@ describe('Creating', () => {
     expect(dayjs.duration(60, 'seconds').toISOString()).toBe('PT1M')
     expect(dayjs.duration(13213, 'seconds').toISOString()).toBe('PT3H40M13S')
   })
+  it('two argument will bubble up to the next (negative number)', () => {
+    expect(dayjs.duration(-59, 'seconds').toISOString()).toBe('PT-59S')
+    expect(dayjs.duration(-60, 'seconds').toISOString()).toBe('PT-1M')
+    expect(dayjs.duration(-13213, 'seconds').toISOString()).toBe('PT-3H-40M-13S')
```
iamkun, line 33: seems this is not the correct result?
-PT59S
-PT1M
-PT3H40M13S

kamontat: Fixed

src/plugin/duration/index.js
```diff
@@ -136,11 +188,11 @@ class Duration {
     if (pUnit === 'milliseconds') {
       base %= 1000
     } else if (pUnit === 'weeks') {
-      base = Math.floor(base / unitToMS[pUnit])
+      base = roundNumber(base / unitToMS[pUnit])
     } else {
       base = this.$d[pUnit]
     }
-    return base
+    return Object.is(base, -0) ? 0 : base
```
iamkun, line 195: Do we have to use Object.is here? I think it is not supported on IE

kamontat: updated. use `a === 0` instead

iamkun:
Cool, thanks

src/plugin/duration/index.js
```diff
@@ -25,6 +25,7 @@ const wrapper = (input, instance, unit) =>
   new Duration(input, unit, instance.$l) // eslint-disable-line no-use-before-define
 
 const prettyUnit = unit => `${$u.p(unit)}s`
+const roundNumber = number => number < 0 ? Math.ceil(number) : Math.floor(number)
```
flamesdev: This could be replaced with `Math.abs(Math.trunc(number)) || 0`, which is shorter.

iamkun: seems not supported in IE :-(

PR #1338, opened 2021-01-15
fix: Significantly improve `Duration` types

zardoy:
This PR improves [Duration plugin](https://day.js.org/docs/en/durations/durations) types.
## Why?
- Autocomplete:
autocomplete

- And for more strict checks (just compare):
pr-1
pr-2

## New types

```ts
// That's how computed type looks like
type plugin.DurationUnitsObjectType = {
    milliseconds?: number;
    seconds?: number;
    minutes?: number;
    hours?: number;
    days?: number;
    months?: number;
    years?: number;
    weeks?: number;
}
```
I also added `CreateDurationType` to reduce code duplication, so `duration()`, `duration.add()` and `duration.substract()` using the same function signature.
Other changes you can view in diff. Please, let me know if don't understand something. I can explain every line of code.

UPD: With this PR TypeScript users **can't** use functions `duration()`, `duration.add()` and `duration.substract()` without args any more. But I don't think it's a breaking change, because I can't find any example usage in docs for this. In this way we just get empty duration object.

P.S. It also would be good to update TypeScript in devDependencies to latest version.

### DIFF
--- a/test/plugin/duration.test.js
+++ b/test/plugin/duration.test.js
@@ -201,7 +201,7 @@ describe('Years', () => {
 describe('prettyUnit', () => {
   const d = dayjs.duration(2, 's')
   expect(d.toISOString()).toBe('PT2S')
-  expect(d.as('Second')).toBe(2)
+  expect(d.as('seconds')).toBe(2)
   expect(d.get('s')).toBe(2)
   expect(dayjs.duration({
     M: 12,
--- a/types/index.d.ts
+++ b/types/index.d.ts
@@ -13,11 +13,11 @@ declare namespace dayjs {
 
   export type OptionType = { locale?: string, format?: string, utc?: boolean } | string | string[]
 
-  type UnitTypeShort = 'd' | 'M' | 'y' | 'h' | 'm' | 's' | 'ms'
+  export type UnitTypeShort = 'd' | 'M' | 'y' | 'h' | 'm' | 's' | 'ms'
 
-  type UnitTypeLong = 'millisecond' | 'second' | 'minute' | 'hour' | 'day' | 'month' | 'year' | 'date'
+  export type UnitTypeLong = 'millisecond' | 'second' | 'minute' | 'hour' | 'day' | 'month' | 'year' | 'date'
 
-  type UnitTypeLongPlural = 'milliseconds' | 'seconds' | 'minutes' | 'hours' | 'days' | 'months' | 'years' | 'dates'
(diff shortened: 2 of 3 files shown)


iamkun:
Thanks. Will the syntax require the latest Typescript run time? Cause I can see some of our users still using very old typescript like 1.x

zardoy:
Thank you for noticing that, but Day.js already uses `Partial`, that [was introduced](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-2-1.html#partial-readonly-record-and-pick) in TypeScript 2.1 https://github.com/iamkun/dayjs/blob/fe5f1d0afbe57b70339e268047e6c3028ca3d59b/types/index.d.ts#L414
And `unknown` type, that [was introduced](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-0.html#new-unknown-top-type) in TypeScript 3.0
https://github.com/iamkun/dayjs/blob/fe5f1d0afbe57b70339e268047e6c3028ca3d59b/types/index.d.ts#L412
But it's not a big problem because if TypeScript does not understand something it will fallback to `any` type.
Just compare **computed** variant of `extend` function:
```ts
// TypeScript 1.8.10 (TypeScript doesn't understand this Generic so it fallbacks to `any` type)
function extend<T, unknown>(plugin: any, option?: T): Dayjs
// TypeScript 2.9.2 (TypeScript still doesn't understand `unknown` type)
function extend<T = any>(plugin: PluginFunc<T>, option?: T): Dayjs
// TypeScript 3.9.7
function dayjs.extend<T = unknown>(plugin: PluginFunc<T>, option?: T): Dayjs
```

So the answer is: It will work, but with fallbacks to `any` type:
```ts
type plugin.CreateDurationType = ((units: any) => Duration) & ((time: number, unit?: any) => Duration) & ((ISO_8601: string) => Duration)
```
And only if the user uses TypeScript 2.x all the types will work **correctly**.
I think it would be good to encourage users to upgrade TypeScript version. Breaking changes for all TypeScript versions could be found here: https://github.com/Microsoft/TypeScript/wiki/Breaking-Changes#typescript-20 .

zardoy:
Don't worry intellisense will work correctly (even if the user still uses TypeScript 1.x): 
pr

iamkun:
LGTM

iamkun:
Thank you, TS expert.

Just wondering if you could help us with another TS issue if you got some time. 

https://github.com/iamkun/dayjs/issues/1281

Issue #1264, opened 2020-12-09
weekOfYear shows wrong calender week at the turn of the year


ttusek:
Describe the bug
The 2020-12-30 should be calender week 53 but it shows as calender week 1.
(similar issue: #760)

Calender weeks:
http://www.whatweekisit.org/

Information

Day.js Version 1.9.7
Browser Chrome latest

iamkun:
Hi, have you set the correct locale?

ttusek:
yes, it is the same issue when i set 'us' | 'de'
https://codepen.io/tompach/pen/OJRRpNE?editors=0012

ChuanfengZhang:
@iamkun I have the same issue. 
https://codepen.io/Chuanfeng/pen/PoGBwrp?editors=0002

iamkun:
You can use `en-gb` locale to get the correct result ATM.


will update some other locale like `en-us`, `de`  to set correct year star day.

Uninen:
I have this same issue. Using `en-gb` as locale doesn't seem to fix it. Can't get to week 53 with any locale I've tried.

iamkun:
@Uninen a reproduction demo please? you can share your code here https://runkit.com/npm/dayjs

Uninen:
Sure, here: https://runkit.com/embed/5wk28cq51vf9

(Note, the dates here are dynamic so this example may or may not work in the future.)

Week for Mon Dec 28 2020 returns 1 (should be 53), week for Sun Jan 10 2021 returns 3 (should be 1) using locale `en-gb`.

iamkun:
@Uninen this is because you haven't load `en-gb` locale https://day.js.org/docs/en/i18n/loading-into-nodejs

Uninen:
Oh! 🤦‍♂️ Thanks, it works with `en-gb` locale when the locale is properly set.

I missed the import as there were no errors or even TypeScript hints with the locale-call even though it wasn't imported.

ChuanfengZhang:
@iamkun What about the default locale en?
https://runkit.com/zhangcf/5ffbb211b8e405001f893751

MarinaFernandes:
@iamkun I'm not sure if this is the best place to report this but the locale `pt` still shows the wrong calender week at the turn of the year.

example: https://runkit.com/marferna/6006c2f0e5d69000136b7ad0

Issue #760, opened 2019-12-29
weekOfYear shows wrong calender week at the turn of the year


Khazl:
**Describe the bug**
I probably found a bug in the weekOfYear plugin.
The 2019-12-30 should be calender week 1 but it shows as calender week 53.

Here my example at codepen:
https://codepen.io/khazl/pen/NWPvoYb?editors=0012

Calender weeks:
http://www.whatweekisit.org/

**Information**
 - Day.js Version 1.8.17
 - Browser Chrome latest

iamkun:
Thanks. Will fix in next release

PR #1348, opened 2021-01-21
fix: update Polish [pl] locale yearStart

grzegorz-s:
This pull request fixes issue with weekOfYear showing wrong calendar week at the turn of the year.

Eg. the 2021-01-21 should be calendar week 3 but it shows as calendar week 4.
(similar issues: #760 #1264)

Calendar weeks:
http://www.whatweekisit.org/


iamkun:
Thanks

Issue #1340, opened 2021-01-18
Inconsistent behaviour for `utcOffset`.


y-abarnikov:
**Describe the bug**
Inconsistent behaviour for `utcOffset`. Probably issue with Daylight Saving Time.
![image](https://user-images.githubusercontent.com/40033983/104925721-dac9c980-59a7-11eb-87b4-4ee00024bb14.png)

**Expected behavior**
Should be `-4` as in the row below.

**Information**
 - Day.js Version [1.10.3]
 - moment-timezone Version [0.5.32]
 - OS: [Mac OS Big Sur]
 - Interpreter Node.js [14.15.0]
 - Time zone: [GMT+2 (Eastern European Time)]

y-abarnikov:
@iamkun

addisonElliott:
FYI, it seems the bug is in moment to me.

Right now the correct offset for `America/New_York` is -5.

See here: https://www.zeitverschiebung.net/en/timezone/america--new_york

hansdaniels:
DST is picked up by `tz`, but strangely, it is added to the hour only, not to the offset:
```javascript
import dayjs from 'dayjs';
import tz from 'dayjs/plugin/timezone.js';
import utc from 'dayjs/plugin/utc.js';
dayjs.extend(tz);
dayjs.extend(utc);

console.log(dayjs.tz(new Date('2020-03-29T00:30:00Z').getTime(), 'Europe/Berlin').format(), 'is correct');
console.log(dayjs.tz(new Date('2020-03-29T01:30:00Z').getTime(), 'Europe/Berlin').format(), 'should be 2020-03-29T03:30:00+02:00, because DST kicked in half an hour before');
```
yields:
```
console.log(dayjs.tz(new Date('2020-03-29T00:30:00Z').getTime(), 'Europe/Berlin').format(), 'is correct');
console.log(dayjs.tz(new Date('2020-03-29T01:30:00Z').getTime(), 'Europe/Berlin').format(), 'should be 2020-03-29T03:30:00+02:00, because DST kicked in half an hour before');
```

addisonElliott:
Not sure your bug is similar to the OPs, unless I'm missing the connection.

But I did confirm and figure out what the problem is with your case. It's actually a pretty easy fix (I think).

## The problem

I'm in CST, which has an offset of -5/-6 (depending on DST). Let's use this as an example.

Given the input you gave: `2020-03-29T00:30:00Z`

[...]

Issue #1001, opened 2020-08-12
Timezone plugin unexpected daylight calculation


ealanisg:
**Describe the bug**
When using the timezone plugin the date calculated by `dayjs` is unexpected for `daylight datetime` calculation, I'm comparing towards moment-timezone.

**Expected behaviour**
```
const moment = require('moment-timezone');
const dayjs = require('dayjs');
const utc = require('dayjs/plugin/utc');
const tz = require('dayjs/plugin/timezone');

dayjs.extend(utc);
dayjs.extend(tz);

const timezone = 'America/Vancouver';
const currentDate = new Date('2020-03-08T12:00-07:00');
const nowMoment = moment(currentDate);
const nowDayJs = dayjs(currentDate);

// milliseconds for currentDate: 1583694000000
console.log(nowDayJs.valueOf() === nowMoment.valueOf()); // -> true

// daylight
// https://www.timetemperature.com/canada/daylight_saving_time_canada.shtml
const plainDate = currentDate.toISOString().slice(0, 10); // 2020-08-08
const dlMoment = moment.tz(plainDate, timezone).set({
  hour: 12,
  minute: 30
});
const dlDayJs = dayjs.tz(plainDate, timezone)
  .set('hour', 12)
  .set('minute', 30);
console.log(dlDayJs.format() === dlMoment.format()); // -> false (true expected)
// 2020-03-07T12:30:00-08:00 !== 2020-03-08T12:30:00-07:00

```

**Information**
- dayjs: 1.8.33
- moment-timezone: 0.5.31
- OS: Ubuntu 18.04.4 LTS
- Nodejs 12.18.0
- Time zone: GMT-07:00 DST (Pacific Daylight Time)

simlu:
What's the status on this?

nathansollenberger-penguinpay:
I have run in to the same issue. Here is my test case, which is a bit simpler than the one above:

Expected:

```
import dayjs from 'dayjs';
const utc = require('dayjs/plugin/utc');
const timezone = require('dayjs/plugin/timezone');
dayjs.extend(utc);
dayjs.extend(timezone);

function getDateInTimezone(date, TZ = 'America/New_York') {
  return dayjs.tz(date, TZ);
}

describe('dayjs.tz', function () {
  it('Handles DST', function () {
    expect(getDateInTimezone('2020-03-02').toDate()).toEqual(new Date('2020-03-02T05:00:00.000Z'));
[...]

PR #1352, opened 2021-01-22
fix: tz with DST date while DST is not active (#1340)

hansdaniels:
@addisonElliott, your last suggestion made my new tests work and also no longer break the existing tests. 
No complains about `keepLocalTime`.


addisonElliott:
Glad to hear it. This should resolve most DST issues people have experienced.

I would recommend adding tests for the bug that this fixes. Using that specific date you had in the issue should work for a good test.

hansdaniels:
I've added 2 `expect`s to an already existing test case for DST in `test/plugin/timezone.test.js`. One date with DST, one without, exactly what I needed. Before I have applied your suggested changes, the DST `expect` failed, afterwards both passed.

The exact date minutes before and after a DST change for 'Europe/Berlin' aren't any more special than dates in December and June for 'America/Los_Angeles' as used in the "DST" testcase.

hansdaniels:
Side-Problems:
- Commiting from Windows is hard, `eslint` complains when having files checked out with CRLF (even if I have told Git to commit with LF).
- The pre-commit hook failed because `eslint` wasn't a known command. I've circumvented this by running `node node_modules/eslint/bin/eslint.js  src/* test/* build/*` manually and commiting with `-n` which shouldn't be necessary to my eyes.

alex-w0:
Is this pull request related to [this issue](https://github.com/iamkun/dayjs/issues/1001)?

addisonElliott:
@alex-w0 I think so, but I'm not entirely sure. This bug fixes UTC offsets for DST-related timezones. Best way is going to be trying the failing tests from the issue after applying this change.

iamkun:
nice catch

addisonElliott:
@iamkun Do you need anything in order to merge this?

ashconnell:
I think this commit has just made things worse.

I'm specifically seeing this issue: https://github.com/iamkun/dayjs/issues/1437

So I checked this commit out on `dev` and added a test for my real world scenario:

```js
it('DST startOf(day)', () => {
  const beforeDST = dayjs
    .tz('2021-04-01 08:00', 'Australia/Melbourne')
    .startOf('day')
    .format()
  expect(beforeDST).toBe('2021-04-01T00:00:00+11:00')
  const afterDST = dayjs
    .tz('2021-04-15 08:00', 'Australia/Melbourne')
    .startOf('day')
    .format()
  expect(afterDST).toBe('2021-04-15T00:00:00+11:00')
})
```

And the output of the second expect comes out like this:

```
2021-04-15T00:00:00+10:00
```

Note that the offset is +10:00 and not +11:00

henhal:
This commit seems to be the reason for quite a few problems, including #1795 , which seems to be similar to what @ashconnell reported almost a year ago. This is really critical.

kevinsimper:
I think this also gives problems
Here you can see a dayjs object
```
    M {
      '$L': 'en',
      '$offset': 60,
      '$d': 2021-01-04T08:00:00.000Z,
      '$x': { '$timezone': 'Europe/Stockholm' },
      '$y': 2021,
      '$M': 0,
      '$D': 4,
      '$W': 1,
      '$H': 9,
      '$m': 0,
      '$s': 0,
      '$ms': 0
    } 
```
and then running toISOString() gives this output now
```
2021-01-04T09:00:00.000Z
```

I upgraded from 1.10.4 to 1.11.0

This is the result in 1.10.4
```
      d {
        '$L': 'en',
        '$offset': 120,
        '$d': 2021-01-04T08:00:00.000Z,
        '$x': { '$timezone': 'Europe/Stockholm' },
        '$y': 2021,
        '$M': 0,
        '$D': 4,
        '$W': 1,
        '$H': 9,
        '$m': 0,
        '$s': 0,
        '$ms': 0
      } 
2021-01-04T08:00:00.000Z
```

PR #1378, opened 2021-02-10
 fix: update Finnish [fi] locale yearStart config

rjokelai:
The Finnish locale should have 2020-12-31 as week 53. According to https://en.wikipedia.org/wiki/Date_and_time_notation_in_Finland
"The week begins with a Monday and week 1 is the week containing the year's first Thursday"


iamkun:
Thanks

PR #1385, opened 2021-02-15
Set Swedish [sv] locale to use ISO 8061 weeks

frehoy:
Sweden uses ISO 8061 weeks, i.e. they follow the convention that the week that contains Jan 4th is the first week of the year.

Followed the example of #1378


iamkun:
THX

PR #1386, opened 2021-02-15
fix: Add ordinal support for Slovenian [sl]

sesam:
This implements Slovenian ordinals.

The language has genders and cases, but in short form just adding a dot at the end is how it's typically done, and really the only option here.


iamkun:
THX

PR #1394, opened 2021-02-28
Change string to number in duration object

### LINKED ISSUE #1389: Duration returns strings instead of numbers
**Describe the bug**
When parsing a duration string, eg `dayjs.duration('P1Y2M3W4DT12H34M56S')` and logging the result, it returns the values as strings instead of numbers:

image

**Expected behavior**
It should return numbers, as given by the return type in `plugins/duration.d.ts`.

**Information**
 - Day.js Version 1.10.4
 - OS: OSX 11.2.1
 - Browser: Chrome 88.0.4324.182
 - Time zone: GMT+7

mateuszpigula:
Fixes #1389 

I based on `parseFromMilliseconds` function.

I noticed there are already two failing tests:
```
 ● Convert › convert to target time

    expect(received).toBe(expected) // Object.is equality
    
    Expected value to be:
      1401624000000
    Received:
      1401620400000

```

### DIFF
--- a/src/plugin/duration/index.js
+++ b/src/plugin/duration/index.js
@@ -79,14 +79,17 @@ class Duration {
     if (typeof input === 'string') {
       const d = input.match(durationRegex)
       if (d) {
-        this.$d.years = Number(d[2])
-        this.$d.months = Number(d[3])
-        this.$d.weeks = Number(d[4])
-        this.$d.days = Number(d[5])
-        this.$d.hours = Number(d[6])
-        this.$d.minutes = Number(d[7])
-        this.$d.seconds = Number(d[8])
-
+        const [,, ...properties] = d
+        const numberD = properties.map(value => Number(value));
+        [
+          this.$d.years,
+          this.$d.months,
+          this.$d.weeks,
+          this.$d.days,
+          this.$d.hours,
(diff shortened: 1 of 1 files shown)


src/plugin/duration/index.js
```diff
@@ -79,15 +79,14 @@ class Duration {
     if (typeof input === 'string') {
       const d = input.match(durationRegex)
       if (d) {
-        [,,
```
iamkun, line 82: let numberD = d.map(_=>Number(_))

will save some bytes I think

mateuszpigula: Yep, great point @iamkun!

src/plugin/duration/index.js
```diff
@@ -79,15 +79,17 @@ class Duration {
...
+        const [,, ...properties] = d
```
iamkun, line 82: it is not advised to use `...` is dayjs cause it will compile to a lot es5 codes to make it work.

mateuszpigula: It is worth mentioning in `CONTRIBUTING.md` as there is now: 
> Day.js is written in ES6.

iamkun:
Cool

PR #1401, opened 2021-03-07
fix: fixed missing types for ArraySupport plugin

### LINKED ISSUE #1170: Typings for arraySupport missed
**Describe the bug**
arraySupport (https://github.com/iamkun/dayjs/issues/1129) lacks Typescript typings

**Expected behavior**
It should be possible to use arrays with typescript

**Information**
 - Day.js Version 1.9.4
 - OS: Win10
 - Browser Chrome
 - Time zone: GMT+3

JasonHK:
I've added an interface called `ConfigTypeMap`, so other plugins could add their own types to `ConfigType`. The same fix could be applied to ObjectSupport in a future pull request.

Fixes #1170

### DIFF
--- a/types/index.d.ts
+++ b/types/index.d.ts
@@ -9,7 +9,11 @@ declare function dayjs (date?: dayjs.ConfigType, format?: dayjs.OptionType, stri
 declare function dayjs (date?: dayjs.ConfigType, format?: dayjs.OptionType, locale?: string, strict?: boolean): dayjs.Dayjs
 
 declare namespace dayjs {
-  export type ConfigType = string | number | Date | Dayjs
+  interface ConfigTypeMap {
+    default: string | number | Date | Dayjs
+  }
+
+  export type ConfigType = ConfigTypeMap[keyof ConfigTypeMap]
 
   export type OptionType = { locale?: string, format?: string, utc?: boolean } | string | string[]
 
--- a/types/plugin/arraySupport.d.ts
+++ b/types/plugin/arraySupport.d.ts
@@ -2,7 +2,7 @@ import { PluginFunc } from 'dayjs'
 
 declare module 'dayjs' {
   interface ConfigTypeMap {
-    arraySupport: [year?: number, month?: number, day?: number, hour?: number, minute?: number, second?: number, millisecond?: number]
+    arraySupport: [number?, number?, number?, number?, number?, number?, number?]
   }
 }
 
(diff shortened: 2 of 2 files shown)


types/index.d.ts
```diff
@@ -9,7 +9,11 @@ declare function dayjs (date?: dayjs.ConfigType, format?: dayjs.OptionType, stri
 declare function dayjs (date?: dayjs.ConfigType, format?: dayjs.OptionType, locale?: string, strict?: boolean): dayjs.Dayjs
 
 declare namespace dayjs {
-  export type ConfigType = string | number | Date | Dayjs
+  interface ConfigTypeMap {
+    default: string | number | Date | Dayjs
+  }
+
+  export type ConfigType = ConfigTypeMap[keyof ConfigTypeMap]
```
iamkun, line 16: what's this used for please? And will this affect the original type if the plugin is imported?

JasonHK: @iamkun The `type` keyword doesn't support [declaration merging](https://www.typescriptlang.org/docs/handbook/declaration-merging.html). And yes, it will add the plugin's type into `ConfigType`.

`ConfigTypeMap[keyof ConfigTypeMap]` basically means "merge every members' type into a union". If no plugin was imported, the type of `ConfigType` is `string | number | Date | Dayjs` (same as before). When the ArraySupport plugin was imported, `ConfigType` will become `string | number | Date | Dayjs | [year?: number, ...]`.

iamkun: looks cool, does this requires higher ts version?

JasonHK: @iamkun Labeled tuple elements require TypeScript 4.0 or higher, I can convert it back to un-labelled ones if you want to support older TypeScript versions.

iamkun: that's what I'm concerned. We should have a fallback support at least ts 3 I think

JasonHK: @iamkun It should work on TypeScript 3 now.

iamkun: can we make sure that `ConfigTypeMap[keyof ConfigTypeMap]` works on ts3?

JasonHK: > can we make sure that `ConfigTypeMap[keyof ConfigTypeMap]` works on ts3?

@iamkun Yes, this will work on TypeScript 3. You can verify it on [TypeScript playground](https://www.typescriptlang.org/play?ts=3.3.3#code/JYOwLgpgTgZghgYwgAgMpwLYAcA2ECycWAUAN7HKXIAmE8ArjmAFzIDOYUoA5sgD7IQ9DACNoAbgpUA9mAAW0VgBE4kSQF9ixMAE8sKdNjzIAvGky4CRANoBrCDukxzRq1gC6k4kA).

iamkun:
THX

PR #1403, opened 2021-03-08
feat: Add ru meridiem

cjskillingstad:
- Added ru meridiem
- Added unit tests
- Using same logic as moment's [ru meridiem](https://github.com/moment/moment/blob/f5233ee5d44ba20079cd4b389262719bb20e23bd/locale/ru.js#L183-L191)


iamkun:
THX

Issue #1418, opened 2021-03-17
Allow custom parse function for two digit years


sbrunecker:
## Current Behavior

`dayjs`  parses a two digit year as 19xx if the value is below 68, otherwise as 20xx.

## Proposal

A possibility to configure this behavior would be nice. As a use case, we need to parse a two digit year as 19xx if the input is greater than the current year, otherwise as 20xx.

In `moment` this can be configured, e.g.:

```
moment.parseTwoDigitYear = function(yearString) {
    return parseInt(yearString) + 2000;
}
```

iamkun:
maybe we can support this via a plugin

sbrunecker:
Would it be an idea to include it as a plugin option to the existing `customParseFormat`? So it could be configured like so: 
```
dayjs.extend(customParseFormat, {
    twoDigitYearParseFn: yearString => (+yearString) + ((+yearString) > (+dayjs().format('YY')) ? 1900 : 2000)
  })
```

iamkun:
fixed in #1421

maurer2:
Hello, I noticed that in the current datejs version (1.11.7), the pivot year seems to be different than mentioned above.
The default pivot year now seems to be 1950, instead of 1968 e.g.

```js
dayjs('12/31/49').format() -> 2049-12-31T00:00:00+00:00
dayjs('01/01/50').format() -> 1950-01-01T00:00:00+00:00
```

Is that correct? Thanks

PR #1421, opened 2021-03-18
feat: support custom two-digit year parse function

sbrunecker:
As proposed in #1418, with this PR a custom two-digit year parse function can be configured in the `customParseFormat` plugin


iamkun:
seems a nice update, should we update the config name from `twoDigitYearParseFn ` to `parseTwoDigitYear` to keep the same with momentjs?

iamkun:
THANKS

PR #1429, opened 2021-03-24
fix: update `monthDiff` remove number conversion

yuler:


iamkun:
what's this PR for, please?

yuler:
I think don't need use `+` wrap again.

iamkun:
This function is from momentjs, you can check their test to see the edge case here

yuler:
I checked at the [moment monthDiff function](https://github.com/moment/moment/blob/develop/moment.js#L3901)

It doesn't wrap with `+`.

PR #1443, opened 2021-04-07
Updating the docs & TS Type File to be able to use strict parsing with UTC plugin

### LINKED ISSUE #1027: Strict parsing support with UTC plugin
Hi,

Looking at the `dayjs` function signature and at the [String + Format](https://day.js.org/docs/en/parse/string-format) documentation there is a strict parsing mode using a 3rd (or 4th when using locale) boolean flag:

    dayjs(date?: dayjs.ConfigType, format?: dayjs.OptionType, strict?: boolean)

Looking at the corresponding [UTC plugin](https://day.js.org/docs/en/parse/utc), it doesn't seem to have such option available:

    dayjs.utc(config?: dayjs.ConfigType, format?: string)

Did I miss something obvious in the documentation?

If not, is it something that's worth be supported in the future?

Thanks!

LucasHFS:
Attemp to fix #1027 . If there are anything I need to change or make, let me know.

### DIFF
--- a/docs/en/Plugin.md
+++ b/docs/en/Plugin.md
@@ -87,13 +87,16 @@ dayjs
   .local()
   .format() //2019-03-06T17:11:55+08:00
 dayjs.utc('2018-01-01', 'YYYY-MM-DD') // with CustomParseFormat plugin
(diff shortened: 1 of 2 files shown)


iamkun:
Thanks

PR #1450, opened 2021-04-09
build: Add npm prepare script

aloisklink:
Adds an `npm prepare` script to package.json

`npm prepare` runs automatically when doing `npm install` from a git dependency.

This will allow users to use dayjs with npm directly from the git repo, e.g. with:

```bash
npm install https://github.com/iamkun/dayjs.git
```

(feel free to test it by installing this branch, with `npm install https://github.com/aloisklink/dayjs.git#add-npm-prepare-script`)

It should be helpful for testing `dayjs`, as well as if people want to use bugfixes before dayjs pushes releases to NPM.

See https://docs.npmjs.com/cli/v7/using-npm/scripts for more details about `npm prepare`.


iamkun:
I'm not sure if there will be some side effects or not. I'll check the docs later.

Issue #929, opened 2020-06-11
customParseFormat strict mode bug


sheffieldnick:
**Describe the bug**
Using the format string "YYYY-MM-DD HH:mm:ss ZZ", parsing certain strings returns Invalid Date when it should return a date:
```
var format = "YYYY-MM-DD HH:mm:ss ZZ";
var ok = dayjs("2018-03-01 00:01:00 +0000", format, true);
var bad = dayjs("2018-04-01 00:01:00 +0000", format, true);

bad is set to {
"$D": 1
"$H": 1
"$L": "en"
"$M": 3
"$W": 0
"$d": Invalid Date
"$m": 1
"$ms": 0
"$s": 0
"$u": undefined
"$y": 2018
}
```
The parsing is successful if strict mode is not used.

**Expected behavior**
`bad` should be a valid date object.

**Information**
 - Day.js Version v1.8.28
 - OS: Windows 7
 - Browser Firefox 77, Chrome 83
 - Time zone: [e.g. UTC+00:00 DST (British Summer Time)]

z0al:
We have the same problem. I created this [codesandbox](https://codesandbox.io/s/affectionate-noether-uc0ku?file=/src/App.js) example to reproduce the issue

waspar:
R.I.P 👌

DDoerner:
I've looked into the code and so far I see only two options for how this bug could be fixed:

1. The timezone information that is extracted by makeParser() could be passed back to the main method in addition to the parsed Date object. It can then be used to format the final result with a custom timezone before the string comparison is made. This however would mean to either require the timezone plugin for the customParseFormat plugin to work or to recreate this functionality in customParseFormat, which might be overkill
2. Rework how strict parsing in customParseFormat works by giving each segment expression information on how to enforce a strict format

iamkun:
Bug confirmed. Strict mode does not work good with parsing timezone (ZZ)

I'm looking for a way to solve it.

AllainPL:
I am trying to migrate some code from moment to dayjs, and this strict check is a big problem for me :(, any idea how to solve it?

ivansieder:
@iamkun is there any known workaround available right now? Or do you eventually have an estimated time until when this can be looked into? Thanks in advance

ivansieder:
[...]

PR #1467, opened 2021-04-21
fix: customParseFormat strict mode bug when parsing with timezones (#929)

ivansieder:


src/plugin/customParseFormat/index.js
```diff
@@ -221,11 +231,18 @@ export default (o, C, d) => {
...
+      const { parsedDate, parsedOffsetMilliseconds } = parseFormattedInput(date, format, utc)
+      this.$d = parsedDate
       this.init()
       if (pl && pl !== true) this.$L = this.locale(pl).$L
-      if (isStrict && date !== this.format(format)) {
-        this.$d = new Date('')
+      const currentOffset = this.utcOffset()
+      if (isStrict) {
+        if (
+          (parsedOffsetMilliseconds !== undefined && date.replace(/[+-]\d\d:?\d\d$/, '') !== this.subtract(currentOffset, 'minute').subtract(parsedOffsetMilliseconds, 'millisecond').format(format.replace(/(ZZ|ZZZ)$/, '')))
```
tomaszgiba, line 241: Why is `ZZZ` (triple zed) here? I don't see in the docs: https://day.js.org/docs/en/parse/string-format

iamkun:
Would you please add the relevant test, please?

ivansieder:
@iamkun I've added the missing tests. Totally forgot about that.

iamkun:
Thanks for the PR, need some time to review this.

iamkun:
Just a small concern, can we re-order the logic a little bit?

Cause to me, it's a little complicated to understand the code of this PR.

ivansieder:
@iamkun I've tried to keep the order of the logic similar to how it was before this PR in order to make merging easier and keep the logic mostly in the same place as it was before. What kind of refactoring did you think about?

src/plugin/customParseFormat/index.js
```diff
@@ -221,11 +231,18 @@ export default (o, C, d) => {
...
-      if (isStrict && date !== this.format(format)) {
-        this.$d = new Date('')
+      const currentOffset = this.utcOffset()
+      if (isStrict) {
+        if (
+          (parsedOffsetMilliseconds !== undefined && date.replace(/[+-]\d\d:?\d\d$/, '') !== this.subtract(currentOffset, 'minute').subtract(parsedOffsetMilliseconds, 'millisecond').format(format.replace(/(Z|ZZ)$/, '')))
+          || (parsedOffsetMilliseconds === undefined && date !== this.format(format))
+        ) {
+          this.$d = new Date('')
+        }
```
imwh0im, line 245:
```suggestion
const replace_date = date.replace(/[+-]\d\d:?\d\d$/, '')
const sub_date = this.subtract(currentOffset, 'minute').subtract(parsedOffsetMilliseconds, 'millisecond').format(format.replace(/(Z|ZZ)$/, ''))

if (isStrict) {
  if (
    (parsedOffsetMilliseconds && replace_date !== sub_date) ||
    (!parsedOffsetMilliseconds && date !== this.format(format))
  ) {
    this.$d = new Date('')
  }
```

it is same logic?

imwh0im (changes requested):

ivansieder:
@imwh0im I've updated the code now with the dev branch and included your suggested changes (changed the variable naming a little bit). Now waiting for further feedback from @iamkun

BreakBB:
Any news on this?

BePo65:
Perhaps it would be a good idea to allow for all time zone formats that dayjs supports in line 253?  
In customParseFormat the definition for zone strings is `const matchOffset = /[+-]\d\d:?(\d\d)?|Z/ // +00:00 -00:00 +0000 or -0000 +00 or Z`.

BePo65:
@iamkun : the check for strictness (even in the original implementation) makes use of the 'format' function of dayjs. But the 'format' definition for the token 'Do' (Day of Month with ordinal) requires the AdvancedFormat plugin.

Probably it would be a good idea to add a hint to the ['String + Format' documentation](https://day.js.org/docs/en/parse/string-format) saying the using the 'Do' format requires the [AdvancedFormat](https://day.js.org/docs/en/plugin/advanced-format) plugin to work.

EriksonBahr:
Hello @iamkun, are you willing to accept this PR? If not, what's missing?

Irian:
Any news on this PR? I see there are conflicts again...

Issue #1473, opened 2021-04-25
New Feature based on dayjs().startOf() function, dayjs().startOf('yearisoweek')


rentianxiang:
**Describe the bug**
This is NOT a bug, it's just a feature i wish dayjs to have.

**Expected behavior**
So my project requires to translate between an ISO8601 date(year week weekday) and an actual Date object.
I found dayjs can translate Date to ISO8601 easily by using isoWeek and isoWeekday, but it does not support for the reverse action.
i tried with dayjs('2021', 'YYYY').isoWeek(1), but it returns the first isoweek of 2020 :(

So i added a new function under dayjs().startOf(), named "yearisoweek" (idk what it should be called..) to calculate the first isoWeek and isoWeekday for a given date (eg. dayjs('2021-04-25', 'YYYY-MM-DD).startOf('yearisoweek) == dayjs('2021-01-04, 'YYYY-MM-DD')), so it gives 2021W011. After it founds the first isoWeek and first isoWeekday of this year, it should be easy to calculate other ISO8601 dates by using the dayjs.add(week, 'week') and dayjs().add(day, 'day') functions.

In the test case, i found my method is not included in moment.js, it cannot be work with the startOf test in the badMutable.test.js, so i put them i isoWeek.test.js, under startOf yearisoweek

again the name is crap..cant figure a better one atm.

this is my first time doing a pr request, let me know where i could improve, thanks!

hope this could help you guys xD

**Information**
 - Day.js Version [v1.10.4]
 - OS: [macOS]
 - Browser [nodejs]
 - Time zone: [UTC+08:00]

PR #1474, opened 2021-04-25
feat: dayjs().startOf(yearisoweek) #1473

rentianxiang:
Describe the bug
This is NOT a bug, it's just a feature i wish dayjs to have.

Expected behavior
So my project requires to translate between an ISO8601 date(year week weekday) and an actual Date object.
I found dayjs can translate Date to ISO8601 easily by using isoWeek and isoWeekday, but it does not support for the reverse action.
i tried with dayjs('2021', 'YYYY').isoWeek(1), but it returns the first isoweek of 2020 :(

So i added a new function under dayjs().startOf(), named "yearisoweek" (idk what it should be called..) to calculate the first isoWeek and isoWeekday for a given date (eg. dayjs('2021-04-25', 'YYYY-MM-DD).startOf('yearisoweek) == dayjs('2021-01-04, 'YYYY-MM-DD')), so it gives 2021W011. After it founds the first isoWeek and first isoWeekday of this year, it should be easy to calculate other ISO8601 dates by using the dayjs.add(week, 'week') and dayjs().add(day, 'day') functions.

In the test case, i found my method is not included in moment.js, it cannot be work with the startOf test in the badMutable.test.js, so i put them i isoWeek.test.js, under startOf yearisoweek

again the name is crap..cant figure a better one atm.

this is my first time doing a pr request, let me know where i could improve, thanks!

hope this could help you guys xD

Information

Day.js Version [v1.10.4]
OS: [macOS]
Browser [nodejs]
Time zone: [UTC+08:00]

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -173,6 +173,20 @@ class Dayjs {
         return instanceFactorySet(`${utcPad}Seconds`, 2)
       case C.S:
         return instanceFactorySet(`${utcPad}Milliseconds`, 3)
+      case 'yearisoweek':
+        if (isStartOf) {
+          const YD = instanceFactory(1, 0).$W
+          let diff
+          if (YD === 1) {
+            diff = 0
+          } else if (YD < 5) {
+            diff = (YD - 1) * -1
+          } else {
+            diff = 8 - YD
+          }
+          return instanceFactory(1, 0).add(diff, 'day')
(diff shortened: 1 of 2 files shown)


iamkun:
Maybe we could support this as a plugin?

Issue #1482, opened 2021-05-03
Wrong loading of plugins/locale breaks the app.


asvny:
**Describe the bug**
Loading plugins before and then loading the locale for `en` breaks the app.

**Expected behavior**
Should work correctly without any error.

**Information**
 - Day.js Version - 1.9.6
 - OS: Mac
 - Browser Chrome
 
 Let's say we have code like this
 ```
dayjs.extend(advancedFormat);
dayjs.locale(require('dayjs/locale/en'))

// Then when you try to do 
dayjs().format('dddd, MMM Do, YYYY');
 ```
It breaks with an error `d.ordinal` is not a function. It is because there is no `ordinal` property in `en.js` file. More specifically this function https://github.com/iamkun/dayjs/blob/dev/src/plugin/advancedFormat/index.js#L6 should be moved to `en.js` file. 

Happy to submit a PR.

asvny:
Created a fix for it https://github.com/iamkun/dayjs/pull/1483

ScoreUnder:
What is the correct order to load these in? I want to use the AdvancedFormat plugin to display day ordinals but I can't figure out how to avoid the `r.ordinal is not a function` bug.

rizowski:
We are only using the `AdvancedFormat` plugin and getting the `r.ordinal is not a function` error. Doesn't state in the documentation that it requires any other plugins for using `Do`

PR #1483, opened 2021-05-03
fix: move en ordinal function to locale file

asvny:
Fix for #1482

### DIFF
--- a/src/locale/en.js
+++ b/src/locale/en.js
@@ -3,5 +3,10 @@
 export default {
   name: 'en',
   weekdays: 'Sunday_Monday_Tuesday_Wednesday_Thursday_Friday_Saturday'.split('_'),
-  months: 'January_February_March_April_May_June_July_August_September_October_November_December'.split('_')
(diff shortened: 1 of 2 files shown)


iamkun:
This will make the core bundle size larger.

What I suggest to do is not to import local from en.js to main/index.js

asvny:
Interesting ... how these changes make core bundle larger ? one is locale and other is a plugin. 

`What I suggest to do is not to import local from en.js to main/index.js`

Not a good idea I reckon .. what if I use other features of en locale.

iamkun:
At present, the en's `ordinal` function is inside the plugin file, not the main bundle.

asvny:
Yeah .. wouldn't it make sense to have in locale file ? locale code inside a non closely related plugin feels weird.

How about in both the files ? might not break the app I guess

iamkun:
Yes, I'd like to have a full-size en locale file, but not imported to the main bundle.

Issue #1407, opened 2021-03-09
dayOfYear produces incorrect results 


amstr1k:
**Describe the bug**
dayOfYear produces incorrect results 

My config dayjs

```js
import dayjs from 'dayjs';
import ru from 'dayjs/locale/ru';
import customParseFormat from 'dayjs/plugin/customParseFormat';
import badMutable from 'dayjs/plugin/badMutable';
import quarterOfYear from 'dayjs/plugin/quarterOfYear';
import isSameOrBefore from 'dayjs/plugin/isSameOrBefore';
import isBetween from 'dayjs/plugin/isBetween';
import isSameOrAfter from 'dayjs/plugin/isSameOrAfter';
import dayOfYear from 'dayjs/plugin/dayOfYear';
import utc from 'dayjs/plugin/utc';
import relativeTime from 'dayjs/plugin/relativeTime';

dayjs.locale(ru);
dayjs.extend(customParseFormat);
dayjs.extend(badMutable);
dayjs.extend(quarterOfYear);
dayjs.extend(isSameOrBefore);
dayjs.extend(isBetween);
dayjs.extend(isSameOrAfter);
dayjs.extend(dayOfYear);
dayjs.extend(utc);
dayjs.extend(relativeTime);

export default dayjs;
```

Execution jest test
```bash
> 461 |         expect(dayjs(`09.04.2020`, `DD.MM.YYYY`).dayOfYear()).toEqual(100);
```

**Expected behavior**
Expected: 100
Received: 1

**Information**
 - Day.js Version 1.10.4
 - OS: Windows 10
 - Time zone: [e.g. GMT+04:00 DST (Moscow)]

PR #1511, opened 2021-05-30
fix: Fix DayOfYear plugin when using BadMutable plugin

seanyeh:
Fixes https://github.com/iamkun/dayjs/issues/1407
Very similar to a previous fix for WeekOfYear: https://github.com/iamkun/dayjs/pull/884

To reproduce:
```
var badMutable = require('dayjs/plugin/badMutable');
var dayOfYear = require('dayjs/plugin/dayOfYear');

dayjs.extend(badMutable);
dayjs.extend(dayOfYear);
dayjs('02.25.2020').dayOfYear(); // Should return 56, but returns 1 instead
```

### DIFF
--- a/src/plugin/dayOfYear/index.js
+++ b/src/plugin/dayOfYear/index.js
@@ -1,7 +1,8 @@
-export default (o, c) => {
+export default (o, c, d) => {
   const proto = c.prototype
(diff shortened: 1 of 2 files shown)


iamkun:
Thanks

PR #1512, opened 2021-05-31
test trigger Travis-CI

sevtdy:
fix: incorrect startOf/endOf behavior with +00:00 timezone

### DIFF
--- a/src/plugin/timezone/index.js
+++ b/src/plugin/timezone/index.js
@@ -114,17 +114,6 @@ export default (o, c, d) => {
     return result && result.value
   }
 
-  const oldStartOf = proto.startOf
-  proto.startOf = function (units, startOf) {
-    if (!this.$x || !this.$x.$timezone) {
-      return oldStartOf.call(this, units, startOf)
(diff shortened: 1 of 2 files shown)


sevtdy:
sorry for open this PR multi-time, look like there is some problem with Travis-CI and I'm trying to solve this with Travis support, this PR is a test for Travis support.

iamkun:
Thanks.

It is odd that the PR sent from your account could not trigger the CI. But the rest looks fine.

iamkun:
I run the `npm test` on my locale machine and it fails with some errors.

sevtdy:
I pass all the tests on my machine, can you share the error?

sevtdy:
finally... 😂

sevtdy:
@iamkun this PR has been open for a long time, just want to know `dayjs` considers this behavior normal or a bug?

iamkun:
According to your fix, is that mean, we should not update the startOf func in a timezone plugin, and the problem is solved?

PR #1524, opened 2021-06-06
Fallback to language only locale + support uppercase locales

codeability-ab:
Fallback to locale xx if xx-yy is provided and does not exist

Also support locales in uppercase such as en-GB

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -21,7 +21,7 @@ const parseLocale = (preset, object, isLocal) => {
       l = presetLower
     }
     const presetSplit = preset.split('-')
-    if (!l && presetSplit.length) {
+    if (!l && presetSplit.length > 1) {
       return parseLocale(presetSplit[0])
     }
   } else {
--- a/src/index.js
+++ b/src/index.js
@@ -12,12 +12,16 @@ const parseLocale = (preset, object, isLocal) => {
   let l
   if (!preset) return L
   if (typeof preset === 'string') {
-    if (Ls[preset]) {
-      l = preset
+    const presetLower = preset.toLowerCase()
+    if (Ls[presetLower]) {
+      l = presetLower
     }
     if (object) {
-      Ls[preset] = object
-      l = preset
+      Ls[presetLower] = object
+      l = presetLower
+    }
+    if (!l && preset.split('-').length > 1) {
+      return parseLocale(preset.split('-')[0])
     }
   } else {
     const { name } = preset
--- a/src/index.js
+++ b/src/index.js
@@ -20,8 +20,9 @@ const parseLocale = (preset, object, isLocal) => {
       Ls[presetLower] = object
       l = presetLower
     }
-    if (!l && preset.split('-').length > 1) {
-      return parseLocale(preset.split('-')[0])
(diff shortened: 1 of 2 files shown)


codeability-ab:
@iamkun will this pull request be considered for merge?

src/index.js
```diff
@@ -12,12 +12,16 @@ const parseLocale = (preset, object, isLocal) => {
...
+    if (Ls[presetLower]) {
+      l = presetLower
     }
     if (object) {
-      Ls[preset] = object
-      l = preset
+      Ls[presetLower] = object
+      l = presetLower
+    }
+    if (!l && preset.split('-').length > 1) {
```
iamkun, line 23: maybe we can simplify the logic to 
const presetSplit = preset.split('-')
if (!l && presetSplit.length) {
  return parseLocale(presetSplit[0])
}

codeability-ab: Yes. I'll fix it!

test/plugin/localizedFormat.test.js
```diff
@@ -96,6 +96,16 @@ it('Uses the localized lowercase formats if defined', () => {
   ['l', 'll', 'lll', 'llll'].forEach(option => expect(znDate.format(option)).toBe(znDate.format(znCn.formats[option])))
 })
 
+it('Uses fallback to xx if xx-yy not available', () => {
+  expect(dayjs('2019-02-01').locale('es-yy').format('L'))
```
iamkun, line 100: it will be more clear to use `format('MMMM')` to test if the locale is correct and not the default English

src/index.js
```diff
@@ -12,12 +12,17 @@ const parseLocale = (preset, object, isLocal) => {
...
+      l = presetLower
     }
     if (object) {
-      Ls[preset] = object
-      l = preset
+      Ls[presetLower] = object
+      l = presetLower
+    }
+    const presetSplit = preset.split('-')
+    if (!l && presetSplit.length > 1) {
```
iamkun, line 24: if (!l && presetSplit.length) {

we could just check the length is above 0

codeability-ab: That was my first attempt but other unit tests break if I don't check that the split actually splits the preset string (i.e it results in more than one array element).

The test "Not supported locale string fallback to previous one (instance)" in locale.test.js fails with "if (!l && presetSplit.length) {"

The "not_supported_locale_string" is split into one element => recursive call, "not_supported_locale_string" is split into one element again => recursive call ... etc

iamkun: oh. I see, you can just change the "not_supported_locale_string" test string 😁

codeability-ab: You mean like "not-supported-locale-string"?
But that string will never split into something valid.
I still think it's a good idea to only recurse if the preset actually splits into 2 parts.

hirbod:
I would love to see this PR merged. It is exactly fixing all the issues I've encountered on my React Native App, when trying to load the correct locale based on system preferences

codeability-ab:
@iamkun Can we merge this one please?

iamkun:
THX

PR #1529, opened 2021-06-11
fix: do not modify dayjs input arguments

kzahel:
I noticed that when passing in an array of date formats:

"If you don't know the exact format of an input string, but know it could be one of many, you can use an array of formats."
(from https://day.js.org/docs/en/parse/string-format) it was actually modifying the input array. This seems pretty bad.

I'm not sure if your build supports the ... syntax. It could be changed to Object.assign or similar if you want.

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -28,12 +28,16 @@ const parseLocale = (preset, object, isLocal) => {
   return l || (!isLocal && L)
 }
 
+function objClone(obj) {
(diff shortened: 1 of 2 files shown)


iamkun:
this is a nice catch, however, ... syntax will result in a large increase of the bundle size, maybe we should looking for some other way to fix this.

PR #1544, opened 2021-06-21
Fix esm plugin typings and add test.

mshima:
Generated esm plugins are generated as `plugin/duration/index.js`, but import still points to `import dayjs from '../index'`, but it should be `import dayjs from '../../index'`.

- Fix imports.
- Add tests.

### DIFF
--- a/build/esm.js
+++ b/build/esm.js
@@ -33,7 +33,9 @@ const localeTypePath = path.join(process.env.PWD, 'esm/locale', `index${typeFile
         const filePath = path.join(pluginDir, p)
         const targetPath = path.join(pluginDir, pluginName, `index${typeFileExt}`)
         const readFile = await promisify(fs.readFile)(filePath, 'utf8')
-        const result = readFile.replace(/'dayjs'/g, "'dayjs/esm'")
+        const result = readFile
+          .replace('"../index"', "'../../index'")
+          .replace(/'dayjs'/g, "'dayjs/esm'")
         await promisify(fs.writeFile)(targetPath, result, 'utf8')
         await promisify(fs.unlink)(filePath)
       }
--- a/type-test/esModuleInterop.ts
+++ b/type-test/esModuleInterop.ts
@@ -1,0 +1,11 @@
+import dayjs from 'dayjs/esm';
+
+import customParseFormat from 'dayjs/esm/plugin/customParseFormat';
+import duration from 'dayjs/esm/plugin/duration';
+import relativeTime from 'dayjs/esm/plugin/relativeTime';
+
+new dayjs.Dayjs();
+
+dayjs.extend(customParseFormat);
+dayjs.extend(duration);
+dayjs.extend(relativeTime);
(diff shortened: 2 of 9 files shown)


type-test/esModuleInterop.ts
```diff
@@ -0,0 +1,11 @@
+import dayjs from 'dayjs/esm';
```
iamkun, line 1: what's this file used for, please?

iamkun:
e.g. in file `esm/locale/af.js`,`import dayjs from '../index'` means to import `root/esm/index.js`, should correct don't we?

type-test/esModuleInterop.ts
mshima: Test esm generated files.

mshima:
> e.g. in file `esm/locale/af.js`,`import dayjs from '../index'` means to import `root/esm/index.js`, should correct don't we?

Yes, it’s just changing plugin directories.

iamkun:
> > e.g. in file `esm/locale/af.js`,`import dayjs from '../index'` means to import `root/esm/index.js`, should correct don't we?
> 
> Yes, it’s just changing plugin directories.
> I think locales folder should be ok.

seems we don't have to change the path here?

mshima:
> seems we don't have to change the path here?

Locale is ok, plug-ins generates a folder due to typings.

iamkun:
> > seems we don't have to change the path here?
> 
> Locale is ok, plug-ins generates a folder due to typings.

Can you please detail it? Seems there's no plugin has the import statement `import ../index.js`

mshima:
Honestly I don’t know why Babel adds it.
Look at `esm/plugins/*/index.js`.

Probably due to typings.
https://github.com/iamkun/dayjs/blob/8e32164855cff724642e24c37a631eb4c4d760c8/types/plugin/duration.d.ts#L2
Changing that to `dayjs` may fix the problem too I will test.

mshima:
The error is just with duration plugin.
Do you want me to drop the tests?

build/esm.js
```diff
@@ -16,7 +16,7 @@ const localeTypePath = path.join(process.env.PWD, 'esm/locale', `index${typeFile
     readLocaleDir.forEach(async (l) => {
       const filePath = path.join(localeDir, l)
       const readFile = await promisify(fs.readFile)(filePath, 'utf8')
-      const result = readFile.replace("'dayjs'", "'../index'")
+      const result = readFile.replace(/'dayjs'/g, "'dayjs/esm'")
```
iamkun, line 19: seems `../index` should be the same file as `dayjs/esm`,  don't it?

mshima: That’s for locale.
Plugins generated structure is different due to typings.
`translation/en.js` vs `plugins/duration/index.js`

mshima: Feel free to commit this change. But it’s not broken.

iamkun: You can check the final file here https://unpkg.com/browse/dayjs@1.10.6/esm/locale/af.js

image


`../index` refers to `esm/index.js`, which I think it's correct?

PR #1548, opened 2021-06-22
Improve romanian locale

weaseldotro:
add better wording for past relative time

### DIFF
--- a/src/locale/ro.js
+++ b/src/locale/ro.js
@@ -19,7 +19,7 @@ const locale = {
   },
   relativeTime: {
     future: 'peste %s',
-    past: 'acum %s',
+    past: 'în urmă cu %s',
     s: 'câteva secunde',
     m: 'un minut',
     mm: '%d minute',


src/locale/ro.js
```diff
@@ -19,7 +19,7 @@ const locale = {
   },
   relativeTime: {
     future: 'peste %s',
-    past: 'acum %s',
+    past: 'în urmă cu %s',
```
iamkun, line 22: can we use `%s în urmă`

I see this on https://github.com/moment/moment/blob/develop/src/locale/ro.js#L55

weaseldotro: No, that's worse. My version is the correct one.
I'll make a pull request to the moment library as well.

iamkun: Thanks, I'll mark this PR for more code review by native speakers before it gets merged.

weaseldotro: Good plan 👍

PR #1549, opened 2021-06-23
fix: parameter type

imwh0im:
change parameter type of plugin `isBetween`

### DIFF
--- a/types/plugin/isBetween.d.ts
+++ b/types/plugin/isBetween.d.ts
@@ -5,6 +5,6 @@ export = plugin
 
 declare module 'dayjs' {
   interface Dayjs {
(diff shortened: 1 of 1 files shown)


types/plugin/isBetween.d.ts
```diff
@@ -5,6 +5,6 @@ export = plugin
 
 declare module 'dayjs' {
   interface Dayjs {
-    isBetween(a: ConfigType, b: ConfigType, c?: OpUnitType | null, d?: string): boolean
+    isBetween(a: ConfigType, b: ConfigType, c?: OpUnitType | null, d?: '()' | '[]' | '[)'): boolean
```
iamkun, line 8: also `(]` ?

imwh0im: Thx 👍

Issue #1554, opened 2021-06-28
more strict String Parser regex


FourwingsY:
### Describe the bug
from `REGEX_PARSE = /^(\d{4})[-/]?(\d{1,2})?[-/]?(\d{0,2})[^0-9]*(\d{1,2})?:?(\d{1,2})?:?(\d{1,2})?[.:]?(\d+)?$/`

when `new Date('2020.05.23')` is parsed in chrome`2020-05-23T00:00:00.000`, and Invalid in safari.
but `dayjs('2020.05.23')` is parsing as `2020-01-01T05:00:00.023`

### Expected behavior
`dayjs('2020.05.23')` should fallback to native Date

### Information
I think the problem is that `[^0-9]*` accepts any characters including `.`
what commonly used cases are `[Tt\s]*`, and it seems okay to me.

iamkun:
would you like to submit a PR with unit test to fix this?

FourwingsY:
I have found that dots are not a valid rule from ISO8601 or things, but it just worked in Chrome only.
So It means, It should not matched in REGEX, and fallback to `Date`

FourwingsY:
Fix merged

PR #1555, opened 2021-06-28
fix: more strict delimiter in REGEX_PARSE

FourwingsY:
resolving #1554

### DIFF
--- a/src/constant.js
+++ b/src/constant.js
@@ -26,5 +26,5 @@ export const FORMAT_DEFAULT = 'YYYY-MM-DDTHH:mm:ssZ'
 export const INVALID_DATE_STRING = 'Invalid Date'
 
 // regex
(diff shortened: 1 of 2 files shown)


src/constant.js
```diff
@@ -26,5 +26,5 @@ export const FORMAT_DEFAULT = 'YYYY-MM-DDTHH:mm:ssZ'
 export const INVALID_DATE_STRING = 'Invalid Date'
 
 // regex
-export const REGEX_PARSE = /^(\d{4})[-/]?(\d{1,2})?[-/]?(\d{0,2})[^0-9]*(\d{1,2})?:?(\d{1,2})?:?(\d{1,2})?[.:]?(\d+)?$/
+export const REGEX_PARSE = /^(\d{4})[-/]?(\d{1,2})?[-/]?(\d{0,2})[Tt\s]*(\d{1,2})?:?(\d{1,2})?:?(\d{1,2})?[.:]?(\d+)?$/
```
iamkun, line 29: why a `Tt` here?

FourwingsY: is T is enough? I searched some regex for ISO8601 format and they were using Tt instead of T. I guess thats for some compatibility of some OS or libraries.

iamkun: I see

Issue #1558, opened 2021-06-30
format throws when used on "Invalid Date" with advancedformat + timezone + utc


SamuelPoulin:
**Describe the bug**
When using the following format string on dayjs(null) the resulting date is "Invalid Date"
```
lll
```

When using the following string, dayjs throws
```
lll z
```
Thrown error:
```
Unhandled Runtime Error

RangeError: date value is not finite in DateTimeFormat.formatToParts()

dayjs(null).format('lll z')
           ^
```

**Expected behavior**
dayjs should display "Invalid Date" and not throw an error

**Information**
 - Day.js Version v1.10.5
 - OS: Windows
 - Browser Firefox 89.0.2 (64-bit)
 - Time zone: EST

imwh0im:
In typescript?

imwh0im:
In Javascript
```javascript
var dayjs = require("dayjs");

console.log(dayjs(null).format('lll z'));  // Invalid Date
```

In Typescript
```typescript
import dayjs from 'dayjs'

console.log(dayjs(null).format('lll z'));
```
returned
```bash
error TS2769: No overload matches this call.
```

i tried reproduced this issue. but not reproduced.
but find returned Error in Typescript. fix it. 

give more details for this issue?

SamuelPoulin:
My bad, I should have mentionned that the `z` format option is only enabled with `advancedFormat`. Here's a working example of the issue:

**index.js**
```javascript
const dayjs = require("dayjs");

const utc = require("dayjs/plugin/utc");
const timezone = require("dayjs/plugin/timezone");
const advancedFormat = require("dayjs/plugin/advancedFormat");

dayjs.extend(utc)
dayjs.extend(timezone)
// dayjs.extend(advancedFormat) // When this is enabled, dayjs throws. When it's not, it outputs 'Invalid Date'.

console.log(dayjs(null).format('lll z'))
```

imwh0im:
i checked plugin `advancedFormat`. 
https://github.com/iamkun/dayjs/blob/dev/src/plugin/advancedFormat/index.js#L12-L51

if you want use the original `format()`. Do not extend `advancedFormat`.
Because it changes the function of `format()` with the plugin advancedFormat.
if you extends `advancedFormat`. You must to follow the rules.

https://day.js.org/docs/en/plugin/advanced-format check this document.

SamuelPoulin:
[...]

PR #1560, opened 2021-07-01
fix: dayjs ConfigTypeMap

imwh0im:
- #1558
- fix dayjs ConfigTypeMap for Typescript
- add test case in `parse.test.ts`


imwh0im:
has not depended on #1558 .
if run `dayjs(null)` In typescript. returned Error `No overload matches this call.`.
https://github.com/iamkun/dayjs/issues/1558#issuecomment-872693646

iamkun:
THX

PR #1565, opened 2021-07-05
fix: advanced format of invalid date

imwh0im:
- #1558 
- add validation for invalid date in plugin `advancedFormat`


imwh0im:
CI verification failed, so i plan to do PR again after correct it.

PR #1566, opened 2021-07-06
fix: add plugin advancedFormat validation

imwh0im:
- #1558 
- add validation for invalid date in plugin advancedFormat


src/plugin/advancedFormat/index.js
```diff
@@ -11,6 +11,11 @@ export default (o, c, d) => { // locale needed later
   // extend en locale here
   proto.format = function (formatStr) {
     const locale = this.$locale()
+
```
iamkun, line 14: if (!this.isValid()) {
      return oldFormat.bind(this)(formatStr)
    }

imwh0im: thx. i've learned from this comment

PR #1572, opened 2021-07-09
Add format object type to index.d.ts

always-maap:
Hi. Dayjs is a super flexible library. I love working with it, writing plugins, and contributing to it. We can change anything with plugins but the typing part does not come along well enough. ``Using interface and not using type aliases`` will improve typing for plugin authors so much better. I changed one of the more important ones in this PR.
```
dayjs(new Date(), { multipleLocales: [] })
```
or any other example. the point is more flexibility in typing for plugin authors. 
**if there is a way to solve this problem currently or any planning for changing the whole typing system(v2 doesn't hit that), let me know <333**

### DIFF
--- a/types/index.d.ts
+++ b/types/index.d.ts
@@ -15,7 +15,9 @@ declare namespace dayjs {
 
   export type ConfigType = ConfigTypeMap[keyof ConfigTypeMap]
 
-  export type OptionType = { locale?: string, format?: string, utc?: boolean } | string | string[]
+  export interface FormatObject { locale?: string, format?: string, utc?: boolean }
+
+  export type OptionType = FormatObject | string | string[]
 
   export type UnitTypeShort = 'd' | 'M' | 'y' | 'h' | 'm' | 's' | 'ms'
 


iamkun:
actually, `dayjs(data, config)` is not a recommend API in our docs. It increases the complexity of the constructor to our end user.

always-maap:
> actually, `dayjs(data, config)` is not a recommend API in our docs. It increases the complexity of the constructor to our end user.

By not being recommended you mean even the third-party library that has their own small api(I mean will never be part of official plugins)?
I think using those two concepts in declaration files is still a good idea. Also, I'm not using it for all configs. only configs that effects output directly. e.g more properties in output. Totally agree with you on this page. sorry for kinda broken en <333

addisonElliott:
> actually, `dayjs(data, config)` is not a recommend API in our docs. It increases the complexity of the constructor to our end user.

If I understand you correctly, you're saying the recommended approach by Day.js is to have something like `dayjs(data).tz(xxx)` over `dayjs(data, { tz: xxx }`. I see where you're coming from.

However, I don't think this means we should restrict 3rd party plugins to using this approach. If someone has a 3rd party plugin, I think it's up to them if they want to add an attribute to the config object or create a separate function to set the attribute. Plus, adding an attribute to the config object is already possible, rather the issue is the Typescript definitions and being able to augment the interface.

So, given that, I think this PR is beneficial to people working on customizing Day.js 👍

always-maap:
> > actually, `dayjs(data, config)` is not a recommend API in our docs. It increases the complexity of the constructor to our end user.
> 
> If I understand you correctly, you're saying the recommended approach by Day.js is to have something like `dayjs(data).tz(xxx)` over `dayjs(data, { tz: xxx }`. I see where you're coming from.
> 
> However, I don't think this means we should restrict 3rd party plugins to using this approach. If someone has a 3rd party plugin, I think it's up to them if they want to add an attribute to the config object or create a separate function to set the attribute. Plus, adding an attribute to the config object is already possible, rather the issue is the Typescript definitions and being able to augment the interface.
> 
> So, given that, I think this PR is beneficial to people working on customizing Day.js 👍

Sure I agree with you on the second part. but for the first example, It's on the author's choice of which way. I just recommend being more flexible on typing for 3rd parties while you can basically do anything with js prototypes. I already moved the config and solved my problem but overall I think it's beneficial.

iamkun:
Agreed, cool

PR #1576, opened 2021-07-13
fix: narrow type for `add` and `subtract`

### LINKED ISSUE #1575: Not all unit in `OpUnitType` is legal unit for `add` and `subtract`
**Describe the bug**
`OpUnitType` contain some 'illegal' unit for `add` and `subtract`. According to [doc](https://dayjs.gitee.io/docs/zh-CN/manipulate/add#%E6%94%AF%E6%8C%81%E7%9A%84%E5%8D%95%E4%BD%8D%E5%88%97%E8%A1%A8).
I was misled by `OpUnitType`, thought `date` can be applied to `subtract`, but result was unexpected. Value wasn't changed. Took me sometime to figure it out.

**Expected behavior**
Correct TS params type for `add` and `subtract` unit.

**Information**
 - Day.js latest
 - OS: Mac0S
 - Browser latest Chrome

susiwen8:
Close #1575

### DIFF
--- a/types/index.d.ts
+++ b/types/index.d.ts
@@ -29,7 +29,7 @@ declare namespace dayjs {
 
   export type OpUnitType = UnitType | "week" | "weeks" | 'w';
   export type QUnitType = UnitType | "quarter" | "quarters" | 'Q';
-  export type ManipulateType = UnitTypeShort | Omit<UnitTypeLong, 'date'> | 'w' | 'week' | 'quarter' | 'Q';
+  export type ManipulateType = Omit<OpUnitType, 'date' | 'dates'>;
   class Dayjs {
     constructor (config?: ConfigType)
     /**
--- a/types/index.d.ts
+++ b/types/index.d.ts
@@ -29,7 +29,7 @@ declare namespace dayjs {
 
   export type OpUnitType = UnitType | "week" | "weeks" | 'w';
   export type QUnitType = UnitType | "quarter" | "quarters" | 'Q';
-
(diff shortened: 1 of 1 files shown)


addisonElliott:
Your fix removes the `UnitTypeLongPlural` type. So any pluralized types would return errors now. Looking at the code, it also applies to `diff()` function. Not only that, this doesn't remove the `dates` plural type.

However, I wonder if it would be better to add support for `date` in add/subtract/diff. It would be relatively easy and make it consistent with `startOf`, `endOf` functions.

FYI, if you use `dayjs.add(1, 'date')`, it will add 1ms to the date. See here: https://github.com/iamkun/dayjs/blob/dev/src/index.js#L243

susiwen8:
@addisonElliott Thanks,

> So any pluralized types would return errors now.

 Indeed pluralized type should be included. 

>  if you use dayjs.add(1, 'date'), it will add 1ms to the date
You are right, I tried myself, `date` act like `ms`

> I wonder if it would be better to add support for date in add/subtract/diff
Agree, I was going to fix this way. But I thought this might be intentional or someting that left `date` behind

iamkun:
Yes, `date` should not be supported in `dayjs.add`, cause add a date make no sense

susiwen8:
@iamkun  should I reopen this PR?

iamkun:
why not 😁

susiwen8:
@iamkun please review 😁

iamkun:
BTW, is any special reason to close it?

susiwen8:
Just thought you might not interested in this pr.

iamkun:
@addisonElliott according to moment.js  date is not support as well.

let's just omit it 

https://momentjs.com/docs/#/manipulating/add/

image

PR #1578, opened 2021-07-14
fix: isBetween for invalid date

yanaemon:
## Describe the bug

On isBetween plugin with invalid date, inclusive returns true, but exclusive returns false.
```
> dayjs('2021-01-01').isBetween(null, null, 'date', '[]')
true
> dayjs('2021-01-01').isBetween(null, null, 'date', '()')
false
```

## Expected behavior

This behavior is very strange compared to moments (return false if either argument is invalid date).
```
> moment('2021-01-01').isBetween(null, null, 'date', '[]')
false
> moment('2021-01-01').isBetween(null, null, 'date', '()')
false
> moment('2021-01-01').isBetween('2020-01-01', null, 'date', '[]')
false
> moment('2021-01-01').isBetween(null, '2022-01-01', 'date', '[]')
false
```

This is because `!d.isBefore`/`!d.isAfter` for invalid date always returns true.
I think it would be better to make this behavior the same as moment.js.

### DIFF
--- a/src/plugin/isBetween/index.js
+++ b/src/plugin/isBetween/index.js
@@ -2,6 +2,9 @@ export default (o, c, d) => {
   c.prototype.isBetween = function (a, b, u, i) {
     const dA = d(a)
     const dB = d(b)
+    if (!dA.isValid() || !dB.isValid()) {
+      return false
+    }
     i = i || '()'
     const dAi = i[0] === '('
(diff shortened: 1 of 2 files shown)


keidarcy:
@iamkun
What about the status of this PR.
Looks like a pretty necessary feature :)

Issue #1537, opened 2021-06-17
[Feature Request] - Date Ranges Overlap


LaravDev:
It would be really useful if we could compare 2 date ranges against each other to know if there is an overlap, just as in the "isBetween" it would be good to have delimiters so we can limit end dates, or start dates, this could possibly be an update for that IsBetween plugin.

As an example you can take a look on this package (that is for moment.js):

https://github.com/rotaready/moment-range#overlaps

binvb:
+1

LaravDev:
For a manual workaround in the meanwhile I ended up by creating the next function:

```
export const rangesOverlaps = (localRange:Array<any>, remoteRange:Array<any>) => {
  const remoteDates:Array<any> = [dayjs(remoteRange[0])]
  const remoteDatesDiff:number = dayjs(remoteRange[1]).diff(dayjs(remoteRange[0]), "day")

  for (let i = 0; i < remoteDatesDiff; i++){
    remoteDates.push(dayjs(remoteDates[remoteDates.length - 1]).add(1, 'day'))
  }

  remoteDates.pop()

  return remoteDates.some(x => x.isBetween(dayjs(localRange[0]), dayjs(localRange[1]), 'day', '[)'))
}
```



```
if (rangesOverlaps([dateRange.start, dateRange.end], [data.checkinDate.toDate(), data.checkoutDate.toDate()])){
          occupiedRooms.value.push(parseInt(data.roomNumber), ...data.extra_rooms)
 }
```

Hopefully this can help anyone else, obviously in this example I did it for my use case, but this can be adapted to any use case i think.

alex-w0:
@iamkun Any plans to merge the pr?

AlexGalays:
Needed this just now! This is quite a bit more useful than things like isSameOrBefore one liners :trollface:

katerlouis:
Also wondering if the pr is being merged :)

katerlouis:
Also wondering if the pr is being merged :)

Issue #1162, opened 2020-10-22
How to generate to date ranges?


engineer-myoa:
Currently, does dayjs support generating date ranges?
Like this one. [moment-range](https://github.com/rotaready/moment-range) extension.

e.g.) 
```javascript
dayjs.ranges(
    dayjs("2020-10-10"),
    dayjs("2020-10-20"),
    "days"
)
// the result is [start, end]
// like [10, 11, 12, 13, 14, 15, 16, 17, 18, 19, 20]
// or [start, end)
// like [10, 11, 12, 13, 14, 15, 16, 17, 18, 19]
```

Please check if dayjs team can considered it.

iamkun:
Looks cool, we could add it as a plugin.

smblee:
Definitely cool idea. This is currently what I do in my application.

```
  let currentDate = dayjs(openTime);

  while (currentDate.isBefore(closeTime) || currentDate.isSame(closeTime)) {
    currentDate = currentDate.add(amount, unit);
  }
```

engineer-myoa:
> Definitely cool idea. This is currently what I do in my application.
> 
> ```
>   let currentDate = dayjs(openTime);
> 
>   while (currentDate.isBefore(closeTime) || currentDate.isSame(closeTime)) {
>     currentDate = currentDate.add(amount, unit);
>     ranges.push(currentDate.format());
>   }
> ```

It looks short, simple and powerful.
But if it become to some plugin, It need to have fool proof.
Is it the amount value mean interval?
Maybe the amount parameter can be adjust it. or swap between value if it had reverse order.

```javascript
startOfRange = dayjs.max(openTime, closeTime);
endOfRange dayjs.min(openTime, closeTime);

let currentDate = dayjs(startOfRange);

while (currentDate.isBefore(endOfRange) || currentDate.isSame(endOfRange)) {
    currentDate = currentDate.add(amount, unit);
    ranges.push(currentDate.format());
  }
```

smblee:
Yea theres a lot you can do with this (moment-range has a LOT more features).

amount and unit are the standard amount/unit (e.g. 30, 'minutes') used in the application. I think this code alone will get us started.

Envision contract to be something like:
`dayjs.range(start, finish, amount, unit, { ...other options })`

`dayjs.range(dayjs(), dayjs().add(1, 'day'), 15, 'minutes')` => [...]

my example code above will be converted to

[...]

PR #1580, opened 2021-07-16
feat: add plugin range

imwh0im:
- #1537


LaravDev:
Please merge this! and thanks a lot in advance.

LaravDev:
Seems like this is ready to be merged... 🙏🙏🙏

coatesap:
Would it be worth naming this 'Period' rather than 'Range'? I'm aware that Moment plugin was called 'range', but usually in programming languages, 'range' is used to generate a sequence of things, whereas this feature is about creating a date 'period'.

The Moment plugin covers the term 'range' by also being iterable, which this could implement. 

I raise this, as #1162 asked for the ability to generate date ranges, and I've got a plugin in the works that would be called via `dayjs.range(start, end, unit, options)`, so I wouldn't want to end up with a naming clash.

src/plugin/range/index.js
```diff
@@ -0,0 +1,65 @@
...
+      s = end
+      e = start
+    }
+
+    this.start = s
+    this.end = e
+  }
+
+  isRange() {
+    return this.start.isValid() && this.end.isValid() && this instanceof DayjsRange
```
AlexGalays, line 17: what's the point of having this as a class method that checks whether the current instance is a range? The answer will always be yes. Also I don't think isRange should report false if one of the dates is invalid, it's still a range. Perhaps an isValid() method would be more useful.

imwh0im: understood this comment. i will change to `isValidRange`.
and validations `this.start`, `this.end`

i guess "end date is after the start state." has already been confirmed in 'constructor'.
need more?

src/plugin/range/index.js
```diff
@@ -0,0 +1,65 @@
+
+export class DayjsRange {
+  constructor(start, end) {
+    let s = start
+    let e = end
+
+    if (s > e) {
```
alex-w0, line 7: Small improvement to do this as a one liner
```suggestion
const { s, e } = start > end ? { s: end, e: start } : { s: start, e: end }
```

alex-w0 (changes requested):

src/plugin/range/index.js
AlexGalays: no you're right, I missed the constructor bit!

imwh0im: Reflected you are feedback. 
Try check again please :pray:

G-11-P:
Passed

types/plugin/range.d.ts
```diff
@@ -0,0 +1,25 @@
...
+declare module 'dayjs' { 
+    class DayjsRange {
+      constructor(startDate?: ConfigType, endDate?: ConfigType)
+
+      startDate: Dayjs
+      endDate: Dayjs
+
+      isRange(): boolean
+      clone(): DayjsRange
+      isOverLap(dateRange: DayjsRange): boolean
```
hmnd, line 16:
```suggestion
      isOverlap(dateRange: DayjsRange): boolean
```

The capital L tripped me up for a bit trying to get the plugin working locally.

types/plugin/range.d.ts
```diff
@@ -0,0 +1,25 @@
...
+export = plugin
+
+declare module 'dayjs' { 
+    class DayjsRange {
+      constructor(startDate?: ConfigType, endDate?: ConfigType)
+
+      startDate: Dayjs
+      endDate: Dayjs
+
+      isRange(): boolean
```
hmnd, line 14:
```suggestion
      isValidRange(): boolean
```

src/plugin/range/index.js
```diff
@@ -0,0 +1,61 @@
...
+
+    if (
+      (this.startDate >= other.startDate && this.startDate <= other.endDate) ||
+      (this.endDate >= other.startDate && this.endDate <= other.endDate) ||
+      (this.startDate <= other.startDate && this.endDate >= other.endDate)
+    ) {
+      return true
+    }
+
+    return false
```
hmnd, line 33: Can be simplified into a single return statement

```suggestion
    return (
      (this.startDate >= other.startDate && this.startDate <= other.endDate) ||
      (this.endDate >= other.startDate && this.endDate <= other.endDate) ||
      (this.startDate <= other.startDate && this.endDate >= other.endDate)
    )
```

hmnd (changes requested):

imwh0im:
@hmnd Thanks your suggested

imwh0im:
https://github.com/imwh0im/dayjs-range-extend

I've awaited #1580, and I can't wait no more.
So I've decided to implement the range function by myself.

range function module name is [dayjs-range-extend](https://github.com/imwh0im/dayjs-range-extend), this Module make below things possible. 
- Support date range
- Validation date range
- Support TypeScript
- Other functions.. 

i've mentioned, i couldn't wait anymore range function. However, want use range function from Day.js.
Unless there is a separate plan for the date range function, [dayjs-range-extend](https://github.com/imwh0im/dayjs-range-extend) is would be left.

src/plugin/range/index.js
```diff
@@ -0,0 +1,57 @@
...
+
+  isValidRange() {
+    return this.startDate.isValid() && this.endDate.isValid()
+  }
+
+  clone() {
+    return new DayjsRange(this.startDate, this.endDate)
+  }
+
+  isOverlap(other) {
```
katerlouis, line 20: how about an option for `isOverlap` to exclude equals?

A from: 16:00
A to: 17:00

compared to 
B from: 17:00
B to: 18:00 

shouldn't return `true` in my opinion when dealing with calendar entries
compared

src/plugin/range/index.js
```diff
@@ -0,0 +1,57 @@
...
+    if (!(other instanceof DayjsRange) || !(other.isValidRange())) {
+      return false
+    }
+
+    return (
+      (this.startDate >= other.startDate && this.startDate <= other.endDate) ||
```
katerlouis, line 26: how about an option for `isOverlap` to exclude equals?

A from: 16:00
A to: 17:00

compared to 
B from: 17:00
B to: 18:00 

shouldn't return `true` in my opinion when dealing with calendar entries

imwh0im: thx. I agreed.

katerlouis:
@iamkun still no plans to merge this? :')

nofacez:
Looking forward for this feature, so needed!

eelior:
Is this project still maintained @iamkun ?
Would love to see this PR merged 🙏

oldschoolbg:
In the absence of this being merged I'm going to have to use @imwh0im plugin... I also second this as being really important to merge if at all possible - it's a really useful tool to have as I am just migrating away from Moment (finally) :)

arielkuzminski:
What's going on with this PR?

jBernavaPrah:
Any news?

ceefour:
Agree with others... Please merge

YokiToki:
```
dayjs.ranges(
    dayjs("2021-07-16"),
    dayjs("2023-07-24"),
    "days"
)
```

vickyRathee:
When can we expect this to be merged?

maddada:
Still need this feature.. please merge it or implement something like it.

mreduar:
It seems that dayjs is no longer a good alternative to moment

maddada:
Went with date-fns (intervals methods) and it's worked perfectly for me.

Issue #1360, opened 2021-01-27
Breaking changes in 1.10.4 


pycka:
**Describe the bug**
#1338 introduced breaking change by removing DurationInputType and DurationAddType types from plugin namespace.
(https://github.com/iamkun/dayjs/commit/4aca4b1b584a15de1146d929f95c944594032f20#diff-ca2786d2c104006cc51bdc06948ed0fad7843dba6f8ca1570eaedea389c9da88L8)

**Expected behavior**
No breaking changes within patch release (if following semantic versioning). Any plans to revert this change?

**Information**
 - Day.js Version [e.g. v1.10.4]
 - Browser [e.g. chrome 62]
 - Time zone: [e.g. GMT-07:00 DST (Pacific Daylight Time)]

iamkun:
cc @zardoy can we still keep the same name?

It will be a fix in the next patch version to make sure there's no BC in a patch release.

zardoy:
@pycka Why should we revert this change? We didn't update the major version because we didn't change any API, we just made types more strict. For example, maintainers of popular library, called Prisma never update the major version, when they change the type names, that aren't described in docs: https://github.com/prisma/prisma/releases/tag/2.15.0#Breaking%20Changes . @pycka Can I know where you use `DuraionInputType`?? It **was not** strict and could accept **any** kind of object: `dayjs.duration({ somethingBad: "49" });` — not ok. I fully agree with you, that we should state this changes in a more clear way (as Prisma do). Also I would recommend to obtain the type of first argument in this way: `Parameters<typeof dayjs.duration>[0]` it will **never** break until BC in API is present.

zardoy:
Feel free to ask any questions!

pycka:
Thanks @iamkun and @zardoy for responding!
Pulling this update broke our CI builds which in itself should be enough to call it a breaking change :) These types were publicly available before so we just used them for convenience when working with Durations. There was nothing in the source code that would imply that these are implementation details, especially considering another public interface Duration used to consume them (expressis verbis).
I don't know Prisma, nor I spend too much with the docs, so it's hard for me to relate to that. From TypeScript point of view these changes weren't backward compatible.

iamkun:
maybe because our user could use 
```
Duration. DurationInputType
```
?

zardoy:
@pycka I have contacted Prisma team and I realised that they **don't** use SemVer.

> From TypeScript point of view these changes weren't backward compatible.

[...]

PR #1581, opened 2021-07-17
Fix dayjs/esm type definitions

BePo65:
Make the code in the 'dayjs/esm' folder published on npmjs usable with typescript.

This pr should kind of include the modifications from pr #751 (what still remains true is the [comment](https://github.com/iamkun/dayjs/pull/751#issuecomment-657888852) of @sullvn, but pr  #751 does not solve this major rework of dayjs anyway).

This pr should be an answer to the problems described in issue #1281.

**What my pr here does**

Update the type definitions in the 'dayjs/esm' folder of the code published to npmjs, so that dayjs can be used with plain typescript and esm and without namespace imports (`import * as x`) and without any settings like `esModuleInterop` or `allowSyntheticDefaultImports`. A working sample project based on this pr and with tests for all packages can be found on [github](https://github.com/BePo65/dayjs-esm-v2-demo).

The pr contains a small modification to the 'esm.js' build script:

I used `__dirname + '../'` as replacement for `process.env.PWD` to be consistent with the 'index.js' build script (and to make it possible to run this script on my windows machine).

**What remains to be done**

I did not find the script to update package.json during publishing, so I have to stay with a placeholder file in 'dayjs/esm'. I suppose that in this script the version tag in package.json is updated.

Part of the solution of the problem with using dayjs as an esm package was adding a copy of the package.json to the folder 'dayjs/esm' with `type=module` and `main=index.js` in it. The rest of the content should come from the package.json in 'dayjs'. This is something that still has to be done and is not part of this pull request.

### DIFF
--- a/build/esm.js
+++ b/build/esm.js
@@ -5,10 +5,7 @@ const { ncp } = require('ncp')
 
 const { promisify } = util
 
-const typeFileExt = '.d.ts'
-const localeDir = path.join(process.env.PWD, 'esm/locale')
-const pluginDir = path.join(process.env.PWD, 'esm/plugin')
-const localeTypePath = path.join(process.env.PWD, 'esm/locale', `index${typeFileExt}`);
+const localeDir = path.join(__dirname, '../esm/locale');
 
 (async () => {
   try {
@@ -20,24 +17,7 @@ const localeTypePath = path.join(process.env.PWD, 'esm/locale', `index${typeFile
       await promisify(fs.writeFile)(filePath, result, 'utf8')
     })
 
-    await promisify(ncp)('./types/', './esm')
-
-    const readLocaleFile = await promisify(fs.readFile)(localeTypePath, 'utf8')
-    const localResult = readLocaleFile.replace("'dayjs", "'dayjs/esm")
-    await promisify(fs.writeFile)(localeTypePath, localResult, 'utf8')
-
-    const readPluginDir = await promisify(fs.readdir)(pluginDir)
-    readPluginDir.forEach(async (p) => {
-      if (p.includes(typeFileExt)) {
-        const pluginName = p.replace(typeFileExt, '')
-        const filePath = path.join(pluginDir, p)
-        const targetPath = path.join(pluginDir, pluginName, `index${typeFileExt}`)
-        const readFile = await promisify(fs.readFile)(filePath, 'utf8')
-        const result = readFile.replace(/'dayjs'/g, "'dayjs/esm'")
-        await promisify(fs.writeFile)(targetPath, result, 'utf8')
-        await promisify(fs.unlink)(filePath)
-      }
-    })
+    await promisify(ncp)('./types/esm/', './esm')
   } catch (e) {
     console.error(e) // eslint-disable-line no-console
   }
--- a/build/index.js
+++ b/build/index.js
@@ -14,6 +14,8 @@ const promisifyWriteFile = promisify(fs.writeFile)
 const localeNameRegex = /\/\/ (.*) \[/
 const formatName = n => n.replace(/\.js/, '').replace('-', '_')
 
+const filterEsmTypes = source => !(fs.lstatSync(source).isDirectory() && source.endsWith(`${path.sep}esm`))
+
 const localePath = path.join(__dirname, '../src/locale')
 
 async function build(option) {
@@ -62,7 +64,7 @@ async function listLocaleJson(localeArr) {
       fileName: './dayjs.min.js'
     }))
 
-    await promisify(ncp)('./types/', './')
+    await promisify(ncp)('./types/', './', { filter: filterEsmTypes })
 
     // list locales
     await listLocaleJson(locales)
--- a/types/esm/constant.d.ts
+++ b/types/esm/constant.d.ts
@@ -1,0 +1,25 @@
+export var SECONDS_A_MINUTE: number
+export var SECONDS_A_HOUR: number
+export var SECONDS_A_DAY: number
+export var SECONDS_A_WEEK: number
+export var MILLISECONDS_A_SECOND: number
+export var MILLISECONDS_A_MINUTE: number
+export var MILLISECONDS_A_HOUR: number
+export var MILLISECONDS_A_DAY: number
+export var MILLISECONDS_A_WEEK: number
+
+export var MS: string
+export var S: string
+export var MIN: string
+export var H: string
+export var D: string
+export var W: string
+export var M: string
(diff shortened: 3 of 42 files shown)


BePo65:
Just a few additional remarks about this pull request for using 'dayjs/esm' with typescript:

### What we want

1. a default export to access the factory functions, the methods and the types
2. named exports for the methods and the types to be used in destructoring imports

### The problems

- for (1) we use overloading: we need 3 elements (the name doesn't matter, but 'dayjs' is used here)
  - 2 overloaded versions of the factory function and
  - a namespace with the methods and types of dayjs.
- for (2) we need named exports of the methods and types

As I didn't find a way to access types defined in a namespace via destructuring imports, I moved the types out of the namespace dayjs. This means that the types can only accessed by destructuring imports (e.g. `import { <name-of-type> } from 'dayjs/esm'`), not via the default import (`import dayjs from 'dayjs.esm'`).

The class 'Dayjs' is part of the 'dayjs' namespace.


### Plugin topics

All types are accessible via the module 'dayjs/esm' as named exports (i.e. via a destructuring import)

- duration (DurationAddType, DurationInputType);
  - according to issue #1360 to use the (deprecated) types we have to use destructuring imports;
  - the default import does not return the types for destructuring
  - if we wrap the type definitions in a namespace (required to acccess the types via the default import) , we cannot use destructuring imports, only the default import.
  - ==> I added the types to the 'dayjs/esm' module as named export (I did that for all types to be consistent with the dayjs type definitions).
- localeData: types go to the module 'dayjs/esm'; interfaces and methods go to the namespace 'dayjs' (to make them accessible via the default import)

iamkun:
Is it possible to re-use the type files under `/type` instead of copying them?

BePo65:
I saw that you modified the "original" type definitions in `esm.js`. Basically a great idea. For some plugins that would still be possible.

The problem arises for all type definitions that contain 'types' and 'namespaces' (e.g. `index.d.ts` ind the base folder or `duration.d.ts` in the plugin folder). Here the necessary modifications are so extensive that IMHO it cannot be done by code.

Unluckily it concerns more than a few plugins: arraySupport, calendar, customParseFormat, dayOfYear, duration, isBetween, .....  
In that definition files, after any `declare module 'dayjs' {` we would have to insert `namespace dayjs {`.

And then we have plugins with more modifications necessary (e.g. `duration` with its deprecated type defnitions).

I did not try to transfer the modifications back to the `dayjs`, but I am afraid that would break most applications.

Sorry for that; I am absolutely no fan of code duplication, but I didn't find a better solution.

iamkun:
I see.

As a matter of this,  we might publish another npm package `dayjs-esm` or something alike? 

Cause it has a lot different with out current umd dayjs

To make it clear.

BePo65:
OK, I see: **1 source** (in github) **2 targets** (in npm).

As we need 2 package.json files anyway, that looks like a good choice to me. Unluckily there is already a `dayjs-esm` on npm (without any information about author. source etc.). I hope it's your package :-)

iamkun:
Yes, I took it. 

However, it's still hard to manage two packages via the same GitHub repo source.

BePo65:
> However, it's still hard to manage two packages via the same GitHub repo source.

For sure that is right, but perhaps some steps are or could be automated? That depends on how you do the release process.

In my projects I use the package "standard-version" to generate the changelog file and modify data (e.g. the new version number) in several files in the project folder and subfolders. This way I keep everything in sync.

In the case of `dayjs` what really remains 'troublesome' and where I don't know how to script in a robust way, is the generation of the 2 sets of type definitions. But here modifications are only necessary, if the interface changes - i.e. not too often.

So if this case arises and if you need some support, I would appreciate to help out - just give me a mention or so.

iamkun:
Maybe, we could create a new project/repo to download the latest dayjs from the npm package `dayjs@latest` after each dayjs release, and update the folder structure as well as the typefiles in esm way. And publish it as dayjs-esm

iamkun:
Or, to just release dayjs@2.0 to make esm as default bundle instead of umd

#1281

BePo65:
IMHO, what we have by now is:
1. a library in path `dayjs` that can be used with javascript as an umd module with `require` or `import`
2. matching type definitions for the umd module to use `dayjs` in a typescript project using `import dayjs from 'dayjs'`
3. a library in path `dayjs/esm` that throws an error when used in a javascript esm project (i.e. using `"type": "module"`), as there is no package.json in 'dayjs/esm' that also has a `"type": "module"` setting
4. type definitions for the esm module in `dayjs/esm` that are modified with a npm script to use `export default`, but throw when using the plugins.

The point 3. can easily be solved by copying the package.json from `dayjs` to this subfolder and adding a `"type": "module"` line.  
For example this could be done when building the esm folder.  
When the version number in the main package.json file is updated, then also the version number in this package.json in `dayjs/esm` must be set.

Point 4 is resolved by pr #1581, and IMHO cannot be easily solved with a script that adapts the type definitions (besides creating the right folder structure and replacing 'export =' with 'export default').

If you just release dayjs@2.0 (and kind of drop support for the umd module) you would leave all those people behind that use your lbrary today - and I think there are lots of them.

And to the idea of a new project solely for the esm variant: even with a new project, you would have to modify the type definitions, when the interface of dayjs or one of the plugins is changed.  
So I would prefer to keep everything in one place (except if you want to switch the basic structure to classes instead of the prototype modifications; but this would be a major refactoring).

BePo65:
> Then, there will be two dayjs functions that share the same global variable and may cause trouble.

es-modules don't use global variables (that's the idea behind all those module systems: avoid collisions in global variables). Therefore this should not be the problem.

hakimio:
@iamkun `dayjs` is one of the last widely used libraries which doesn't have proper ES module support. It would be really nice if you could decide if you want fully migrate to ESM and drop support for CJS (which imho is the best option) or go with the solution in this PR and have some duplicated code.

iamkun:
This PR did solve some significant problems related to ESM in dayjs.

However, there will be a massive breaking change if we get this PR merged. 

It's a hard decision to make Tbh.

I'm not sure if fixing it in v2.x is a better choice?

BePo65:
> I'm not sure if fixing it in v2.x is a better choice?

Of course it is; with the start of [dayjs 2.0](https://github.com/iamkun/dayjs/issues/1879) it doesn't make sense any more to add such a hard change to the system.

I am nearly finished with the issues I was commenting and the pr I created for them, so that I will concentrate to contribute a little bit to dayjs 2.0.

I will close this PR - thanks for the discussion ☺️

PR #1589, opened 2021-07-22
Fix esm duration typings.

mshima:
Fixes esm duration typing reference.

Contains the required part of https://github.com/iamkun/dayjs/pull/1544.


mshima:
@iamkun would be nice have this merged before next release.
This is preventing us from switching to esm.

iamkun:
Thanks

PR #1592, opened 2021-07-26
Update Slovenian local

oparex:
Current version of relativeTime in Slovenian locale is inconsistent for future and past together, it only works for the future. Past is changed to be more like "%s ago" in english. One is added to minute, hour, day, month and year to be consistent with other languages.

### DIFF
--- a/src/locale/sl.js
+++ b/src/locale/sl.js
@@ -20,17 +20,17 @@ const locale = {
   },
   relativeTime: {
     future: 'čez %s',
-    past: 'pred %s',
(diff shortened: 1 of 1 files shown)


iamkun:
Maybe we could update sl.js locale like https://github.com/moment/moment/blob/develop/src/locale/sl.js


ref: https://github.com/iamkun/dayjs/blob/dev/src/locale/uk.js#L13

oparex:
Even better!

PR #1594, opened 2021-07-28
Add entry points for es6 imports

szabadkai:
ran prettier/format-package on package.json
added entries for es6 imports


kbsali:
@szabadkai could that fix the issue I reported here : https://github.com/metonym/svelte-time/issues/9 ?

PR #1602, opened 2021-08-10
Use secure url for badgesize

adaex:
This PR modified the url of `img.badgesize.io` to `https` protocol to avoid camo proxying used by github.

View details about [Proxying User Images](https://github.blog/2014-01-28-proxying-user-images).

image


iamkun:
Thanks

Issue #1603, opened 2021-08-12
Wrong typo for utcOffset method from UTC plugin


NikiO-INO:
**Describe the bug**
Current typo for `utcOffset` method from UTC plugin is wrong. That method could accept and properly use `offset` parameter as string, here is an example. https://github.com/iamkun/dayjs/blob/e2feb6bfa555fe3412934cbf18ac93f51ff34d99/src/plugin/utc/index.js#L83

**Expected behavior**
`utcOffset` method should accept offset parameter either as string or number.

**Information**
 - Day.js Version v1.10.6
 - OS: MacOS v11.2.3
 - Browser Safari v14.0.3
 - Time zone: GMT+03:00 MSK

PR #1604, opened 2021-08-12
fix: changed utcOffset method typo from UTC plugin

NikiO-INO:
Fixes https://github.com/iamkun/dayjs/issues/1603


iamkun:
Thanks

PR #1612, opened 2021-08-18
fix: (locale: bg) correct abbreviations & more common plural forms

TonyBogdanov:


iamkun:
seems these's some different with https://github.com/moment/moment/blob/develop/locale/bg.js

TonyBogdanov:
@iamkun Because that one's incorrect as well. I've just pushed a PR there too: https://github.com/moment/moment/pull/5930

PR #1624, opened 2021-09-03
Add parsing ASP.NET style time duration for Duration plugin

luozz1994:
The format is an hour, minute, second string separated by colons like `23:59:59`. The number of days can be prefixed with a dot or  a space separator like so `7.23:59:59` or `7 23:59:59`. Partial seconds are supported as well `23:59:59`.
```
dayjs.duration('23:59:59');
dayjs.duration('7.23:59:59');
dayjs.duration('7 23:59:59');
dayjs.duration('23:59');
```
**Note**: This usage method depends on `Duration` plugin


iamkun:
Well, I think dayjs should only parse standard ISO format.

Maybe it will be better to publish this ASP duration parser dayjs plugin as a separate package.

PR #1627, opened 2021-09-04
Add iraqi arabic locale

msal4:
Add missing iraqi arabic locale that uses the standard arabic month names instead of the western translated names of months.


iamkun:
Thanks

PR #1640, opened 2021-09-22
fix: add config type definition to objectSupport plugin

andrekovac:
Currently Typescript gives an error when using a date in object notation as provided by the `objectSupport` plugin. 

For example this will give a TypeScript error: 

```ts
dayjs.utc({
  y: 2021,
  M: 10,
  d: 1,
});
```

This PR fixes it.


ryanknittel-fdx:
This PR would be really helpful. Currently, parse ObjectSupport is broken in TypeScript.

image

tonydaly:
Any updates on this, would be super useful!

PR #1647, opened 2021-09-28
Add objectSupport argument type

ypresto:
Gives concrete argument type to objectSupport plugin :)


ypresto:
Although after creating patch I found #1640, I sent PR because this also covers set/add/subtract.

ryanknittel-fdx:
This PR would be really helpful. Currently, parse ObjectSupport is broken in TypeScript.

image

PR #1654, opened 2021-10-01
[Typescript] Fix ManipulateType type

burtek:
Should be [`Exclude`](https://www.typescriptlang.org/docs/handbook/utility-types.html#excludetype-excludedunion) rather than [`Omit`](https://www.typescriptlang.org/docs/handbook/utility-types.html#omittype-keys) as the latter works on objects while former works on unions.


simonbw:
I came here to suggest this same change.

burtek:
Closing as it's already fixed in `master` as per https://github.com/iamkun/dayjs/blob/05cba7e2ac6327011169b56ce4fb7e51e11eb1d4/types/index.d.ts#L32

Issue #306, opened 2018-08-17
dayjs().isValid();怎么用？


kikaendeavor:
`dayjs()` 返回当前系统时间    
`dayjs().isValid();` 肯定为true   
那么传入一个参数呢？   
 `dayjs(‘2018-08-107’).isValid();`   
 `dayjs(‘2018-13-01’).isValid();`   
 `dayjs(‘1-10-01’).isValid();`   
结果都是true！   

是我打开方式不对吧 “...“

rfoel:
It's a bit weird... If one inputs `2018-13-01` the dayjs object will be `2019-01-01`...

xxyuk:
@rfoel 

Javascript did the same.

`new Date(2018, 13 , 01)` => valid date

kikaendeavor:
Actually，    
new Date(2018,13,01)             => 2019-01-31T16:00:00.000Z      
new Date('2018-08-107')        => Invalid Date   
new Date('1-10-01')                => Invalid Date   

But   
dayjs(‘2018-13-01’).isValid();   => true   
dayjs(‘2018-08-107’).isValid(); => true   
dayjs(‘1-10-01’).isValid();         => true    

so, there must be something wrong with the isValid function.   
`
isValid() {
    return !(this.$d.toString() === 'Invalid Date')
}
`

xxyuk:
@kikaendeavor 

`new Date('1-10-01')` => Wed Jan 10 2001 00:00:00

kikaendeavor:
@xxyuk    

woooh~~~~    
the result is different between chrome and firefox.

PR #375, opened 2018-10-11
Improved validation if strings matches a valid date

JonasSchubert:
This should resolve #306 and #320

Issue #320, opened 2018-08-31
Validation .isValid() doesn't work always


dakshshah96:
The `.isValid()` validation doesn't function correctly under the following condition:

- When the the day part is 31 for a month which didn't have 31 days.

Some instances where `.isValid()` returns `true` for a date that is invalid:

- `dayjs().isValid('1995-02-31')`
- `dayjs().isValid('1995-03-31')`
- `dayjs().isValid('1995-04-31')`
- `dayjs().isValid('1995-06-31')`
- `dayjs().isValid('1995-09-31')`
- `dayjs().isValid('1995-11-31')`

mixrich:
@dakshshah96 
`.isValid()` method has't attributes and passed value is discarded like if you call `dayjs().isValid()`. And you know, that `dayjs()` return current date which is always valid

mixrich:
Moreover, `isValid` doesn't validate if passed date is `exists`, it's just validate that date was parsed correctly. So, if you pass `dayjs('2018-11-41').isValid()` the date will be parsed as `2018-12-10` and `isValid` will return `true` anyway

prantlf:
You should not overestimate the quality of such validation. You will be able to catch nonsense like `MM=15`, `DD=42` or `DD=30` in February, if the user exchanges values of `DD` with `MM`, or types something totally wrong. However, you won't be able to ensure, that the whole date - particularly `DD` and `MM` - will be set, as the user intended from an input like `02/05/2018`. Was it May 2 or February 2? The user will still have to be aware of the actual format like `DD/MM/YYYY`. The date picker control usually helps better with ensuring that, than a robuster parser.

majstorzelja:
Is this an option to validate if date exist?

```
function validate(date, format) {
  return dayjs(date, format).format(format) === date;
}

validate('2019/02/31', 'YYYY/MM/DD') // false
```

YangAi:
> Is this an option to validate if date exist?
> 
> ```
> function validate(date, format) {
>   return dayjs(date, format).format(format) === date;
> }
> 
> validate('2019/02/31', 'YYYY/MM/DD') // false
> ```

Thanks! It's very helpful.

ness-EE:
I'm just swapping over from Moment to DayJS. The unit tests I have in place to validate incorrect entries like 2000-02-31 pass with Moment and fail with DayJS

bitnbytesio:
facing same issue, isValid() always returns true

```js
const dayjs = require("dayjs")
[...]

PR #1658, opened 2021-10-05
feat(strict-plugin): add plugin for strict month parsing #306 #320

JonasSchubert:
This should resolve the issues #306 and #320 and replaces the pull request #375 

As suggested in the mentioned PR, we could use this as an plugin instead of adding it to the core functionality


src/plugin/strict/index.js
```diff
@@ -0,0 +1,45 @@
...
+      // Treat this condition as an invalid date
+      if (!m || m > 11 || m < 0) return new Date(NaN)
+      let daysPerMonth = DAYS_PER_MONTH[d[2] - 1]
+      if (m === 1 // If month is February increase days per month by 1 if we are in a leap year
+        && (((d[1] % 4 === 0) && (d[1] % 100 !== 0)) || (d[1] % 400 === 0))) daysPerMonth += 1
+      // Treat this condition as an invalid date
+      if ((!!d[3] && (d[3] > daysPerMonth || d[3] < 1))
+        || (!!d[4] && (d[4] > 23 || d[4] < 0))
+        || (!!d[5] && (d[5] > 59 || d[5] < 0))
+        || (!!d[6] && (d[6] > 59 || d[6] < 0))) return new Date(NaN)
```
fragsalat, line 24: I'm not a maintainer of this project but stumbled upon this because of February date validation. Just some feedback, thx for your contribution but this code is not readable. Could you may destruct the match result into variables with readable names? No one knows which value is in d[4] or d5[5]

JonasSchubert: I tried to keep the footprint small, but you are right. I am using destructuring now and named the variables like date formats to still keep it lightweight. What do you think?

fragsalat:
:(

PR #1666, opened 2021-10-14
fix: Correct `dayjs.extend`'s return type

melusc:
`dayjs.extend` returns `dayjs` (lowercase d), not `Dayjs`.

This means TypeScript allows chaining `extend` calls like `dayjs.extend(...).extend(...)` instead of having to do `dayjs.extend(...); dayjs.extend(...)`.


melusc:
@iamkun 

Not to bother you, but is there anything keeping this simple change from being merged? Just cause it's been open for two years now.

PR #1678, opened 2021-10-31
fix: +00:00 timezone bug

DimaBur:
When we have +00:00 timezone offset like 'Europe/London" dayjs gets users timezone offset, which causes incorrect time showing in different cases.

**Example**

Current behavior:
`dayjs('2021-10-31T15:00:00+00:00').tz('Europe/Moscow').format() // 2020-12-20T23:30:00Z`

Behavior after fix:
`dayjs('2021-10-31T15:00:00+00:00').tz('Europe/Moscow').format() // 2021-10-31T23:30:00+00:00`

### DIFF
--- a/src/plugin/utc/index.js
+++ b/src/plugin/utc/index.js
@@ -114,7 +114,8 @@ export default (option, Dayjs, dayjs) => {
 
   proto.valueOf = function () {
     const addedOffset = !this.$utils().u(this.$offset)
-      ? this.$offset + (this.$x.$localOffset || (new Date()).getTimezoneOffset()) : 0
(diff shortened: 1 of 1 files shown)


src/plugin/utc/index.js
```diff
@@ -114,7 +114,8 @@ export default (option, Dayjs, dayjs) => {
 
   proto.valueOf = function () {
     const addedOffset = !this.$utils().u(this.$offset)
-      ? this.$offset + (this.$x.$localOffset || (new Date()).getTimezoneOffset()) : 0
```
NamPNQ, line 117: I think you can use `??` operator to fix this

felipehertzer:
Hey @iamkun is there any plan to merge this fix? I'm having the same problem with Europe/London. Thanks

PR #1682, opened 2021-11-02
fix: fix wrong esmodule path

xty1992a:
package.json should provide the `module` to guide esmodule project import dayjs from the right path.

umd format try to find the global variable `exports`. that not exist in the esmodule environment
> umd code like this
```js
"object"==typeof exports&&"undefined"!=typeof module
```


G-11-P (commented):
Merge

PR #1448, opened 2021-04-09
fix: utcOffset when local in DST but date is not

aloisklink:
Fixes a bug in the valueOf() function after setting a utcOffset, when the the local timezone offset changes.

I added a test for it to `test/timezone.test.js`, so it should be reproducible on other people's computers as well.

Without this fix, when trying to run `npm test` in my timezone (British Summer Time - BST), there are quite a lot of off-by-one errors, e.g.

```
  ● DST, a time that never existed Fall Back › 2012-11-04 02:00:00

    expect(received).toBe(expected) // Object.is equality
    
    Expected value to be:
      1352012400000
    Received:
      1352016000000

      206 |       expect(d.format()).toBe('2012-11-04T02:00:00-05:00')
      207 |       expect(d.utcOffset()).toBe(-300)
    > 208 |       expect(d.valueOf()).toBe(1352012400000)
      209 |     })
      210 |   })
      211 | })
      
      at forEach (test/plugin/timezone.test.js:208:27)
          at Array.forEach (<anonymous>)
      at Object.<anonymous> (test/plugin/timezone.test.js:204:21)

  ● keepLocalTime › keepLocalTime

    expect(received).toBe(expected) // Object.is equality
    
    Expected value to be:
      "2013-11-18T17:55:00+01:00"
    Received:
      "2013-11-18T18:55:00+02:00"

      263 |   const base = dayjs.tz('2013-11-18 11:55', 'America/Toronto')
      264 |   it('keepLocalTime', () => {
    > 265 |     expect(base.tz('Europe/Berlin').format()).toBe('2013-11-18T17:55:00+01:00')
      266 |     expect(base.tz('Europe/Berlin', true).format()).toBe('2013-11-18T11:55:00+01:00')
      267 |   })
      268 | })
      
      at Object.<anonymous> (test/plugin/timezone.test.js:265:47)

```

PR #1699, opened 2021-11-15
Fix: Correct get valueOf for utc plugin

NamPNQ:
For get correct offset DST, use `this.$d. getTimezoneOffset()` instead of `new Date()`


aloisklink:
This seems to be the same fix as PR #1448, but there's another test so both can potentially be merged :+1:

NamPNQ:
Thank @aloisklink, I give you the PR to your branch https://github.com/aloisklink/dayjs/pull/1/ to add my test

PR #1746, opened 2021-12-16
Update de.js

cazualdev:
Updating the monthsShort value in the de.js locale to match the German standard like de-at and de-ch where there is a period to abbreviate the necessary months.


iamkun:
ref https://github.com/moment/moment/blob/develop/src/locale/de.js#L28

PR #1760, opened 2022-01-09
fix: allow dayjs.tz()

edrichhans:
Allow empty params for `dayjs.tz()` since we're currently already allowing `undefined` for `ConfigType`


edrichhans:
@iamkun I dont know why the build failed for `locale`...

Issue #1720, opened 2021-11-30
Case sensitivity parsing with customParseFormat plugin 'MMM' or 'MMMM'


devi-green:
Is there a way to ignore case when parsing the month formats 'MMM' or 'MMMM'?

dayjs("2020-NOV-02", "YYYY-MMMM-DD", false); - invalid
dayjs("2020-Nov-02", "YYYY-MMMM-DD", false); - valid
dayjs("2020-nov-02", "YYYY-MMMM-DD", false); - invalid

eliandoran:
Hello,

It appears this functionality is not present yet. I needed this functionality as well, so I implemented a PR: https://github.com/iamkun/dayjs/pull/1769

BePo65:
Just as an additional argument: the parsing in moment are case insensitive too (as the tests in your pr prove).

And as a funny aside: moment even parses 'Nov' as input when using the long format ('MMMM') as in your first comment :-)

PR #1769, opened 2022-01-16
customParseFormat: Add case insensitive search for `MMM` and `MMMM` format

eliandoran:
Parsing date inputs such as `21-December-18` (MMMM) and `2018 Feb 03` (MMM) works just fine, however as soon as the user enters inputs such as `21-december-18` or `2018 Feb 03`, the date will not be parsed.

This PR implements a method which searches through the months with case insensitive.

Motivation for this PR: Was working on a application which parses user dates, but it would fail when users typed months using only small letters.

Initially wanted to implement case insensitive search only if `strict` mode is off, however it appears that case insensitivity is already handled in strict mode as well (see `recognizes midnight in small letters` test).

### DIFF
--- a/src/plugin/customParseFormat/index.js
+++ b/src/plugin/customParseFormat/index.js
@@ -18,6 +18,17 @@ let parseTwoDigitYear = function (input) {
   return input + (input > 68 ? 1900 : 2000)
 }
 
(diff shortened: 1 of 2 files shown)


BePo65:
fixes issue #1720

Issue #1237, opened 2020-11-23
DayJS parses wrong year value for the dates "0000-01-01" and "0001-01-01"


matheusot:
**Describe the bug**
When trying to parse the dates "0000-01-01" and "0001-01-01" DayJS returns the years 1900 and 1901 respectively:
```
const dayjs = require('dayjs')
console.log(dayjs("0000-01-01", "YYYY-MM-DD"))
console.log(dayjs("0001-01-01", "YYYY-MM-DD"))
```
Output:
```
d {
  '$L': 'en',
  '$d': 1900-01-01T03:06:28.000Z,
  '$x': {},
  '$y': 1900,
  '$M': 0,
  '$D': 1,
  '$W': 1,
  '$H': 0,
  '$m': 0,
  '$s': 0,
  '$ms': 0
}
d {
  '$L': 'en',
  '$d': 1901-01-01T03:06:28.000Z,
  '$x': {},
  '$y': 1901,
  '$M': 0,
  '$D': 1,
  '$W': 2,
  '$H': 0,
  '$m': 0,
  '$s': 0,
  '$ms': 0
}
```

**Expected behavior**
The expected behavior is that the returned object has the correct year. Setting the year manually we can see the correct behavior working:
```
const dayjs = require('dayjs')
console.log(dayjs("2020-01-01", "YYYY-MM-DD").year(0))
console.log(dayjs("2020-01-01", "YYYY-MM-DD").year(1))
```
Output:
```
d {
  '$L': 'en',
  '$d': 0000-01-01T03:06:28.000Z,
  '$x': {},
  '$y': 0,
  '$M': 0,
  '$D': 1,
  '$W': 6,
  '$H': 0,
  '$m': 0,
  '$s': 0,
  '$ms': 0
}
d {
  '$L': 'en',
  '$d': 0001-01-01T03:06:28.000Z,
  '$x': {},
  '$y': 1,
  '$M': 0,
  '$D': 1,
  '$W': 1,
  '$H': 0,
  '$m': 0,
  '$s': 0,
  '$ms': 0
}
```

**Information**
 - Day.js Version v1.9.6
 - OS: Windows
 - Browser: N/A (Node.js v12.18.3)
 - Time zone: GMT-03:00

iamkun:
will this is a new concept to me. What's the year '0000' stand for, pls?

matheusot:
[...]

PR #1777, opened 2022-01-23
fix: incorrect parsing of year string with leading zeros (#596)

### LINKED ISSUE #596: Incorrect parsing of year string with leading zeros
Thanks for a great library, just has moved a project to it.

**Describe the bug**
Incorrect parsing of dates in the format "YYYY-MM-DD" when YYYY contains leading zeros.
dayjs("0001-01-01") is Jan 01 1901

This issue is related to the usage of the following constructor of Date in JS 
`new Date(year, monthIndex)`; 
[source code](https://github.com/iamkun/dayjs/blob/dev/src/index.js#L60)

**Expected behavior**
dayjs("0001-01-01") should be Jan 01 0001

**Information**
 - Day.js Version 1.8.13
 - OS: Windows 10 1809 x64
 - Browser Chrome 70/Firefox 62
 - Time zone: GMT+0300

rudants:
fix #596

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -62,12 +62,12 @@ const parseDate = (cfg) => {
     if (d) {
       const m = d[2] - 1 || 0
       const ms = (d[7] || '0').substring(0, 3)
-      if (utc) {
-        return new Date(Date.UTC(d[1], m, d[3]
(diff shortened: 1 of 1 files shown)


BePo65:
This is a little more compact version of PR #1862; solves issue #1237 too.

PR #1781, opened 2022-01-25
fix(parse): fix parsing milliseconds less than three digits

masontikhonov:
Currently strings without timezone and with milliseconds less than three digits (and without trailing zeros) are being parsed incorrectly.

For example (pay attention to milliseconds):
* String `2022-01-25T09:54:22.5` expected to be parsed as `2022-01-25T09:54:22.500`, but instead it's treated as `2022-01-25T09:54:22.005`.
* String `2022-01-25T09:54:22.42` expected to be parsed as `2022-01-25T09:54:22.420`, but instead it's treated as `2022-01-25T09:54:22.042`.

This PR proposes a solution for this issue, by padding milliseconds with zeros to three characters (if less).

The proposed changes are covered by randomized tests.

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -61,7 +61,7 @@ const parseDate = (cfg) => {
     const d = date.match(C.REGEX_PARSE)
     if (d) {
       const m = d[2] - 1 || 0
-      const ms = (d[7] || '0').substring(0, 3)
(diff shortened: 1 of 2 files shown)


oaraujocesar (approved):

masontikhonov:
@iamkun Hi! Did you have a chance to look at this PR?

PR #1798, opened 2022-02-07
add kurdish (sorani) language

swara-mohammed:


src/locale/ckb.js
```diff
@@ -0,0 +1,40 @@
+// Kurdish [ckb]
```
iamkun, line 1: accroding to https://en.wikipedia.org/wiki/List_of_ISO_639-2_codes 

should we rename the filename to ku.js?

swara-mohammed: We have two big and different dialects in Kurdish, sorani and Kurmanji. Each of them has different suffixes and alphabets, because unfortunately we don't have a Country we couldn't create a unified language. So if, for example, Sorani receives a suffix in a project, Kurmanji will be deprived of services.

iamkun: This is a bit difficult for me to understand. Would you please ask some of the native speakers to review this PR before it gets merged?

AkarM13: We have two different dialects, Sorani and Kurmanji. The language code for Kurmanji is usually "kmr" and the language code for Sorani is "ckb" which means Central Kurdish. What he means is if we have plans to add the Kurmanji variant to dayjs there will be conflicts. We should rename the "ku" key to "ckb" because that's what the original locale suggests in dayjs.

swarakaka: Hello @iamkun 
Is there any attempt to merge this PR?

swarakaka: @iamkun 
could you please accept this PR?

PR #1800, opened 2022-02-09
Added relative time to .gl (Galician) and fixed ordinals

anxogcd:
I've added relative time expressions to Galician language and also fixed ordinals in this language.


nuno-andre:
The Galician word for _ago_ is _hai_, not _fai_.

> _It has now been a year since sth took place_
> _Hoxe **fai** un ano que pasou algo_

vs.

> _A year ago sht took place_
> _**Hai** un ano que pasou algo_


https://dubidasdogalego.wordpress.com/2017/03/23/hai-fai-vai/

PR #1809, opened 2022-02-18
fix: make setDefaultTimezone a global default timezone (#1227)

### LINKED ISSUE #1227: day.js default timezone problem
Hi there. I am having trouble in the following scenario:
When I set the default timezone, using dayjs.tz.setDefault("America/New_York") and then create new object using
the dayjs() constructor, the newly created object does not have the timezone set. The only way to have timezone on the object is
by using the dayjs.tz() constructor, but in this case the expected arguments are completely different.
So, is there a way to set default timezone and be able to create objects with the standart constructor with the default timezone 
set?
Code example:
dayjs.tz.setDefault("America/New_York");

dayjs("20/11/2020","DD/MM/YYYY"); // This object does not have the default timezone set

dayjs.tz(); //Here, we have the default timezone set to America/New_York

**Information**
 - Day.js Version [e.g. v1.0.0]
 - Browser [e.g. chrome 62]
 - Time zone: [e.g. GMT-07:00 DST (Pacific Daylight Time)]

FlareStone:
close #1227 

before:
the `dayjs.tz.setDefaultTimezone` in plugin/timezone only changed the timezone of `dayjs.tz()`.
after:
the `dayjs.tz.setDefaultTimezone` in plugin/timezone will changed the timezone of `dayjs()` too.

### DIFF
--- a/src/plugin/timezone/index.js
+++ b/src/plugin/timezone/index.js
@@ -1,4 +1,4 @@
-import { MIN, MS } from '../../constant'
+import { MILLISECONDS_A_MINUTE, MILLISECONDS_A_SECOND, MIN, MS } from '../../constant'
 
 const typeToPos = {
   year: 0,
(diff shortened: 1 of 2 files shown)


kihyunkimlee:
why did you close?

PR #1812, opened 2022-02-22
Fix/ fix some bug of utcOffset

### LINKED ISSUE #1803: utc.utcOffset does not return a new instance when keepLocalTime is set to true
**Describe the bug**
`utcOffset(value, keepLocalTime)` method of `utc` plugin is not returning a new `Dayjs` instance when `keepLocalTime = true`. This then breaks everything, because the internal variable of the original instance are not properly re-initialized with the new `$offset` and `$u` variables. This also breaks `startOf` in `timezone` plugin and probably more?

```typescript
/**
 * EXPECTED SYSTEM DATE SETTINGS:
 * DATE: 2022-02-10T10:00:00Z
 * TIMEZONE: Europe/Prague (GMT+1)
 */

const dayjs = require('dayjs')
const utc = require('dayjs/plugin/utc')

dayjs.extend(utc)

const date = dayjs('2022-02-10T10:00:00Z')
const dateWithZero = date.utcOffset(0, true)

console.log(date.format()) // incorrect - 2022-02-10T11:00:00Z, expected 2022-02-10T11:00:00+01:00
console.log(dateWithZero.format()) // correct - 2022-02-10T11:00:00Z
console.log(dateWithZero.add(1, 'hour').format()) // incorrect - 2022-02-10T11:00:00Z, expected 2022-02-10T12:00:00Z
console.log(date === dateWithZero) // incorrect - true
```

The issue is here, the `inc` should probably be cloned at line 92? Not sure.
https://github.com/iamkun/dayjs/blob/4a7b7d07c885bb9338514c234dbb708e24e9863e/src/plugin/utc/index.js#L90-L95

**Expected behavior**
New `Dayjs` instance with proper internal variables should be returned.

**Information**
 - Day.js Version 1.10.7
 - Time zone: GMT+1
 - broken since https://github.com/iamkun/dayjs/commit/9f488e5aca92f0b4c2951459436829d79f86d8d7

FlareStone:
1. fix utc plugin when utcOffset(0, true)

before: `const d = dayjs().utcOffset(0, true)` and `d.clone().format() === d.format()` is false
after: fixed

2. fix error when set utc offset continuously
before: when i use utcOffset continuously more than two times, the value will be error
after: fixed

3. fix #1803

### DIFF
--- a/src/plugin/utc/index.js
+++ b/src/plugin/utc/index.js
@@ -94,8 +94,7 @@ export default (option, Dayjs, dayjs) => {
       return ins
     }
     if (input !== 0) {
-      const localTimezoneOffset = this.$u
-        ? this.toDate().getTimezoneOffset() : -1 * this.utcOffset()
+      const localTimezoneOffset = this.toDate().getTimezoneOffset()
       ins = this.local().add(offset + localTimezoneOffset, MIN)
       ins.$offset = offset
       ins.$x.$localOffset = localTimezoneOffset
--- a/src/plugin/utc/index.js
+++ b/src/plugin/utc/index.js
@@ -53,7 +53,7 @@ export default (option, Dayjs, dayjs) => {
 
   const oldInit = proto.init
   proto.init = function () {
-    if (this.$u) {
+    if (this.$u && this.$utils().u(this.$offset)) {
       const { $d } = this
       this.$y = $d.getUTCFullYear()
       this.$M = $d.getUTCMonth()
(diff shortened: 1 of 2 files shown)


PengJiyuan:
Thanks a lot, you saved my day!

SkaceKamen:
I think your fix is not really working correctly, since the incorrect tests mentioned here are not failing: https://github.com/iamkun/dayjs/issues/1803#issuecomment-1036064306

Issue #1804, opened 2022-02-11
utc.valueOf doesn't account for DST correctly


janzipek:
**Describe the bug**
The `utc.valueOf` method doesn't correctly account for DST when `$x.$localOffset` is not set:
https://github.com/iamkun/dayjs/blob/4a7b7d07c885bb9338514c234dbb708e24e9863e/src/plugin/utc/index.js#L116-L117
the issue is it creates new `Date` object and retrieves the browser timezone offset from that, but if the user is in DST and the date that's being casted to timestamp is not (or vice-versa), the result is incorrect.

Minimal repro:
```javascript
/*
 Expected system date settings:
 TIMEZONE: Europe/Prague (+1)
 DATE: 2022-02-11
*/

const d = dayjs('2014-06-01T12:00:00Z').tz('America/Los_Angeles')
console.log(d.format()) // correct - 2014-06-01T05:00:00-07:00
console.log(new Date(d.valueOf()).toISOString()) // incorrect 2014-06-01T11:00:00.000Z, expected 2014-06-01T12:00:00.000Z
console.log(d.valueOf()) // incorrect 1401620400000, expected 1401624000000
```

**Expected behavior**
The browser timezone offset should be calculated from the date itself, not new instance. Something like this:
```javascript
  proto.valueOf = function () {
    const addedOffset = !this.$utils().u(this.$offset)
      ? this.$offset + (this.$x.$localOffset || this.$d.getTimezoneOffset()) : 0
    return this.$d.valueOf() - (addedOffset * MILLISECONDS_A_MINUTE)
  }
```

**Information**
 - Day.js Version v1.10.7

xvaara:
there is a pull request to fix this #1448 and there is a plugin in the pulls conversation to fix this while we wait.
but since it's a year old and not merged (even though it's a big problem), one has to wonder where to focus of this project is.

PR #1815, opened 2022-03-01
Fixed utc.valueOf not taking DST into account

SkaceKamen:
Use the current date to determine internal timezone offset instead of using a new Date instance, since the new Date returns a different offset based on whenever DST is in effect at the moment.

This PR fixes https://github.com/iamkun/dayjs/issues/1804


xvaara:
this is same as #1448
but since it's a year old and not merged (even though it's a big problem), one has to wonder where to focus of this project is.

iamkun:
Thanks, fixed in #1448

PR #1819, opened 2022-03-03
fix: valueOf not correct when cross summertime

huoyuxin:
eg: 

- system timezone Europe/Warsaw

- system date 2022-03-01

dayjs.tz.setDefault('Europe/Warsaw')
dayjs.tz().endOf('month').valueOf()  -> 1648760399999,  it should be 1648763999999

### DIFF
--- a/src/plugin/utc/index.js
+++ b/src/plugin/utc/index.js
@@ -114,7 +114,7 @@ export default (option, Dayjs, dayjs) => {
 
   proto.valueOf = function () {
     const addedOffset = !this.$utils().u(this.$offset)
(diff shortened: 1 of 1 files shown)


xvaara:
this is same as #1448
but since it's a year old and not merged (even though it's a big problem), one has to wonder where to focus of this project is.

huoyuxin:
> this is same as #1448 but since it's a year old and not merged (even though it's a big problem), one has to wonder where to focus of this project is.

does it mean that mr will not be merged, and summertime users like me can only fixed it in project ..

PR #1831, opened 2022-03-14
Update de.js

PhilippMeissner:
[[This PR adds the abbreviations for German months including a `.` in the end, as in `September` -> `Sept.`.
[Duden DE](https://www.duden.de/rechtschreibung/Sept_)


iamkun:
I'm not sure if this is a proper fix, seems we are the same with https://github.com/moment/moment/blob/develop/src/locale/de.js#L28

PhilippMeissner:
> I'm not sure if this is a proper fix, seems we are the same with https://github.com/moment/moment/blob/develop/src/locale/de.js#L28

The abbreviation for `September` is different though. The proper abbreviation is `Sept.`, not `Sep`.

iamkun:
Thanks

PR #1846, opened 2022-04-09
Update pt-br.js

jonashenrique90:
This PR improves the weekdaysMin for brazilian portuguese.


iamkun:
I'll mark this PR to wait for some native speaker review.

also you can check https://github.com/moment/moment/blob/develop/src/locale/pt-br.js as a reference

CaffeJappa:
I'm a Brazilian Portuguese native speaker, the improved weekdaysMin this guy made is indeed more used than the one in use by DayJS ;)

yanpgabriel:
I'm also a native speaker and I can confirm that using 2ª, 3ª, 4ª, 5ª, and 6ª is not at all conventional.

Issue #1728, opened 2021-12-01
Kurdish locale strings contains non-standard kurdish characters


akamfoad:
**Describe the bug**

Kurdish locale strings contains non-kurdish standard letters for instance:
- Using  [`ه` -> `U+0647`] instead of [`ە` -> `U+06D5`] 
- Using [`ك` -> `U+0643`] instead of [`ک` -> `U+06A9`]

**Expected behavior**
To use right Kurdish Unicode characters as des

#### Kurdish language alphabet
| Unicode | Character | Unicode | Character | Unicode | Character | Unicode | Character |
| ------- | --------- | ------- | --------- | ------- | --------- | ------- | --------- |
| U+0626  | ئ         | U+0627  | ا         | U+0628  | ب         | U+067E  | پ         |
| U+062A  | ت         | U+062C  | ج         | U+0686  | چ         | U+062D  | ح         |
| U+062E  | خ         | U+062F  | د         | U+0631  | ر         | U+0695  | ڕ         |
| U+0632  | ز         | U+0698  | ژ         | U+0633  | س         | U+0634  | ش         |
| U+0639  | ع         | U+063A  | غ         | U+0641  | ف         | U+06A4  | ڤ         |
| U+0642  | ق         | U+06A9  | ک         | U+06AF  | گ         | U+0644  | ل         |
| U+06B5  | ڵ         | U+0645  | م         | U+0646  | ن         | U+0648  | و         |
| U+06C6  | ۆ         | U+0647  | ه         | U+06D5  | ە         | U+06CC  | ی         |
| U+06CE  | ێ         |

#### Numbers:
| Unicode | Number | Unicode | Number | Unicode | Number | Unicode | Number |
| ------- | ------ | ------- | ------ | ------- | ------ | ------- | ------ |
| U+0660  | ٠      | U+0661  | ١      | U+0662  | ٢      | U+0663  | ٣      |
| U+0664  | ٤      | U+0665  | ٥      | U+0666  | ٦      | U+0667  | ٧      |
| U+0668  | ٨      | U+0669  | ٩      |

**Information**
 - Day.js Version v1.10.7
 - OS: Irrelevant
 - Browser Irrelevant
 - Time zone: Irrelevant

PR #1848, opened 2022-04-09
fix: Kurdish locale strings and formatted output contains non-standard kurdish characters

akamfoad:
1. move: arabic to english number maps into constants:
  **Arabic** numbers are common in many `RTL` languages, which makes sense for `Arabic`->`English and `English`->`Arabic` numbers mapping be factored out into a more general file so that other locales can import it and use it without the need to duplicate in the source code it.
2. fix: https://github.com/iamkun/dayjs/issues/1728
  - fix usage of non-standard Kurdish characters in ku locale
  - add preparse and postformat to ku locale so that the formatted dates will contain correct Arabic numbers that Kurdish language uses.
3. fix: month names to be more aligned towards Kurdish month names as explained in this [Wikipedia page](https://ckb.wikipedia.org/wiki/%DA%95%DB%86%DA%98%DA%98%D9%85%DB%8E%D8%B1%DB%8C_%D8%B2%D8%A7%DB%8C%DB%8C%D9%86%DB%8C)


mhmd-azeez (approved):
lgtm

iamkun:
Thanks.

Would you please revert the change "move: arabic to english number maps into constants:"? 

I think it's better to put them inside each locale file.

akamfoad:
@mhmd-azeez Thanks for the review, I really appreciate it.

akamfoad:
@iamkun Of course, done!

iamkun:
Cool

PR #1862, opened 2022-04-19
fix: parseDate stop the js Date behavior that maps 2 digit years to 1900 - 1999

shye0000:
This is the fix of the issue https://github.com/iamkun/dayjs/issues/1237

To achieve this, I just call the `setFullYear` method of the Date object to force the date object to have the exact year.

plus after reading the tests code, I've understood that one of the goals of the `parseDate` method is to align with momentjs, so in my opinion this fix would reinforce the alignment.

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -65,14 +65,24 @@ const parseDate = (cfg) => {
   if (typeof date === 'string' && !/Z$/i.test(date)) {
     const d = date.match(C.REGEX_PARSE)
     if (d) {
+      let dObj = null
+      const y = d[1]
       const m = d[2] - 1 || 0
       const ms = (d[7] || '0').substring(0, 3)
+      const dParams = [y, m, d[3] || 1, d[4] || 0, d[5] || 0, d[6] || 0, ms]
       if (utc) {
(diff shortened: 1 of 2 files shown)


ericbuehl (approved):

BePo65:
This PR solves issue #596 too.

BePo65:
This PR solves issue #596 too.

shye0000:
@iamkun yo it's been a while since this PR has been created, would you plz take a look and maybe potentially include it in a future release?

timmd909:
Any updates on this?

shye0000:
@iamkun 哥们儿，看看咱这个PR呗 ;)
@iamkun yo man, plz check this out ;)

Alluvi0n:
I tried migrating from moment to dayjs, and this issue is preventing the year from being edited in <input type="date" /> elements, as the year briefly becomes '0000-XX-XX' (which is defaulted to 1900) as the user types.

@iamkun - I'm being forced to roll back our migration because daysjs is incompatible with <input type="date" /> value binding. This needs to be addressed before dayjs can be adopted.

PR #1865, opened 2022-04-21
fix: add OpUnitType to quarterOfYear startOf/endOf types

jbmiles:
Fixes a type error I was getting using dayjs with the quarterOfYear plugin.

[example](https://codesandbox.io/s/dayjs-types-5umhqk)


iamkun:
Thanks

mtt87:
👋 Was there a regression with this? I can't really figure out when the code was changed (perhaps there was a force push that rewrote the history?) but seems like this fix is not there anymore. 

I'm looking now at `dev` branch and it doesn't seem to be there anymore
https://github.com/iamkun/dayjs/blob/dev/types/plugin/quarterOfYear.d.ts
```ts
isBefore(date?: ConfigType, unit?: QUnitType): boolean
```

PR #1873, opened 2022-04-27
fix(docs): typo and grammar

dev-itsheng:
When I read the documentation for day.js, I noticed the following sample code:

```javascript
dayjsClass.prototype.format = function(arguments) {
    // original format result
    const result = oldFormat(arguments)
    // return modified result
}
```

I think it should make some changes.

1. `arguments` is a reserved word and may have compatibility issues in some special scenarios.

2. Some parameters of the function in the example are not only one. For example, `isSameOrBefore` accepts two parameters, and only the first one can be obtained by the method in the example.

So I did the refactoring using ES6's spread operator.

---

Also, I've fixed some syntax errors (either code, or text) in the documentation that I've seen.

### DIFF
--- a/types/index.d.ts
+++ b/types/index.d.ts
@@ -24,7 +24,7 @@ declare namespace dayjs {
   export type UnitTypeLong = 'millisecond' | 'second' | 'minute' | 'hour' | 'day' | 'month' | 'year' | 'date'
 
   export type UnitTypeLongPlural = 'milliseconds' | 'seconds' | 'minutes' | 'hours' | 'days' | 'months' | 'years' | 'dates'
-  
+
   export type UnitType = UnitTypeLong | UnitTypeLongPlural | UnitTypeShort;
 
   export type OpUnitType = UnitType | "week" | "weeks" | 'w';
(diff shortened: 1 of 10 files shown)


types/index.d.ts
```diff
@@ -80,7 +80,7 @@ declare namespace dayjs {
      *
      * Months are zero indexed, so January is month 0.
      *
-     * Accepts numbers from 0 to 11. If the range is exceeded, it will bubble up to the next year.
```
iamkun, line 83: seems "accepts" is better?

iamkun:
Thanks, I agree with you in most of the cases.

Besides, we have a new doc repo https://github.com/dayjs/dayjs-website

types/index.d.ts
dev-itsheng: Because I found that the first sentence of this comment is 'Set the month.', similarly, the first sentence of the other method's comment is 'Get' or 'Set', so I think it should be consistent,

iamkun:
thanks, I'll close this pr first, since all of our docs has been moved to the new doc site

PR #1876, opened 2022-04-29
Improve PT-BR translation for README-pt-br.md

ThallesP:
I translated the part about the documentation, which is the only thing that isn't translated yet in the README.


iamkun:
THX

PR #1887, opened 2022-05-09
Add a method to calculate weeks in current month

HHN0609:
Add a method to calculate weeks in current month， including the Incomplete weeks.


HHN0609:
I think this function is useful in transaction.

iamkun:
Thanks. We intend to limit the public API to make the core bundle simple.

Feel free to ship it as a plugin via a npm package.

PR #1889, opened 2022-05-09
Fix: performance issue in tz plugin.

### LINKED ISSUE #1236: Cache toLocaleString
Calls to "toLocaleString" in the timezone plugin is extremly slow:
https://github.com/iamkun/dayjs/blob/ed9629b5ab2895652111fc854e6081422ed5c010/src/plugin/timezone/index.js#L99

The same problem was discussed in luxon: https://github.com/moment/luxon/issues/352

The solution for this performance problem is caching like they did in luxon: https://github.com/moment/luxon/commit/bb77d5e6b968b0a2cc1be72d295f40a43a6687fb

Just use Intl.DateTimeFormat instead of "toLocaleString" (it always produce the equivalent result) and cache these instances of Intl.DateTimeFormat.

pekkis:
Fixes performance issue in timezone plugin by using a cached `Intl.DateTimeFormat` formatter instead of `toLocaleString()`. No functional changes, just an order of magnitudes faster.

Fixes #1236 .

### DIFF
--- a/src/plugin/timezone/index.js
+++ b/src/plugin/timezone/index.js
@@ -33,6 +33,28 @@ const getDateTimeFormat = (timezone, options = {}) => {
   return dtf
 }
 
+const localeStringifierCache = {}
+const getLocaleStringifier = (timezone) => {
+  const localeStringifier = localeStringifierCache[timezone]
+  if (localeStringifier) {
+    return localeStringifier
+  }
+
+  const newLocaleStringifier = new Intl.DateTimeFormat('en-US', {
+    timeZone: timezone,
+    day: 'numeric',
+    month: 'numeric',
+    year: 'numeric',
+    hour: 'numeric',
+    minute: '2-digit',
+    second: '2-digit'
+  })
+
+  localeStringifierCache[timezone] = newLocaleStringifier
+  return newLocaleStringifier
+}
+
+
 export default (o, c, d) => {
   let defaultTimezone
 
@@ -95,7 +117,7 @@ export default (o, c, d) => {
   proto.tz = function (timezone = defaultTimezone, keepLocalTime) {
     const oldOffset = this.utcOffset()
     const date = this.toDate()
-    const target = date.toLocaleString('en-US', { timeZone: timezone })
+    const target = getLocaleStringifier(timezone).format(date)
     const diff = Math.round((date - new Date(target)) / 1000 / 60)
     let ins = d(target).$set(MS, this.$ms)
       .utcOffset((-Math.round(date.getTimezoneOffset() / 15) * 15) - diff, true)


src/plugin/timezone/index.js
```diff
@@ -33,6 +33,28 @@ const getDateTimeFormat = (timezone, options = {}) => {
...
+const localeStringifierCache = {}
+const getLocaleStringifier = (timezone) => {
+  const localeStringifier = localeStringifierCache[timezone]
+  if (localeStringifier) {
+    return localeStringifier
+  }
+
+  const newLocaleStringifier = new Intl.DateTimeFormat('en-US', {
+    timeZone: timezone,
+    day: 'numeric',
```
iamkun, line 45: should we leave these configs to default?

pekkis: I simply configured it with trial-and-error until it 1) worked and 2) produced the same results that the old code did while doing some 10000 iterations on random dates and benchmarking the performance. I lack expertise and context here to judge this any further.

iamkun: Thanks, if so, share the same logic with  getDateTimeFormat ?

pekkis: I of course tried reusing the `getDateTimeFormat()` as-is before trying anything else, but everything crashed and burned. Then I thought of reusing *some* of the same logic and caching with timezone, format options and options, but decided against that because it would just have made it more complex.

pekkis: I'm not 100% sure whether my message is easily understood, so I will try to be clearer.

It was meant to be a question. Would you prefer me to refactor this to share logic between the two caching functions? I do not think that it will make the code clearer or much shorter in this case, so I will sit on this until further notice.

Verde19: Hi. I'm not too familiar with Github so please forgive me if this is not the best place to post this.

I'm using dayjs for a website and I've been making some changes to it recently, specifically adding timezones to some functions. With the latest addition, I've noticed a drastic performance issue.

I came across this thread, but I realized this solution hasn't been applied to the production code I have in my node_modules/dayjs/plugin/timezone.js file (version 1.11.4). So I decided to give it a try on my own:

Simply adding the getLocaleStringifier function, as is, on top of the code and editing "const target" to use it, I noticed a huge **performance gain of about 10 times**. I use dayjs.tz() several times in a loop, in nested functions and within some ES6 array methods, so my code may not be very efficient. Regardless, this loop execution time went down from more than 5 seconds to less than 600 ms.

I hope this helps and this improvement reaches the live dayjs version soon. Find also attached two screenshots from before and after the change.

image

image

logeyG:
Hi all, can I ask why this PR was closed? We ran into a similar issue with using .tz() in a nested loop. It seems as though the solution is straightforward. I'm happy to own this and open up another PR if necessary.

pekkis:
> Hi all, can I ask why this PR was closed? We ran into a similar issue with using .tz() in a nested loop. It seems as though the solution is straightforward. I'm happy to own this and open up another PR if necessary.

I closed it after a couple months or so when it did not move and the whole issue became meaningless via the issue not moving and me coming to a conclusion that it is wisest to seek out an alternative solution and I don't have the extra energy to keep following this.

I think the code itself is fine and solves the issue, though, as far as I trust my limited sense of context within the greater dayjs library. Feel free to grab it and make a new PR!

logeyG:
Thanks @pekkis, just opened up a new PR here: https://github.com/iamkun/dayjs/pull/2019

Appreciate your quick response!

src/plugin/timezone/index.js
defghy: image
It seems this 2 function not generate same format;

In my case, getLocaleStringifier worked fine, use getDateTimeFormat make date not correct

Issue #1852, opened 2022-04-11
Output is NaN for a specific date format


nikhil-kumar-160:
**Describe the bug**
When I try to use dayjs with customParseFormat to parse a string like  "20 Nov, 2022" to a date object, It is not able to parse it. I get NaN in return.

Code 
` dayjs('20 Nov, 2022', DD MMM, YYYY').toDate()`

If i remove the comma, present after the Month name, It works fine.

**Expected behavior**
It should return a date object.

I am using it in a React Native Project.

**Information**
 - Day.js Version = v1.11.0
 - OS: =  iOS
 - Browser [e.g. chrome 62]
 - Time zone: [e.g. GMT+05:30 IST (Indian Standard Time)]

BePo65:
I could tack the problem down: the comma is not a character that is identified as a separator in customParseFormat plugin.

@iamkun is there a special reason, why comma (or other special characters) are not defined as 'separator' characters in `formattingTokens`? And why the dot character defined in `formattingTokens` as separator character is not included in the const `matchWord` used to parse words (like the name of a month)?

I can easily create a pr for this issue; it is only to add the comma to the 2 constants mentioned above.
I also added some tests to 'customParseFormat.test.js' to identify the problem and to verify the solution.

PR #1913, opened 2022-05-27
Fix issue #1852 - comma not recognized as a separator character

BePo65:
The comma is a very common word separator used in dates with short or long month names (see issue #1852); so it should be part of the characters, recognized as word separators.

When fixing this issue, I noticed that the characters used as word separators in 'formattingTokens' (split format string into tokens) and 'matchWord' (find separators between date parts) differ: the underscore was missing in 'formattingTokens' - fixed that too.


iamkun:
Thanks.

We should `comma` as a separator

Issue #1596, opened 2021-08-03
Provide way to specify multpile formats when creating a date in UTC.


DavidRigglemanININ:
If trying to parse a date in which you also need to enable it validates correctly (i.e. using strict mode), you can do so with the following
```
dayjs(it, ["YYYY-MM-DDTHH:mm:ss.SSS[Z]", "YYYY-MM-DDTHH:mm:ss[Z]"], true);
```
However, if using the UTC plugin, it doesn't seem to allow specifying multiple formats so I can only do 
```
const date = dayjs.utc(it, "YYYY-MM-DDTHH:mm:ss.SSS[Z]", true);
```
It would be great the behavior of utc could be extended to support multiple formats as well. Or if anyone knows a workaround to get strict validation and UTC both working together, I'd love to hear it.

BePo65:
Think we should change the line 18 in `dayjs/types/plugin/utc.d.ts` from
```
export function utc(config?: ConfigType, format?: string, strict?: boolean): Dayjs
```
to:
```
export function utc(config?: ConfigType, format?: OptionType, locale?: string, strict?: boolean): Dayjs
export function utc(config?: ConfigType, format?: OptionType, strict?: boolean): Dayjs
```

DavidRigglemanININ:
Sorry for not responding sooner as I completely missed the notification about your message. 

Anyhow, I don't think that proposed solution would help. The issue is that I need to be able to specify multiple formats, meaning that it needs to support a list (i.e. array type). The issue isn't about missing the ability to specify the locale (although others may find that helpful).

nathansolidatus:
I have the same issue here, it'd be really nice to have the same parameters as the `dayjs()` methods.

This is my workaround : 
```
const day = dayjs(dateString, SupportedDateFormats, true)

const offset = day.utcOffset()
const res = day.utc().add(offset, 'minutes')
```

BePo65:
I have traced down the issue: the problem is caused by the 'strict' parameter.
Without it, my tests run fine.
But with 'strict=true', dayjs fails to parse the input string, as long as the offset of the input string does not by chance happen to meet the offset of your current time zone.

The plugin customParseFormat tries to validate that the format of the input string has exactly the same format as one of the format strings by formatting the parsed date using the given format parameters and requires the result to match the input string.

To solve the issue, the line `if (isStrict && date != this.format(format)) {` must be replaced by a better comparison.
[...]

Issue #1902, opened 2022-05-16
Incorrect date parsing when format is 'MMM-YY'


mpycio:
**Describe the bug**
'Apr-22' is getting parsed as 22/04/2001, despite providing parse format as 'MMM-YY', https://jsfiddle.net/Lhv5fpyw/4/

**Expected behavior**
Date should be parsed correctly as 01/04/2022.

**Information**
 - Day.js Version  1.11.2
 - OS: OSX
 - Browser : Chrome 101
 - Time zone: BST

astoilkov:
I have the same issue. Posting my dummy workaround here:

```ts
function fixDayjs(input: string): Dayjs | undefined {
  const date = dayjs(input)
  
  if (date.isValid()) {
    // https://github.com/iamkun/dayjs/issues/1902
    // https://github.com/iamkun/dayjs/issues/1251
    if (date.year() === 2001 && !input.split(' ').includes('2001')) {
      return date.set('year', dayjs().year())
    }

    return date
  }

  return undefined
}
```

joeyipjoeyip:
I have another incorrect parsing bug

**Describe the bug**
Converting "173000123" with format "HHmmssSSS" to get Unix timestamp in milliseconds returns a negative value, which when parsed back, returns "Mon Dec 12 1729 03:00:00 GMT+0736", https://jsfiddle.net/joeyip/8t2jd1bm/

**Expected behavior**
The time part should be 17:30:00

**Information**
- Day.js Version 1.11.2
- OS: Windows 10
- Browser : Firefox 91.9.1esr
- Time zone: GMT+8

BePo65:
Would be fixed with pr #1914

Issue #1750, opened 2021-12-26
CustomParseFormat does't use a timezone


tenorok:
**Describe the bug**

When using input format as array then output format does't take specified timezone.

```js
import dayjs from 'dayjs';
import utc from 'dayjs/plugin/utc';
import customParseFormat from 'dayjs/plugin/customParseFormat';
import timezone from 'dayjs/plugin/timezone';

dayjs.extend(utc);
dayjs.extend(timezone);
dayjs.extend(customParseFormat);
dayjs.tz.setDefault('Europe/Moscow');

// Incorrect, output:
// 31.05.2021 21:00
console.log(
    dayjs
        .tz('01.06.2021 00:00', ['DD.MM.YYYY HH:mm'], 'Europe/Moscow')
        .format('DD.MM.YYYY HH:mm'),
);
```

**Expected behavior**

When using just a string for input format the timezone takes correctly.
```js
// Correct, output
// 01.06.2021 00:00
console.log(
    dayjs
        .tz('01.06.2021 00:00', 'DD.MM.YYYY HH:mm', 'Europe/Moscow')
        .format('DD.MM.YYYY HH:mm'),
);
```

**Information**
 - Day.js Version: v1.10.7
 - OS: macOS 12.0.1 (21A559)
 - Browser: NodeJS v15.14.0
 - Time zone: GMT+03:00

**Links**
* Doc about format: https://day.js.org/docs/en/parse/string-format
* PR with supports of array for format: https://github.com/iamkun/dayjs/pull/906/files

Haixing-Hu:
yes, I met the same problem. Any fix?

BePo65:
I could reproduce the issue even with the current version (v1.11.2).
I am preparing a PR for issue #1596, which solves even this problem. So I expect that this problem will be solved not too far in the future ;-)

usd-yamazaki:
https://github.com/iamkun/dayjs/issues/944
https://github.com/iamkun/dayjs/pull/1914
I hope that these issues, including this one, will be resolved.

Issue #1734, opened 2021-12-04
format by milliseconds， 'SS' or 'S' is not work


Suntgr:
dayjs().format('mm-ss S')     ->  08-40 S
dayjs().format('mm-ss SS')   ->  08-40 SS
dayjs().format('mm-ss SSS') ->  08-40 047

if format by  ‘S’ and 'SSS' is not work, why it is in list of all available parsing tokens [https://dayjs.gitee.io/docs/en/parse/string-format#list-of-all-available-parsing-tokens](url)

I'm not sure if it's a bug

wizount:
So with me

ephys:
I'm hoping support for `S` & `SS` can be considered, as we've encountered this issue while trying to fix our date formatting in https://github.com/sequelize/sequelize/pull/15838 😅

WikiRik:
This seems to be a duplicate of #1331 and might be fixed by #1914

PR #1914, opened 2022-05-27
Fix issue #1596 - Provide way to specify multiple formats when creating a date in UTC

BePo65:
The issue #1596 shows that the plugin 'customParseFormat' does not give a valid date, when used with utc parsing dates with offset.
I identified 2 interdependant reasons for this:

1. customParseFormat.parse gives wrong arguments to dayjs when called by utc plugin with an array of formats.  
Details:  
calling `utc(...)` for parsing the utc plugin creates a cfg object with a correct 'utc' property ('true'); but the recursion in the parse function of 'customParseFormat' just transfers all the arguments to the call to 'dayjs' (`const result = d.apply(this, args)`). But the args property must be a cfg object.

3. parsing a date with offset in strict mode always returns 'Invalid Date'.  
Details:  
`(isStrict && date != this.format(format))` is an elegant way to find out, if the parsed date respects all the separators as defined in the format. But it fails, as soon as an offest comes into play; in this case the result of formatting the parsed date corresponds to the given format, but the used offset depends an the current local timezone and is not necessarily the same as the offset in the date to be parsed (e.g. '11:00+01' is the same as '12:00+02').
As a result I had to change the way we identify a 'strict' match (including the handling of overflows in strict mode).

I did not squash the single commits, so that it is easier to analyze the single modifications.
And I added tests for all affected scenarios.


BePo65:
And when we are at it, perhaps we should also add a list of the deviations from `moment` to the documentation:

| title | parameters | dayjs | moment |
|-------|------------|-------|--------|
| invalid date with overflow | ('35/22/2010 99:88:77', 'DD-MM-YYYY HH:mm:ss') | '08-11-2011 04:29:17' | 'Invalid date' |
| invalid date with overflow, strict | ('35/22/2010 99:88:77', 'DD-MM-YYYY HH:mm:ss', true) | 'Invalid Date' | 'Invalid date' |
| '0' day or month (using default values) | ('1970-00-00', 'YYYY-MM-DD') | '1970-01-01' | 'Invalid date' |
| '0' day or month (using default values), strict | ('1970-00-00', 'YYYY-MM-DD', true) | 'Invalid Date' | 'Invalid date' |
| date not matching format | ('10/12/2014', 'YYYY-MM-DD') | '01-01-2014' | '12-20-2010' |
| date not matching format, strict | ('10/12/2014', 'YYYY-MM-DD', true) | 'Invalid Date' | 'Invalid date' |
| first match vs. longest match | ('2012-05-28 10:21:15', ['YYYY', 'YYYY-MM-DD', 'YYYY-MM-DD HH:mm:ss']) | '2012-01-01 00:00:00' | '2012-05-28 10:21:15' |
| first match vs. longest match, strict | ('2012-05-28 10:21:15', ['YYYY', 'YYYY-MM-DD', 'YYYY-MM-DD HH:mm:ss'], true) | '2012-05-28 10:21:15' | '2012-05-28 10:21:15' |

BePo65:
Solves issue #1616 too.
Solves issue #1750 too.
Solves issue #1902 too.

BePo65:
I finally found the "older" pr #929 that solves at least the issue with the offset in strict mode, when the format contains a `Z` token.

* the pr #929 takes the approach of improving the way to identify a 'strict match' (`'date != this.format(format)'`) while
* this pr tries to identify 'strictness' while parsing the date parts; this method allows to fix issues with parsing that occurred e.g. when the sequence of date parts did not match the sequence of format token e.a.

So now you have the choice :-)

@iamkun by now there are a lot of open issues and quite some pr. Is there a way I could help you to reduce the number of open issues and pr?

src/index.js
```diff
@@ -38,9 +38,15 @@ const dayjs = function (date, c) {
...
-  const cfg = typeof c === 'object' ? c : {}
+  let cfg = {}
+  if ((typeof c === 'object') && !Array.isArray(c)) {
+    // clone c (contains cfg object)
+    cfg = JSON.parse(JSON.stringify(c))
+  }
   cfg.date = date
-  cfg.args = arguments// eslint-disable-line prefer-rest-params
+  if (cfg.args === undefined) {
+    cfg.args = [...arguments] // eslint-disable-line prefer-rest-params
```
iamkun, line 48: it's better to avoid using the spread operator, cause it will increase the final compiled js bundle size dramatically.

BePo65: I replaced the spread operator with Array.prototype.slice.  
Searching for other files using the spread operator I found a few, but there the solution is not so simple, as here the spread operator is used to extend an object.

Perhaps it makes sense to think about what systems to support. I opened issue #1925 to discuss that topic.

iamkun:
thanks, please allow me some time to go through this cool PR

BePo65:
You've got all the time in the world; if my documentation of the changes needs improvement, feel free to ask.

src/plugin/customParseFormat/utils.js
```diff
@@ -0,0 +1,26 @@
...
+}
+
+// code duplication as we cannot use 'isLeapYear' plugin here;
+// 'isLeapYear' is only working with Dayjs objects
+function isLeapYear(year) {
+  return ((year % 4 === 0) && (year % 100 !== 0)) || (year % 400 === 0)
+}
+
+// eslint-disable-next-line import/prefer-default-export
+export const daysInMonth = (year, month) => {
```
iamkun, line 13: we can use the built-in function 
```
dayjs(year + '-' + month).daysInMonth()
```

WikiRik:
@BePo65 any chance you could update this PR soon? Or would you be fine if I tried to update it in a new PR?

src/plugin/customParseFormat/utils.js
BePo65: sorry for the late reply, but somehow I missed this comment :-(

Of course we could use `dayjs(year + '-' + month).daysInMonth()` instead of my 'auxiliary' method.

But the problem is that we do not have the dayjs object available to the `checkOverflow()` function.

Therefore I created the `daysInMonth` method in the customParseFormat/utils file.  
As an additional bonus, this function runs faster than creating a new dayjs object (to be exact, more than 1 instance is created for getting daysInMonth).

What do you think about this?

BePo65:
> @BePo65 any chance you could update this PR soon? Or would you be fine if I tried to update it in a new PR?

I added 2 tests for your issue #1734 and found out that I missed a comment from @iamkun - replied to this comment too.

So now I hope we have everything in place so that the review of this pr can be completed. Let's hope the best 😄

le0pard:
Is any help needed with this PR?

WikiRik:
@le0pard you can review this PR, but mostly we're waiting on a review by @iamkun

CHANGELOG.md
```diff
@@ -1,3 +1,11 @@
+## [1.11.7](https://github.com/iamkun/dayjs/compare/v1.11.6...v1.11.7) (2022-12-06)
```
WikiRik, line 1: I think the changes to CHANGELOG.md should be reverted as they're not in scope for this PR

BePo65: This line comes from dayjs 1.11.7 - the base of my pr and is autogenerated by the dayjs build system (I suppose). The line itself is the link to the git changes (as in any versions before). But you knew all this before 😄 

So I don't know why this is a "changed file" here, as I didn't change it in my pr.
in the cuurent master branch (1.11.9) the line looks unchanged and AFAIK github should / could rebase a pr automatically on merge.
So what should I change in the changelog.md file?

WikiRik: GitHub says these lines are added by your PR. Maybe if you rebase this PR to `dev` these changes would be removed since they are then already implemented

BePo65: Sounds good, as this pr will merge to dev.
But.... somehow it looks like dev is not up to date, For me it looks like the dev branch somehow is stuck on v1.11.6, while master moved on to 1.11.9.

So I think / hope that all this mess will disappear, when this pr is merged into master (if this will ever happen :-)

WikiRik: Did you check out the `dev` from your fork by any chance? It seems that `dev` is the default branch of this repo and things have been merged to it a few days ago

BePo65: oh man this is a mess with the branches in this repo:
master contains the line to v1.11.9 while dev is not up to date with master.

I think I will rebase the pr to dev as soon as @iamkun answers to my comment on his remark to 'utils.js' below.

PR #1920, opened 2022-06-02
feat: add diff360 plugin

ERPedersen:
This plugin returns the number of days between two dates based on a 360-day calendar (twelve 30-day months),

The 360-day calendar is a method of measuring durations used in financial markets, in computer models, in ancient literature, and in prophetic literary genres.

It uses the following algorithm to calculate the difference (Turn off dark mode to see):

image


iamkun:
Thanks!

In order to keep the same API with moment.js, we intend to not accept custom plugins that have different APIs with moment.js.

It will be better to publish it as an npm package and manage it yourself.

PR #1926, opened 2022-06-05
Fix configuration for eslint (linting subdirectories; activating jest rules

BePo65:
The configuration for eslint had 2 minor topics:

- lintingwas only activated for directories 'src', 'test' and 'build' but not for the subdirectories of these directories
- jest rules for eslint were loaded (property "plugins" in .eslintrc.json) but not activated

Fixes these topics and removed the newly identified warnings.

### DIFF
--- a/test/plugin/duration.test.js
+++ b/test/plugin/duration.test.js
@@ -139,16 +139,7 @@ describe('Clone', () => {
 })
 
 describe('Milliseconds', () => {
-  expect(dayjs.duration(500).milliseconds()).toBe(500)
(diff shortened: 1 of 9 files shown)


test/plugin/duration.test.js
```diff
@@ -139,16 +139,7 @@ describe('Clone', () => {
 })
 
 describe('Milliseconds', () => {
-  expect(dayjs.duration(500).milliseconds()).toBe(500)
```
iamkun, line 142: this PR is a little confused to me. 

Is these unit cases modification not relevant to this PR?

PR #1932, opened 2022-06-07
fix: wrong ordinal for fr local

R3tep:
This pull request fix a wrong ordinal definition for fr local.

### DIFF
--- a/src/locale/fr.js
+++ b/src/locale/fr.js
@@ -34,7 +34,7 @@ const locale = {
     yy: '%d ans'
   },
   ordinal: (n) => {
-    const o = n === 1 ? 'er' : ''
+    const o = n === 1 ? 'er' : 'e'
     return `${n}${o}`
   }
 }


iamkun:
Is fr locale's ordinal different under different format token? 

ref https://github.com/moment/moment/blob/develop/src/locale/fr.js#L82

iamkun:
Thanks, let's get this merged

xavier-villelegier:
Hey @R3tep, why this change ? In french when it's the 1st of July for exemple we say "1er juillet 2022", but for the 2nd, we never write "2e juillet 2022" or "3e juillet 2022" ? What was your use case ?

cc @iamkun

CBD-iBiZa:
> Hey @R3tep, why this change ? In french when it's the 1st of July for exemple we say "1er juillet 2022", but for the 2nd, we never write "2e juillet 2022" or "3e juillet 2022". This might be correct mathematically speaking, but not in term of dates ? What was your use case ?
> 
> cc @iamkun

For week of year, with ordinal ([wo format](https://day.js.org/docs/en/plugin/advanced-format)), the output in french is not correct.
- 1st (week) => 1re (semaine), not 1er (semaine)
- 2nd (week) => 2e (semaine), not 2 (semaine)
- 3rd (week) => 3e (semaine), not 3 (semaine)
- ...

It was perfectly ok [in Moment](https://github.com/moment/moment/blob/develop/src/locale/fr.js#L98), not with DayJS+WeekOfYear plugin.

Issue #1923, opened 2022-06-03
Incorrect parsing of fractional seconds in ISO-8601 when no timezone is specified


benj-dobs:
**Describe the bug**
DayJS thinks that the time `'00:00:00.5'` is five milliseconds after midnight, but it should be 500 milliseconds after midnight.

Reproduction:

```ts
const dayjs = require('dayjs');
const dateWithoutSpecifiedTimezone = dayjs('2000-01-01T00:00:00.5');
const dateInUTC = dayjs('2000-01-01T00:00:00.5Z');

dateInUTC.toISOString(); // '2000-01-01T00:00:00.500Z' - Correct
dateWithoutSpecifiedTimezone.toISOString(); // '2000-01-01T00:00:00.005Z' - Incorrect - should be '2000-01-01T00:00:00.500Z'
```

**Expected behavior**
It should interpret the number of seconds  `ss.SSS` as a decimal fraction, no matter how many decimal places are given, regardless of whether there is a timezone specified

**Information**
 - Day.js Version: 1.11.2
 - OS: macOS 12 
 - Browser: Bug reproduced in Safari and Node.js
 - Time zone: BST (UTC+1)

BePo65:
In this regard dayjs is consistent to what moment does. If we look at the documentation of moment we see:
>Note that the number of S characters provided is only relevant when parsing in strict mode. In standard mode, S, SS, SSS, SSSS are all equivalent, and interpreted as fractions of a second. For example, .12 is always 120 milliseconds, passing SS will not cause it to be interpreted as 12 milliseconds.

So IMHO the result of your example is correct, as dayjs tries to be as close to moment as possible (without sacrificing the size).

benj-dobs:
Amazing, thank you!

PR #1936, opened 2022-06-09
Fix issue1923 - parsing of milliseconds like moment does

BePo65:
With this fix, milliseconds in ISO date string are parsed like moment does (digits get padded to re right with '0' to 3 places).

Solves issue #1923.

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -66,7 +66,7 @@ const parseDate = (cfg) => {
     const d = date.match(C.REGEX_PARSE)
     if (d) {
       const m = d[2] - 1 || 0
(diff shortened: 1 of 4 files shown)


benj-dobs:
I'm not that familiar with this codebase but this looks like it does the job - thank you!

jimmyltsinn:
I think the issue could cause problem on many system that compare time with ms precision. Our team has locally patched according to this PR and so far so good. Hope this PR can be merged soon 🎉

Deadid:
Any chance to have this merged? We've faced this issue multiple times and it would be really helpful to get this addressed

PR #1944, opened 2022-06-14
feat: diff plugin

lamartire:
Adds `diff` plugin to get difference between two dates in given units.

Usage example:

```js
dayjs('2022-06-11').diff('day', '2022-06-30') // 19
```

The functionality is very useful when you need to determine specific time gap and it isn't enough to have `23` limit for hours or `59` for minutes:

```js
dayjs('2022-06-11 01:00').diff('day', '2022-06-11 03:00') // 120
```

The plugin inspired by `date-fns` `differenceInMilliseconds`, `differenceInSeconds` etc. functions.

I haven't found where I should document the plugin. @iamkun could you please tell me, where is the documentation sources are located in?

### DIFF
--- a/src/plugin/diff/index.js
+++ b/src/plugin/diff/index.js
@@ -1,0 +1,40 @@
+import {
+  MILLISECONDS_A_DAY,
+  MILLISECONDS_A_HOUR,
+  MILLISECONDS_A_MINUTE,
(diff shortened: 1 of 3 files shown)


lamartire:
Closed due `dayjs` already has `diff` method

PR #1949, opened 2022-06-16
fix: fix type error that can’t use "isoWeek" as type in isSameOrBefor…

Creabine:
fix type error that can’t use "isoWeek" as type in isSameOrBefore/isSameOrAfter plugin

In isSameOrBefore/isSameOrAfter plugin, they just use `isSame`, `isAfter` and `isBefore` function,  these functions all use `ISOUnitType` as unit type, but isSameOrBefore/isSameOrAfter plugin use `OpUnitType` as unit type. So they won't allowed `isoWeek` as unit, and that is wrong.

Please review my code and release it ASAP.  Thank you so much.


Creabine:
@iamkun could you plz tell me is there anything wrong? why is there no feedback?

PR #1966, opened 2022-06-28
Update italian and italian ( switzerland ) locale

Wit3:
- Added first letter capitalize ( like english )


Wit3:
In #458 replace capitalize with non first capitalize 😕

PR #1975, opened 2022-07-05
Convert more plugins and locales for Dayjs 2.0

PascalHonegger:
Hello, we're using dayjs in our angular app and would appreciate the ESM version, which is why I'm trying to help push version 2.0 further 😉 .

I converted two easy plugins and tried to convert the german language, I'm looking for feedback before I continue with other plugins to ensure I don't cause any conflicts.

@sxzz I'm struggling to make the `relativeTime` work with the newly provided types, the implementation for formatting also seems missing at this point. Do you have any inputs in that regard?


sxzz:
I'm migrating `relativeTime` plugin to v2 branch.

PascalHonegger:
I've put my locale migration on hold until you've migrated `relativeTime`, I also added the unit tests and cleaned up the code so both plugins (`isSameOrAfter` and `isSameOrBefore`) are ready and could be merged if they look good to you, I'll look into more plugins once this MR is merged

PascalHonegger:
The failed unit test has nothing to do with my changes, looks like a time dependent unit test. Will you look into it or shall I?

PR #1983, opened 2022-07-11
test: add tests for plugin isToday

BePo65:
feat: add tests for plugin `isToday` to dayjs 2.0.

Unify layout of plugin code - all plugins should have the same base code structure.
Fix changes recommended by eslint.

### DIFF
--- a/src/plugin/isToday.ts
+++ b/src/plugin/isToday.ts
@@ -1,15 +1,15 @@
 import type { Dayjs, Plugin } from '..'
 
-export type isTodayType = () => boolean
+export type IsToday = () => boolean
 
 declare module '../types' {
   export interface Extend {
-    isToday: isTodayType
+    isToday: IsToday
   }
 }
 
 const plugin: Plugin = (cls, fn) => {
-  const isToday: isTodayType = function (this: Dayjs): boolean {
+  const isToday: IsToday = function (this: Dayjs): boolean {
     const comparisonTemplate = 'YYYY-MM-DD'
     const now = fn()
     return this.format(comparisonTemplate) === now.format(comparisonTemplate)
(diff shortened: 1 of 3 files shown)


src/plugin/isToday.ts
```diff
@@ -1,19 +1,20 @@
 import type { Dayjs, Plugin } from '..'
 
-export type IsToday = (this: Dayjs) => boolean
+export type isTodayType = () => boolean
```
sxzz, line 3: IMHO, the name of typing should be `IsToday`.

BePo65: As this is an interface, you are of right - my fault.

src/plugin/isToday.ts
```diff
@@ -1,6 +1,6 @@
 import type { Dayjs, Plugin } from '..'
 
-export type IsToday = (this: Dayjs) => boolean
```
sxzz, line 3: Why do you remove this here?

BePo65: According to the typescript [documentation](https://www.typescriptlang.org/docs/handbook/2/functions.html#declaring-this-in-a-function) the context (`this`) is a parameter that is 'implicit' added to a class method signature.

Therefore we do not have to add it to the type definition (and IMO we should not add it to the type definition, as we will never call this method with a context as a first parameter and adding it to the type definition will probably produce more user questions than help the user understand the method).

On the other hand, the `this` property in the function definition is required, as otherwise the context would be of the 'any' type (as we add this method programmatically to the Dayjs class) - which is not what we want.

sxzz: OK make sense.

sxzz (approved):

PR #1984, opened 2022-07-11
test: add tests for plugin toArray

BePo65:
feat: add tests for plugin `toArray` to dayjs 2.0.

Unify layout of plugin code - all plugins should have the same base code structure.

### DIFF
--- a/src/plugin/toArray.ts
+++ b/src/plugin/toArray.ts
@@ -1,24 +1,36 @@
 import type { Dayjs, Plugin } from '..'
 
+export type toArrayType = () => [
+  number,
+  number,
+  number,
+  number,
+  number,
+  number,
+  number
+]
+
 declare module '../types' {
   export interface Extend {
-    toArray: typeof toArray
+    toArray: toArrayType
   }
 }
 
-function toArray(
-  this: Dayjs
-): [number, number, number, number, number, number, number] {
-  return [
-    this.year(),
-    this.month(),
-    this.date(),
-    this.hour(),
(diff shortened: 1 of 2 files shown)


src/plugin/toArray.ts
```diff
@@ -1,24 +1,36 @@
...
+    return [
+      this.year(),
+      this.month(),
+      this.date(),
+      this.hour(),
+      this.minute(),
+      this.second(),
+      this.millisecond(),
+    ]
+  }
```
sxzz, line 32: The `toArray` function can be extracted to the top level.

BePo65: As `toArray` doesn't use any properties of the plugin function (`const plugin: Plugin = (cls) => { ...`) that is technically correct.  
But my point is that all plugins should be based on a common template and follow some rather strict rules to ensure long term maintainability of dayjs, Examples of these rules would be
+ all plugin code must live within the `const plugin: Plugin = (cls, fn) => { ...` definition
+ all type names must start with an uppercase letter (ok, I have to fix this here)
+ all exported functions must be defined as `const` to ensure immutability

sxzz: I agree with some of your points, but I still think there is no great need to maintain formal consistency, which would essentially limit the flexibility of writing plugins. Besides, it would be better for performance and code reading to hoist `toArray`.

BePo65: So I will 'reset' the layout.

PR #1989, opened 2022-07-15
feat: time based greeting

Dev-Rise:
- Time-based greeting

** I didn't add it for all available languages if it is necessary I will do it.
** I create this PR just to be sure changes are made correctly, before I spend time filling greeting rules in all languages.


iamkun:
Thanks for this lovely PR.

It is handy in some cases.

Maybe, it's better to ship this greet API as a separate plugin, not via dayjs core lib.

iamkun:
> > Thanks for this lovely PR.
> > It is handy in some cases.
> > Maybe, it's better to ship this greet API as a separate plugin, not via dayjs core lib.
> 
> So should I publish it by myself as a separate plugin "dayjs-greet"? If I understand correctly it should be done like for example https://www.npmjs.com/package/dayjs-parser Am I right?

IMO, this is the right way to publish dayjs plugins.

Dev-Rise:
Ok, I created: https://www.npmjs.com/package/dayjs-greet
I think we can close this PR

PR #2010, opened 2022-08-04
fix: wrong ordinal for french locale

ctessier:
In french, when writing dates, there is no ordinal for numbers over 1.

- lundi 1er août 2022 :heavy_check_mark: 
- jeudi 11 août 2022 :heavy_check_mark: 
- jeudi 11e août 2022 :x: 

In other contexts, this could be correct, but not for dates.
See for instance on this French website of the government : https://www.service-public.fr/particuliers/actualites

### DIFF
--- a/src/locale/fr.js
+++ b/src/locale/fr.js
@@ -34,7 +34,7 @@ const locale = {
     yy: '%d ans'
   },
   ordinal: (n) => {
-    const o = n === 1 ? 'er' : 'e'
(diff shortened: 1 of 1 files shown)


xavier-villelegier:
For more context: this is basically a revert of  the regression introduced by this PR https://github.com/iamkun/dayjs/pull/1932  in the v1.11.4

iamkun:
Thanks, let's get it to revert to the original one

PR #2011, opened 2022-08-04
fix: ordinal for nl not working

ctessier:
In dutch locale, the ordinal doesn't show properly, some letters get interpreted.

Before the fix: :x:
```javascript
dayjs().locale('nl').format('Do'); // 44e
dayjs('2022-08-23').locale('nl').format('Do'); // 230te
```

After the fix: :heavy_check_mark: 
```javascript
dayjs().locale('nl').format('Do'); // 4de
dayjs('2022-08-23').locale('nl').format('Do'); // 23ste
```


iamkun:
Thanks

PR #2015, opened 2022-08-05
fix types for utcOffset

capaj:


iamkun:
This dependent on [UTC ](https://day.js.org/docs/en/plugin/utc)plugin to work, and should be added to the utc plugin type file

capaj:
yes, sorry, realized this later, forgot to close

Issue #1905, opened 2022-05-18
`dayjs.tz` incorrectly handles offsets containing minutes and seconds


tbaliukynas:
**Describe the bug**
Creating a date with a timezone offset that contains minutes and seconds is not properly handled by `dayjs.tz` function. Offset seems to be rounded to a nearest hour and newly created `dayjs.Dayjs` object provides incorrect time information.

For example, timezone in Kyiv up until 1924 had an offset of UTC+02:02:04 (https://www.timeanddate.com/time/zone/ukraine/kyiv):
![image](https://user-images.githubusercontent.com/36598966/169157824-f5659ffd-7d1c-452d-ac21-6543d31206a5.png)

In Moment.js library this edge case is handled correctly. Here is a simple test that demonstrates how `dayjs.tz` should behave:
```ts
describe('tz', () => {
  describe('given a timezone with UTC+02:02:04 offset', () => {
    // Intl.DateTimeFormat().resolvedOptions().timeZone;     -> "Europe/Kiev" 
    // new Date("1900-06-01T12:00:00Z").getTimezoneOffset(); -> -122.06666666666666
    // new Date("1900-06-01T12:00:00Z").valueOf();           -> -2195899200000 

    test('handles it correctly', () => {
      const kiev = dayjs('1900-06-01T12:00:00Z').tz('Europe/Kiev');
      const Mkiev = moment('1900-06-01T12:00:00Z').tz('Europe/Kiev');
      expect(kiev.format()).toBe('1900-06-01T14:02:04+02:02');
      expect(kiev.format()).toBe(Mkiev.format());
      expect(kiev.valueOf()).toBe(-2195899200000);
      expect(kiev.valueOf()).toBe(Mkiev.valueOf());
      expect(kiev.utcOffset()).toBe(122.06666666666666);
      expect(kiev.utcOffset()).toBe(Mkiev.utcOffset());
      expect(kiev.utc().format()).toBe("1900-06-01T12:00:00Z");
      expect(kiev.utc().format()).toBe(Mkiev.utc().format());
    });
  });
});
```

**Information**
 - Day.js Version: 1.11.2
 - OS: Windows 10, version 21H2
 - Browser: Node.js 16.15.0
 - Time zone: [Europe/Vilnius](https://en.wikipedia.org/wiki/Time_in_Lithuania) (UTC offset: +02:00; DST: +03:00)

tbaliukynas:
[...]

Issue #1996, opened 2022-07-21
Old dates have weird minute offsets in utc


irgipaulius:
**Describe the bug**
Take this as an example:
```js
// loop same day every 11 years for the past 200 years
for (let i = 2022; i > 1800; i -= 11) {
  console.log(dayjs(`12/31/${i}`).utc(true).toDate());
}
```js

**Expected behavior**
```
2022-12-31T00:00:00.000Z
2011-12-31T00:00:00.000Z
2000-12-31T00:00:00.000Z
1989-12-31T00:00:00.000Z
1978-12-31T00:00:00.000Z
1967-12-31T00:00:00.000Z
1956-12-31T00:00:00.000Z
1945-12-31T00:00:00.000Z
1934-12-31T00:00:00.000Z
1923-12-31T00:00:00.000Z
1912-12-31T00:00:00.000Z
1901-12-31T00:00:00.000Z
1890-12-31T00:00:00.000Z
1879-12-31T00:00:00.000Z
1868-12-31T00:00:00.000Z
1857-12-31T00:00:00.000Z
1846-12-31T00:00:00.000Z
1835-12-31T00:00:00.000Z
1824-12-31T00:00:00.000Z
1813-12-31T00:00:00.000Z
1802-12-31T00:00:00.000Z
```

**Actual behavior**
```
2022-12-31T00:00:00.000Z
2011-12-31T00:00:00.000Z
2000-12-31T00:00:00.000Z
1989-12-31T00:00:00.000Z
1978-12-31T00:00:00.000Z
1967-12-31T00:00:00.000Z
1956-12-31T00:00:00.000Z
1945-12-31T00:00:00.000Z
1934-12-31T00:00:00.000Z
1923-12-31T00:00:00.000Z
1912-12-31T00:06:00.000Z   /// what???
1901-12-31T00:06:00.000Z
1890-12-31T00:06:00.000Z
1879-12-31T00:03:44.000Z   /// huh??!
1868-12-31T00:03:44.000Z
1857-12-31T00:03:44.000Z
1846-12-31T00:03:44.000Z
1835-12-31T00:03:44.000Z
1824-12-31T00:03:44.000Z
1813-12-31T00:03:44.000Z
1802-12-31T00:03:44.000Z   /// it goes on like this until 100 AC
```


**Information**
 - Day.js Version ^1.11.4
 - OS: MacOs Big Sur 11.6
 - NodeJS
 - Time zone: [e.g. GMT+03:00 DST]

[...]

PR #2016, opened 2022-08-06
Fix issue #1905 - problems with special timezone offsets

BePo65:
This PR solves problems with timezone offset that do not contain complete or half hours or that contain seconds. There are a few offsets of this kind in historical dates.

It should solve issue #1905 and part of issue #1996.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,15 @@
+## [1.11.4](https://github.com/iamkun/dayjs/compare/v1.11.3...v1.11.4) (2022-07-19)
+
+
+### Bug Fixes
+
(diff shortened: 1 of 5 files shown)


BePo65:
Just a few more informations to the commits of this pr:

* the rounding to the nearest 15min in utcOffset() solves a problem with firefox that just does not exist any more. I tested the code in the current FF, Chrome and node LTS and for me it looks like this bug in FF was fixed a long time ago (somewhere on the way from FF24 to FF103). On the other hand this rounding creates a lot of problems with historical dates, where weird minute offset are not uncommon.
* the second problem appears in timezone offset containing seconds. In such a case dayjs shows a zone string that lookes like `+02:02.026666`. The pr changes this to the correct value of `02:02` just as moment does it.

Both cases do not have an impact on current dates where to my knowledge such offsets do not exist.

Issue #2031, opened 2022-08-20
Parsing localized time string including meridiem with format 'a hh:mm' ignores meridiem.


rodumani:
**Describe the bug**

Parsing localized time string including meridiem with format 'a hh:mm' ignores meridiem.

In Korean, "am" is "오전" and "pm" is "오후". There is no uppercase/lowercase in Korean. 

However, dayjs customParseFormat doesn't recognize the meridiem and only returns hours in am. 

**Expected behavior**

```javascript
import dayjs from 'dayjs';
import 'dayjs/locale/ko.js';

import customParseFormat from 'dayjs/plugin/customParseFormat.js';
dayjs.extend(customParseFormat);

const amStr = '오전 10:00';
const pmStr = '오후 10:00';
const format = 'a h:mm';

const am = dayjs(amStr, format);
console.assert(am.hour() === 10, 'am.hour() === 10');
const pm = dayjs(pmStr, format);
console.assert(pm.hour() === 22, 'pm.hour() === 22');
```

```bash
$ node index.js
Assertion failed: pm.hour() === 22
```

The second assert should be passed, but it fails because it returns the hour as 10.

**Information**
 - Day.js Version: 1.11.5
 - OS: macOS 12.5
 - Browser: Node v16.14.0 / Chrome 104.0.5112.101
 - Time zone: GMT +09:00 KST

BePo65:
The problem stems from the fact that CustomPareFormat tries to implement some kind of a 'meridiemParse' function that uses the 'meridiem' function of the locale file to find out which hour values are considered 'am' and 'pm'. This function supposes that the 'meridiem' function returns the string for 'pm' for anything after '12' and the string for 'am' for hours '12' and before.

But some locales consider noon as 'pm' among them 'ko'.

If we look to the details of meridiem then we will find that noon / midnight is a very problematic case (and that the chinese locales use more verbs for time ranges, not only 'am' and 'pm'). Details can be found e.g. in [wikipdia](https://en.wikipedia.org/wiki/12-hour_clock#Confusion_at_noon_and_midnight).

The question is, if we need some kind of 'meridiemParse' function like [moment](https://momentjs.com/docs/#/customization/am-pm-parsing/) does. @iamkun any opinion on this? I think this would not be too problematic.

sukvvon:
[...]

PR #2032, opened 2022-08-20
Fix parsing localized meridiem always returns am #2031

rodumani:
## Changes
- Fix customParseFormat.meridemMatch function to properly parse localized meridiem
- Add test cases in korean locale

## Describe the bug
Parsing localized time string including meridiem with format 'a hh:mm' ignores meridiem.

In Korean, "am" is "오전" and "pm" is "오후". There is no uppercase/lowercase in Korean.

However, dayjs `customParseFormat` doesn't recognize the meridiem and only returns hours in am.

## Describe fixes
I found that the `customParseFormat.meridemMatch` function iterates from 1 to 24 and check if there is a match. 

However, it sets the `isAfternoon` variable by comparing if the `i` is **greater than 12** which should be **greater than or equal to 12**.

## Expected behavior
```javascript
import dayjs from 'dayjs';
import 'dayjs/locale/ko.js';

import customParseFormat from 'dayjs/plugin/customParseFormat.js';

const amStr = '오전 10:00';
const pmStr = '오후 10:00';
const format = 'a h:mm';

const am = dayjs(amStr, format);
console.assert(am.hour() === 10, 'am.hour() === 10');
const pm = dayjs(pmStr, format);
console.assert(pm.hour() === 22, 'pm.hour() === 22');
```
```bash
$ node index.js
Assertion failed: pm.hour() === 22
```

The second assert should be passed, but it fails because it returns the hour as 10.

I have added a test case of it which was failing before applying the changes of this PR.
```javascript
it('Korean customParseFormat', () => {
  expect(dayjs('2019-01-01').locale('ko').format('YYYY-MM-DD'))
    .toBe(moment('2019-01-01').locale('ko').format('YYYY-MM-DD'))

  expect(dayjs('오전 1:00', 'a h:mm').locale('ko').format('YYYY-MM-DD'))
    .toBe(moment('오전 1:00', 'a h:mm').locale('ko').format('YYYY-MM-DD'))

  expect(dayjs('오후 1:00', 'a h:mm', 'ko').format('a h:mm'))
    .toBe(moment('오후 1:00', 'a h:mm', 'ko').format('a h:mm'))
})
```

### DIFF
--- a/src/plugin/customParseFormat/index.js
+++ b/src/plugin/customParseFormat/index.js
@@ -52,7 +52,7 @@ const meridiemMatch = (input, isLowerCase) => {
     for (let i = 1; i <= 24; i += 1) {
       // todo: fix input === meridiem(i, 0, isLowerCase)
       if (input.indexOf(meridiem(i, 0, isLowerCase)) > -1) {
-        isAfternoon = i > 12
+        isAfternoon = i >= 12
         break
       }
     }
--- a/test/locale/ko.test.js
+++ b/test/locale/ko.test.js
@@ -1,0 +1,57 @@
+import MockDate from 'mockdate'
+import moment from 'moment'
+import dayjs from '../../src'
+import relativeTime from '../../src/plugin/relativeTime'
+import '../../src/locale/ko'
+import customParseFormat from '../../src/plugin/customParseFormat'
+
+dayjs.extend(relativeTime)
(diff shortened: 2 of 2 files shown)


rodumani:
Closes #2031

brandon-swipedon:
@iamkun are we able to get this merged?

PR #2034, opened 2022-08-21
Fixed Kurdish relative time conversion issue

SarkarKurdish:
When using the relative time plugin and transfer a time and use the `ku`(kurdish) locale the translated time to text should look like this

Expected Result
`٢ خولەک لەمەوپێش`

Result
`لەمەوپێش ٢ خولەک`

the issue was just a variable misplacement.


SarkarKurdish:
@iamkun could you review this?

Issue #2027, opened 2022-08-17
Order of plugin extends breaks plugin functions


Worly:
**Describe the bug**
I'm trying to use two plugins (duration and objectSupport) in my Angular app. While using datejs#add function with argument of type Duration (implemented in duration plugin), I get Invalid Date if the plugin objectSupport is extended **after** the duration plugin. If objectSupport is extended **before** the duration plugin then add method works fine.

**Demo**
CodeSandbox link https://codesandbox.io/s/festive-sutherland-thxuux?file=/src/app/app.module.ts
In app.module.ts, change the order of lines 10 and 11, and look at console after refreshing Browser preview tab.

**Expected behavior**
No matter the order of plugin extends, all plugin functions should work as expected.

**Information**
 - Day.js Version ^1.11.5
 - OS: Windows 10
 - Browser Chrome/103.0.5060.134

BePo65:
As a basic pattern, plugins create a chain of function calls for overwritten / extended functions (like `parse` or `add`).  
Therefore the sequence of extending dayjs **is** important (and should be documented in the documentation site).

Some plugins depend on other plugins. In that case this is part of the documentation and means that the dependant plugin must be `extend`-ed after the prerequisite plugin.

`duration` and `objectSupport` are a special case: the problem is that a ` Duration`  object is `instanceof Duration` and of course also `instanceof Object` and that information is used e.g. in parsing. So the duration plugin must be extend-ded after the objectSupport plugin and that information is missing in the documentation.

Perhaps you want to create a PR in the [docu repo](https://github.com/dayjs/dayjs-website)?

iamkun:
Maybe we can update the logic 
```
argument instanceof Object // PLUS, also not a Duration object
```
to fix this issue?

-----

btw, @BePo65 , do you mind taking a look at a minor update in the comments in #1914 so that we can get it merged?

BePo65:
I am afraid that this will not work, as the class `Duration` is not known in the `objectSupport` plugin (or anywhere outside of the duration plugin).

We could add something like this to the start of the initialization of the objectSupport plugin:
```
  if (dayjs.duration !== 'undefined') {
    console.error('the plugin objectSupport should be extended before the plugin Duration')
  }
```
but IMHO two plugins should not directly reference each other. And wouldn't we have also do this for plugins that depend on other plugins?

Although nobody reads the documentation 😄 I would simply update the documentation (where this consequence is not documented until today).

[...]

PR #2038, opened 2022-08-24
Fix objectSupport collides with Duration plugin - issue #2027

BePo65:
Using the plugins objectSupport and Duration in one file creates false result in .add and .subtract when `extend(Duration)` comes before `extend(objectSupport)` - see issue #2027.

Fixed this problem and added tests to avoid regression.


iamkun:
THX

Issue #1872, opened 2022-04-26
startOf with timezone plugin gives wrong output for toISOString


SachaAvazashvili:
**Current behavior**

```
console.log(dayjs.tz('2022-03-01', 'Europe/Paris').startOf('day').format()); // correct 2022-03-01T00:00:00+01:00
console.log(dayjs.tz('2022-03-01', 'Europe/Paris').format()); // correct 2022-03-01T00:00:00+01:00

console.log(dayjs.tz('2022-03-01', 'Europe/Paris').startOf('day').toISOString()); // incorrect 2022-03-01T00:00:00.000Z
console.log(dayjs.tz('2022-03-01', 'Europe/Paris').toISOString()); // correct 2022-02-28T23:00:00.000Z
```

**Expected behavior**

```
console.log(dayjs.tz('2022-03-01', 'Europe/Paris').startOf('day').format()); // 2022-03-01T00:00:00+01:00
console.log(dayjs.tz('2022-03-01', 'Europe/Paris').format()); // 2022-03-01T00:00:00+01:00

console.log(dayjs.tz('2022-03-01', 'Europe/Paris').startOf('day').toISOString()); // 2022-02-28T23:00:00.000Z
console.log(dayjs.tz('2022-03-01', 'Europe/Paris').toISOString()); // 2022-02-28T23:00:00.000Z
```

I can't figure why but there is a one hour offset when I use `toISOString` after `startOf('day')`.

**Information**
 - Day.js Version 1.11.1
 - Time zone: Europe/Paris

PR #2061, opened 2022-09-12
fix: 🐛 startOf keepLocalTime issue on UTC

yutin1987:
for startOf issue only.

✅ Closes: #1803
✅ Closes: #1872


Spirit04eK:
Why did you close?

yutin1987:
I don’t think the owner wants to review it, but you can use it.

Issue #1891, opened 2022-05-09
Dayjs: r.ordinal is not a function at advanceFormat.js


Kashish-Chaudhary:
**Describe the bug**
When I try to format the date to show ordinal for date it breaks the app and shows error r.ordinal is not a function.
Code:
import dayjs from 'dayjs';
import advancedFormat from 'dayjs/plugin/advanceFormat';
dayjs.extend(advancedFormat);
const formattedDate = dayjs(endDate).format("MMMM Do, YYYY");

**Expected behavior**
It should format date in this format => December 31st, 2022

**Information**
 - Day.js Version [1.10.7]
 - OS: [iOS]
 - Browser [chrome [100.0.4896.127]]

Kashish-Chaudhary:
@iamkun Can please clarify this issue?

tziyang-lum:
`'Do'` format seems broken. Getting the same error as above

Kashish-Chaudhary:
That's correct as when we remove `'Do'` in format error goes away.

BePo65:
I didn't try the code in Chrome, but only on my node.js system.
At least in your code above there is a typo in the second line:
>import advancedFormat from 'dayjs/plugin/advanceFormat';

should be `import advancedFormat from 'dayjs/plugin/advancedFormat';` (you see the 'd' before the 'Format'?)

Perhaps this was the culprit - otherwise I will have to test it on chrome ;-)

Kashish-Chaudhary:
@BePo65 In actual implementation I used `import advancedFormat from 'dayjs/plugin/advancedFormat'` only but while posting I typed it wrong. I've updated it though.

BePo65:
Tested the following code in chrome (vxx) and Firefox (vxx) and got this result:
![issue1891](https://user-images.githubusercontent.com/6582465/173173363-1b06ebef-1526-4f5c-93ff-0a1296b4cef2.jpg)


```
<!DOCTYPE html>
<html lang="de">
<head>
<meta charset="utf-8">
<title>dayjs issue 1891</title>
</head>
<body>
  <h1 id="header">dayjs issue 1891</h1>
  <p id="result">waiting...</p>
  <script src="https://unpkg.com/dayjs@1.11.3/dayjs.min.js"></script>
  <script src="https://unpkg.com/dayjs@1.11.3/plugin/advancedFormat.js"></script>
  <script>
    dayjs.extend(window.dayjs_plugin_advancedFormat)

[...]

PR #2071, opened 2022-09-26
refactor: move declaration for english 'ordinal' to 'en' locale file

BePo65:
All locale files have a definition for 'ordinal' except the 'en' one. For this locale the definition of 'ordinal' is part of the 'AdvancedFormat' plugin. This is inconsistent not quite logical.

This pr solves this topic and should solve issue #1891

### DIFF
--- a/src/locale/en.js
+++ b/src/locale/en.js
@@ -3,5 +3,10 @@
 export default {
   name: 'en',
   weekdays: 'Sunday_Monday_Tuesday_Wednesday_Thursday_Friday_Saturday'.split('_'),
-  months: 'January_February_March_April_May_June_July_August_September_October_November_December'.split('_')
+  months: 'January_February_March_April_May_June_July_August_September_October_November_December'.split('_'),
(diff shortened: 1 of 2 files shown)


iamkun:
Thanks.

But there's another problem, format('Do') function is also added in `advancedFormat` plugin, that is to say, even if added ordinal to 'en', we still can't get correct format('Do') result without advancedFormat plugin 🤔

BePo65:
Of course that is correct. But IMO code / architecture should be consistent. 
By now we have the situation that for 'en' we have the definition of  the ordinal function in the plugin AdvancedFormat and for all other locales we have it in the corresponding locale file.

If someone want to do format('Do'), he needs AdvancedFormat, so this does not make a functional difference, but architecture is more than pure functionality (I am sure you agree, if you look at my programs from 30 years ago 😄).

iamkun:
well, that sounds reasonable, let's get it merged.

PR #2089, opened 2022-10-21
update packages to latest versions

BePo65:
Update Packages to current versions.

New version of eslint has new default rules. I did not change the eslint configuration to compensate for this, as I have the impression that dayjs stays as close as possible to the recommended settings.  
I also activated the eslint plugins for jest and import (seemed to be missing).  
Fixing the resulting warnings / errors created a lot of change files,

When updating "plugin-rollup-babel" (now called "@plugin/rollup-babel) the size returned by size-limit increases from 2.65kB to 2.99kB (I didn't have the time to trace this down).


BePo65:
Just as a little help for the eslint induces changes, here a list of the new / changed rules:

**from activated plugins** (import; jest)
- import/order
- jest/expect-expect
- jest/no-alias-methods
- jest/no-conditional-expect
- jest/no-identical-title
- jest/no-standalone-expect
- jest/valid-title


**from eslint rules that are new / changed in latest version**
- arrow-parens
- default-param-last
- function-call-argument-newline
- function-paren-newline
- implicit-arrow-linebreak
- lines-between-class-members
- no-constructor-return
- no-else-return
- no-multiple-empty-lines
- object-curly-newline
- operator-linebreak

iamkun:
Thanks. 

The thing I care about most is the bundle size. As a tiny lib, 2.65kB to 2.99kB size increase seems a huge number and should be considered in the first place.

BePo65:
The question is: why does '@plugin/rollup-babel' create a larger bundle and can we reduce it again by tweaking the configuration.

As I don't have any experience with bundlers: do you have any idea?  
Should we open an issue there?

iamkun:
it will always be a good idea to keep the bundle size as small as we can. it's the core feature of our lib.

BePo65:
I will look, if I find a solution. If this will take me too long, perhaps we should only update the rest until we find  good solution / better configuration.

BePo65:
So, what did I find out:
+ size-limit v0.18.0 with the configuration from dayjs sends the defined input ('dayjs.min.js') to webpack and gets the size of the resulting packed file. This way size-limit mimics the build process of an application using webpack.  
But 'dayjs.min.js' is a file already packed by rollup and terser.  
If we set `"webpack": false` in the configuration of size-limit we get a size of 2.88kB.
+ the size of dayjs is especially important, when dayjs is used in a web page using a script tag. For this use case, dayjs creates 'dayjs,min.js'. The size of this file after using gzip (like a webserver would) is 2932 bytes.
+ after upgrading to rollup v3 the content of 'dayjs.min.js' and these values remain unchanged.
+ the latest version of size-limit switched from webpack to esbuild. But as mentioned before, we don't need to bundle the already bundled 'dayjs.min.js'. The output of size-limit is now 2.95kB (without any change to the size of the gzipped file 'dayjs.min.js'.
+ gzip-size always shows 2.95kB.

**The size and content of 'dayjs.min.js' did not change after updating the packages** (I used 'winmerge' to verify this).'

For me this means:
+ configure size-limit to not use a packer / minifier
+ drop gzip-size (as the result is what we should see from size-limit too).
+ use size-limit only to verify during the ci builds that the generated bundle did not get too large.
+ and perhaps we should set the limit for size-limit a few bytes higher (to consider the correct size of the gzipped file).

PR #2128, opened 2022-11-16
dayjs 2.0 - plugin utc - based on pr 2050

BePo65:
This pull request adds the plugin 'utc' to dayjs 2.0.

The main problems creating this plugin were:
+ sometimes there is no way to avoid the access to the internal fields of the class `Dayjs`
+ I did not find a way to add a setter to a getter already defined in Dayjs (overloading an existing method in a class from another module). In the end either typescript did not evaluate the syntax or did not generate javascript with the desired functionality.

As a result I had to add hooks to the Dayjs class to enable extensions via plugins.

**Adding new fields and methods via type definitions in the plugin module:**
In the first plugins for dayjs 2.0, all methods and fields that are added via type definitions to the 'core' dayjs module, are added to the 'base class' `Extend`.  
This will create the right functionality, but the architectural reason for doing it this way is not clear to me.

So in this plugin here, I added methods and fields directly to the Dayjs class.


leemoria:
hello，When will the latest version be released， Now is 1.11.6

PR #2171, opened 2022-12-20
Fixed typo in be.js locale

bielarusajed:


belarusaed:
"АТ" is russian-way to short "Аўторак" (Tuesday)
"АЎ" is how Microsoft short "Аўторак" (Tuesday) in their apps and in Windows OS:
image
image
Also, it is common to use "АЎ" instead of "АТ" in paper calendars in Belarusian language.

Issue #2189, opened 2023-01-04
export = locale in index.d.ts doesnt allow synthetic import


Alex33400:
**Describe the bug**
When I want to import dayjs as synthetic import in my lib, I get the following error : 

export = dayjs;
      ~~~~~~~~~~~~~~~
    This module is declared with using 'export =', and can only be used with a default import when using the 'allowSyntheticDefaultImports' flag

**Expected behavior**
Application can build with this line : "import dayjs from 'dayjs';"

**Information**
 - Day.js Version [e.g. v1.11.7]
 - OS: [Windows]
 - Browser: Chrome 108
 - Time zone: GMT +1

BePo65:
As a starter: this does not only affect the 'core' program of dayjs (as handled in your pr #2190), but any plugin and locale file.

From your pr I suppose you are using dayjs in an esm environment. But dayjs is a commonjs library (the esm first implementation is still under construction; mostly called "dayjs 2.0", but according to the authors it will probably get 3.0).

I made a pr for fixing all the (default) type definitions for esm (pr #1581) and the answer was "let's implement it in dayjs 2.0" (at least that is my interpretation 😄).

The typescript [handbook ](https://www.typescriptlang.org/docs/handbook/modules.html#export--and-import--require)says that `export =` in the file ZipCodeValidator requires an `import zip = require("./ZipCodeValidator");` in the module using it.
So perhaps that is an acceptable workaround until we get "dayjs 2.0".

trim21:
> From your pr I suppose you are using dayjs in an esm environment. But dayjs is a commonjs library

But it also break esm build, you can't `import dayjs from 'dayjs/esm'` without `allowSyntheticDefaultImports` flag.

BePo65:
How about using `import zip = require("dayjs")` as the handbook says instead of `import dayjs from 'dayjs/esm`?

Or `import * as dayjs from 'dayjs'` as the dayjs [documentation](https://day.js.org/docs/en/installation/typescript) recomments?

In my (esm) application, I gave up to try to use 'dayjs/esm' because of the effects you see in your app too.

trim21:
TBH, `import * as dayjs from 'dayjs'` is not semantic correct.

This means it import all named exports and put them as properties of `dayjs`. It means `dayjs` here is a object, it's not callable.

and `import dayjs = require("dayjs");` actually only works in `.d.ts`, but not `.ts`, you can't compile it with tsc.

```text
[...]

PR #2190, opened 2023-01-04
fix: allow synthetic import of dayjs

Alex33400:
fix issue #2189, tested with Angular


Alex33400:
as discussed in issue #2189, we should wait for a new dayjs architecture probably that will probably release with dayjs 2.0

PR #2222, opened 2023-02-01
fix: Mark `date` parameter as optional in timezone plugin

andrew-atwood-infinitusai:
LIke `dayjs()`, `dayjs.tz()` falls back to the default time internally if no parameter is passed. But the current Typescript types do not allow for this existing, expected, and correct functionality.


iamkun:
Thanks

PR #2226, opened 2023-02-04
Added valueOf method to Duration class

pvmeijel:
The Duration class in MomentJS contains the valueOf method (https://github.com/moment/moment/blob/develop/src/lib/duration/as.js). This method returns the duration in milliseconds. This PR adds this method also to the Dayjs Duration class.


iamkun:
Thanks

Issue #2207, opened 2023-01-15
Timezone plugin returns incorrect values or null


dawidzawadzki:
**Describe the bug**

The Timezone plugin returns incorrect values or null on iOS when Hermes engine is enabled. 

**Information**
 - Day.js: 1.11.7 
 - OS: iOS
 - React Native: 0.71
 - Time zone: GMT+01:00 

**Steps to reproduce**
1. `npx react-native init AwesomeProject`
2. `yarn add dayjs`

**Code example**
```
import React from 'react';
import {SafeAreaView} from 'react-native';
import dayjs from 'dayjs';
import timezoneDayjsPlugin from 'dayjs/plugin/timezone';
import utcDayjsPlugin from 'dayjs/plugin/utc';

dayjs.extend(utcDayjsPlugin);
dayjs.extend(timezoneDayjsPlugin);

const App = () => {
  const timestamp = '2014-06-01 12:00';
  const tz = 'America/New_York';
  const dayjsLocal = dayjs(timestamp);

  console.log(dayjsLocal.tz(tz).format('YYYY-MM-DDTHH:mm:ss'));
  return <SafeAreaView />;
};

export default App;
```
**Expected behavior**

Example code should return `"2014-06-01T06:00:00"`

nemphys:
This is driving me nuts!

I have been using dayjs and the timezone plugin in a node project for more than 2 years now and, just recently (and without any change related to dayjs whatsoever) it started producing invalid dates whenever the timezone plugin is used!

What is even weirder is that it only happens on my Mac mini, whereas everything looks normal on several ubuntu servers where the app is also running. All machines have the same node version (16.19.0).

Any ideas?

nemphys:
Nevermind, apparently this happened due to the homebrew icu4c update to 72.1...
The good news is that node 18.3.0 seems to have mitigated the issue and works properly with the new icu version.

Pareder:
I faced the same issue, on iOS everything is good, but on Android I got `null` and `Invalid Date`. Maybe it is somehow related to https://github.com/facebook/react-native/issues/28097. Have you found a solution, @dawidzawadzki ?

dawidzawadzki:
@Pareder  Unfortunately not. I've disabled Hermes on iOS for now.

durgaswaroop:
This is happening for me too. But it seems to be random. In one run it gives proper values and in another they are all `null` values. I was debugging the issue for so long!!

KingAmo:
[...]

PR #2227, opened 2023-02-05
fix: offset calculation of timezone plugin

Pareder:
Update calculation of timezone offset. This change was caused by this issue https://github.com/iamkun/dayjs/issues/2207. The thing is that [Hermes](https://reactnative.dev/docs/hermes) in React Native doesn't fully support Intl API (https://github.com/facebook/hermes/issues/23). Let me explain.

This code
```
const target = date.toLocaleString('en-US', { timeZone: timezone })
```
works correct, it returns the date in `en-US` format, but after that when returned string is passed to `Date` constructor it returns `Date { NaN }`
```
const target = date.toLocaleString('en-US', { timeZone: timezone })
new Date(target) // Date { NaN }
```

Date constructor cannot recognise this format, so calculation of difference is broken. And I think that we can use `dayjs` to do backward formatting.

**Note**: in my opinion we can improve somehow the calculation of timezone offset not to depend on some APIs support (get rid of `Date.toLocaleString`).

### DIFF
--- a/src/plugin/timezone/index.js
+++ b/src/plugin/timezone/index.js
@@ -96,7 +96,8 @@ export default (o, c, d) => {
     const oldOffset = this.utcOffset()
     const date = this.toDate()
     const target = date.toLocaleString('en-US', { timeZone: timezone })
-    const diff = Math.round((date - new Date(target)) / 1000 / 60)
+    const targetDate = d(target, 'M/D/YYYY H:mm:ss A')
+    const diff = Math.round(this.diff(targetDate, 'ms') / 1000 / 60)
     let ins = d(target).$set(MS, this.$ms)
       .utcOffset((-Math.round(date.getTimezoneOffset() / 15) * 15) - diff, true)
     if (keepLocalTime) {


iamkun:
not sure if this is a correct fix, is there a way to add some test cases to prove it right?

Pareder:
@iamkun None of timezone tests is broken, what test do you want me to add? Do you want to simulate `Date { NaN }` behavior or what?

src/plugin/timezone/index.js
```diff
@@ -96,7 +96,8 @@ export default (o, c, d) => {
     const oldOffset = this.utcOffset()
     const date = this.toDate()
     const target = date.toLocaleString('en-US', { timeZone: timezone })
-    const diff = Math.round((date - new Date(target)) / 1000 / 60)
+    const targetDate = d(target, 'M/D/YYYY H:mm:ss A')
```
iamkun, line 99: can we use d(target) directlly?

src/plugin/timezone/index.js
```diff
@@ -96,7 +96,8 @@ export default (o, c, d) => {
...
+    const diff = Math.round(this.diff(targetDate, 'ms') / 1000 / 60)
```
iamkun, line 100: we don't need 'ms', it's the default one

src/plugin/timezone/index.js
Pareder: Actually we can't use it, because `parseDate` method returns `new Date` (https://github.com/iamkun/dayjs/blob/dev/src/index.js#L79), which is the reason of this PR and in case of using `customParseFormat` plugin without passing `format` argument it calls original `parseDate` method under the hood https://github.com/iamkun/dayjs/blob/dev/src/plugin/customParseFormat/index.js#L257. So my fix will work only when using `customParseFormat` plugin, what do you think about it?

iamkun: True, that's my concern. If we implement it in this way, the timezone plugin must rely on customParseFormat plugin to work that our users may get confused

Pareder: I got your point. Maybe I can find a better solution to calculate timezone offset not to depend on some APIs support.

Pareder: @iamkun Could you please check the new solution? I am using `formatToParts` method of `Intl.DateTimeFormat` instance and getting year, month, day, etc. from its result. What do you think? I tried this in my React Native environment and it works perfectly. All tests are passed correctly as well.

Pareder:
@iamkun Could you please review it?

KingAmo:
any update on this ?

artemkhalygov:
@iamkun Please merge this pull request

jgtorrejon:
Tried applying a patch but still does not work on iOS 😔

milesingrams:
I didnt see this PR and wrote a similar one that solves the same problem using dayjs to parse the date string. Here is it. Please approve either one of these @iamkun. Pleasssssse! Would hugely appreciate it. We are having to use the patch-package trick to overwrite these lines in our codebase until its fixed.

https://github.com/iamkun/dayjs/pull/2843

PR #2233, opened 2023-02-07
Fix: objectSupport parsing objects with invalid units

gastronleroux:
Currently, if `dayjs` is extended with `objectSupport`, providing any kind of object that is not undefined, a Date or an Array ends up with a valid date. This behavior is unexpected and prone for false positives.
Before:
```js
dayjs({ value: { year: 2023 } }).isValid() // => true (current date)
```
After:
```js
dayjs({ value: { year: 2023 } }).isValid() // => false
```

### DIFF
--- a/src/constant.js
+++ b/src/constant.js
@@ -20,7 +20,6 @@ export const M = 'month'
 export const Q = 'quarter'
 export const Y = 'year'
 export const DATE = 'date'
-export const UNITS = [MS, S, MIN, H, D, W, M, Q, Y, DATE]
 
 export const FORMAT_DEFAULT = 'YYYY-MM-DDTHH:mm:ssZ'
 
--- a/src/constant.js
+++ b/src/constant.js
@@ -20,6 +20,7 @@ export const M = 'month'
 export const Q = 'quarter'
 export const Y = 'year'
 export const DATE = 'date'
+export const UNITS = [MS, S, MIN, H, D, W, M, Q, Y, DATE]
 
 export const FORMAT_DEFAULT = 'YYYY-MM-DDTHH:mm:ssZ'
 
--- a/src/plugin/objectSupport/index.js
+++ b/src/plugin/objectSupport/index.js
@@ -1,4 +1,6 @@
-import { UNITS } from '../../constant'
+import { D, Y, M, H, MIN, S, MS, DATE } from '../../constant'
+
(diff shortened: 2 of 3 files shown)


iamkun:
seems there is no need to check the keys since it is passed by our user.

gastronleroux:
> seems there is no need to check the keys since it is passed by our user.

True, in the context of end-user interaction, I don't see a problem here. However, looking at the bigger picture - should a tool for building dates create one as valid if the input is invalid (and without any warnings)? It's more a matter of helping the developer here.
I have already come across such a situation where it turned out to be problematic when migrating an app from vue 2 to vue 3 with composition api. A computed value was not passed correctly, `.value` was omitted. Because at first glance the entire view looked correct, this oversight was only reported by end users saying the date did not match the expected one. The situation could be avoided if the date would appear as null-like in the first place (or if cr would have been stricter, but no one's perfect 😄)

iamkun:
THX, the current logic will just ignore the error key passed in, which has no effect on the result. 🤔

gastronleroux:
> THX, the current logic will just ignore the error key passed in, which has no effect on the result. 🤔

The current logic will treat objects with invalid units as empty objects, so the effect it has is it returns the current date instead of a null-date. I would expect `dayjs('invalid date')` to return the same thing as `dayjs({ invalid: 'date' })` for reasons mentioned in the comment above + consistency (even if practically the latter can only happen if the dev is at fault 😄)

src/constant.js
```diff
@@ -20,6 +20,7 @@ export const M = 'month'
 export const Q = 'quarter'
 export const Y = 'year'
 export const DATE = 'date'
+export const UNITS = [MS, S, MIN, H, D, W, M, Q, Y, DATE]
```
iamkun, line 23: all right, that makes sense.

The last thing we should take care of is how to make this constant file as small as possible. Since it will bundle to the main bundle file.

Maybe we can try to use the units above? (not sure)

gastronleroux: Good point. We're fetching object values for each of them explicitly, it makes sense to import each of them separately.

iamkun:
Everything looks cool to me.

But, I just take a look at moment.js, they will treat these to the current date either.

The same question comes again, whether should we handle it or not

image

iamkun (approved):

PR #2244, opened 2023-02-21
dayjs.diff improve performance

Qquanwei:
在高频场景下 dayjs.diff 会成为性能瓶颈, 发现在进行 `dayjs.diff(time, 'day')` 计算时会有无效的一些计算，于是优化了在进行 diff day 场景下的性能

image

test case

```js
const dayjs =  require('./dayjs.min.js');

let startTime = Date.now();
let endTime = Date.now() + 30 * 24 * 60 * 60 * 1000;

for(let i = 0; i < 100000; ++i) {
    dayjs(startTime).diff(endTime, 'day');
}
```

before

image

after

image

执行时间由 1.35s/100000次 提升到 0.2s/100000次

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -319,18 +319,20 @@ class Dayjs {
     const that = dayjs(input)
     const zoneDelta = (that.utcOffset() - this.utcOffset()) * C.MILLISECONDS_A_MINUTE
     const diff = this - that
-    let result = Utils.m(this, that)
+    const getMonth = () => Utils.m(this, that)
 
-    result = {
-      [C.Y]: result / 12,
-      [C.M]: result,
-      [C.Q]: result / 3,
-      [C.W]: (diff - zoneDelta) / C.MILLISECONDS_A_WEEK,
-      [C.D]: (diff - zoneDelta) / C.MILLISECONDS_A_DAY,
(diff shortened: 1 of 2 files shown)


src/index.js
```diff
@@ -319,18 +319,20 @@ class Dayjs {
...
+      [C.M]: () => getMonth(),
+      [C.Q]: () => getMonth() / 3,
+      [C.W]: () => (diff - zoneDelta) / C.MILLISECONDS_A_WEEK,
+      [C.D]: () => (diff - zoneDelta) / C.MILLISECONDS_A_DAY,
+      [C.H]: () => diff / C.MILLISECONDS_A_HOUR,
+      [C.MIN]: () => diff / C.MILLISECONDS_A_MINUTE,
+      [C.S]: () => diff / C.MILLISECONDS_A_SECOND
+    }[unit] || (() => diff) // milliseconds
+
+    result = result()
```
iamkun, line 335: image
a lot of functions code will be added, which will increase the bundle size.
if we change this to `switch...case`, it will both improve performance, as well as minify the bundle code.

Qquanwei: done

Qquanwei:
可以合了么

PR #2264, opened 2023-03-22
fix: clone dates with timezone no offset

### LINKED ISSUE #2254: UTC flag is set during timezone initialization if passed TZ has offset 0 (e.g. "Europe/London")
**Describe the bug**
When calling `.tz` with a timezone that is at utc offset 0 (such as `Europe/London` or `Europe/Dublin`), the UTC flag is set on the dayjs object. This causes date arithmetic and formatting to use UTC, rather than the particular timezone that was set, and can often cause the system timezone to leak into calculations.

An example:
```
const date = dayjs("2023-01-01T00:00:00Z").tz("Europe/London"); // January 1st 2023
console.log(date.isUTC()) // true
```
The flag is not set when the timezone is in daylight savings, such that the offset is no longer 0:
```
const dateWithDST = dayjs("2023-04-01T00:00:00Z").tz("Europe/London"); // April 1st 2023
console.log(dateWithDST.isUTC()) // false
```

The culprit is a line in `utcOffset` that sets the flag if the passed offset is `0` and `keepLocalTime` is true. There's a call in the `.tz` method that meets this criteria, so the flag is set during timezone initialization. I propose that we should explicitly set the $u flag to false during the `.tz` call, unless the timezone passed in is `UTC`.

**Expected behavior**
The UTC flag shouldn't be set during timezone initialization.

**Information**
 - Day.js Version: Latest
 - OS: MacOS
 - Browser: Any
 - Time zone: `America/New_York`, `Europe/London` for testing

### LINKED ISSUE #2265: Cloning a date with a timezone without offset results in a different date
**Describe the bug**
First of all, thanks for the great library! We’ve found a bug when cloning a date with a timezone without offset. This can easily be replicated with the following test:

```
const instance = dayjs('2022-01-01T03:00:00.000Z').tz('Europe/London')
const another = instance.clone()
expect(another.format()).toEqual(instance.format())
```

After cloning, the date is not the same as the original one.

**Expected behavior**
Cloning should not change the date when a timezine without offset is used.

**Information**
 - Day.js Version v1.11.7
 - OS: Linux
 - Browser NodeJS
 - Time zone: WET +00:00

gerbenmeyer:
This PR closes #2265: Cloning a date with a timezone without offset results in a different date.

The idea behind this PR is to only treat timezone UTC as a UTC date, and not all other timezones without offset (as is currently the case). Hence, this PR would also resolve #2254.


allenm:
👍

I need this

adam-drozdz-ie (approved):

adam-drozdz-ie:
Why don't we merge it?

gerbenmeyer:
Good question, I also really need this, and it's solving multiple issues.

phwes:
Will this be merged anytime soon? As @gerbenmeyer says, it solves a lot of issues.

PR #2272, opened 2023-03-31
[d.ts] first parameter date is optional in isSame(), isBefore(), isAfter()

FourwingsY:
I don't know why this PR https://github.com/iamkun/dayjs/pull/1929 closed,
But it is quite useful to use `birthday.isBefore()` with no params in typescript.


iamkun:
thanks

PR #2291, opened 2023-04-19
shortened up the wrong translated month with 3 characters

duesentrieb26:


src/locale/de.js
```diff
@@ -29,7 +29,7 @@ const locale = {
   weekdaysShort: 'So._Mo._Di._Mi._Do._Fr._Sa.'.split('_'),
   weekdaysMin: 'So_Mo_Di_Mi_Do_Fr_Sa'.split('_'),
   months: 'Januar_Februar_März_April_Mai_Juni_Juli_August_September_Oktober_November_Dezember'.split('_'),
-  monthsShort: 'Jan._Feb._März_Apr._Mai_Juni_Juli_Aug._Sept._Okt._Nov._Dez.'.split('_'),
+  monthsShort: 'Jan._Feb._Mär._Apr._Mai_Jun._Jul._Aug._Sep._Okt._Nov._Dez.'.split('_'),
```
iamkun, line 32: should we keep the same with https://github.com/moment/moment/blob/develop/src/locale/de.js#L28

duesentrieb26: I made a deeper research to find out all abbreviations for the month names. I've found this official paper from the german standard http://guest.engelschall.com/~sb/download/Date-Calc/DIN1355/ and ISO 8601
there it is written down.
I changed also the name for March to "Mrz."
So i think that this change also concern moment js

iamkun:
Thanks.

It will be better if some native speaker cloud review this PR

PR #2307, opened 2023-05-16
Update weekdaysMin of vi's locale

anhtuan219:
Hi, I'm working on a personal project and I want to make a date time picker input in Vietnamese. Currently, I'm using the date time picker of Material UI with date library dayjs (which is one of the best date libraries in my opinion :)). Everything it's great, but I noticed that label for weekdays (from Monday to Saturday) is just a letter 'T' (like the image below) and I can't distinguish weekdays (even though I'm Vietnamese). It seems that Material UI uses just the first letter of `weekdaysMin` for the label. I fixed that problem by customizing the locale. I think this could make a difference between `weekdaysShort` and `weekdaysMin` so I made this pull request. I don't know if my solution suits the library so please let me know if this change is not acceptable, thank you
before


anhtuan219:
OK, thanks for the reply. I can overwrite the locale in my project to show what I need so I will close this PR

PR #2330, opened 2023-06-09
Fix/minMax

dchueri:
### Motivation:
If I automatically get a list of dates from somewhere, that list might contain 'nulls', or even just return a single valid argument. At the moment, if not treated, the user will receive a TypeError or 'null' as a return, even with valid arguments.

### Makes two adjustments:
- If an array has a 'null' arg, doesnt return a TypeError, now just ignore the 'null' args and consider the others:

Example:
```
Before this alteration:
dayjs.max(dayjs("1995-03-16"), null, dayjs("2023-03-16")).format()
---> TypeError: Cannot read properties of null (reading 'isValid')

After:
dayjs.max(dayjs("1995-03-16"), null, dayjs("2023-03-16")).format()
---> 2023-03-16T00:00:00-03:00
```

- The second adjustment is when provided just one arg, instead of returning null, it returns the only given value.

Example:
```
Before this alteration:
dayjs.max(dayjs("1995-03-16"))
---> null

After:
dayjs.max(dayjs("1995-03-16"), null, dayjs("2023-03-16")).format()
---> 1995-03-16T00:00:00-03:00
```

I also added some tests to validate this propose.

### DIFF
--- a/src/plugin/minMax/index.js
+++ b/src/plugin/minMax/index.js
@@ -1,12 +1,20 @@
 export default (o, c, d) => {
   const sortBy = (method, dates) => {
-    if (!dates || !dates.length || !dates[0] || (dates.length === 1 && !dates[0].length)) {
+    if (
+      !dates ||
+      !dates.length ||
+      (dates.length === 1 && !dates[0]) ||
+      (dates.length === 1 && Array.isArray(dates[0]) && !dates[0].length)
+    ) {
       return null
     }
     if (dates.length === 1 && dates[0].length > 0) {
       [dates] = dates
     }
-    let result
+    while (dates.indexOf(null) !== -1) {
+      dates.forEach((date, index) => (!date ? dates.splice(index, 1) : null))
+    }
+    let result;
     [result] = dates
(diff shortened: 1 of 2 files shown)


dchueri:
https://github.com/iamkun/dayjs/issues/2306

src/plugin/minMax/index.js
```diff
@@ -1,12 +1,20 @@
...
+      (dates.length === 1 && Array.isArray(dates[0]) && !dates[0].length)
+    ) {
       return null
     }
     if (dates.length === 1 && dates[0].length > 0) {
       [dates] = dates
     }
-    let result
+    while (dates.indexOf(null) !== -1) {
+      dates.forEach((date, index) => (!date ? dates.splice(index, 1) : null))
```
iamkun, line 15: cloud we use `Array.prototype.filter` here?

dchueri: really, I will

iamkun (approved):

iamkun:
Cheers

Issue #2336, opened 2023-06-15
Duration add/subtract doesn't take into account how many days in a month


alexander-ulaev:
**Describe the bug**
methods `add` and `subtract` with duration object doesn't take into account how many days in a month

```javascript
// actual

dayjs('2023-02-01').add(dayjs.duration(1, 'month')) === "2023-01-30"
dayjs('2023-03-01').subtract(dayjs.duration(1, 'month')) === "2023-01-30"
``` 

**Expected behavior**
```javascript
// expected

dayjs('2023-02-01').add(dayjs.duration(1, 'month')) === "2023-02-01"
dayjs('2023-03-01').subtract(dayjs.duration(1, 'month')) === "2023-02-01"
``` 

**Information**
 - Day.js v1.11.8
 - OS: Ubuntu 23.04
 - Browser: chrome 114.0.5735.133
 - Time zone: GMT +04:00 Europe/Samara

PR #2337, opened 2023-06-15
fix: Duration add/subtract

alexander-ulaev:
fix Duration add/subtract doesn't take into account how many days in a month #2336

### DIFF
--- a/src/plugin/duration/index.js
+++ b/src/plugin/duration/index.js
@@ -257,6 +257,24 @@ class Duration {
   asYears() { return this.as('years') }
 }
 
+const addDuration = (date, duration) =>
+  date.add(duration.years(), 'year')
+    .add(duration.months(), 'month')
+    .add(duration.days(), 'day')
+    .add(duration.hours(), 'hour')
+    .add(duration.minutes(), 'minute')
+    .add(duration.seconds(), 'second')
+    .add(duration.milliseconds(), 'millisecond')
+
+const subtractDuration = (date, duration) =>
+  date.subtract(duration.years(), 'year')
+    .subtract(duration.months(), 'month')
+    .subtract(duration.days(), 'day')
+    .subtract(duration.hours(), 'hour')
+    .subtract(duration.minutes(), 'minute')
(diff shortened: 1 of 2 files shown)


src/plugin/duration/index.js
```diff
@@ -257,6 +257,24 @@ class Duration {
...
 }
 
+const addDuration = (date, duration) =>
+  date.add(duration.years(), 'year')
+    .add(duration.months(), 'month')
+    .add(duration.days(), 'day')
+    .add(duration.hours(), 'hour')
+    .add(duration.minutes(), 'minute')
+    .add(duration.seconds(), 'second')
+    .add(duration.milliseconds(), 'millisecond')
```
iamkun, line 267: a little trick to minimize the code, pass the third argument to combine add and subtract together.
```
// manipulateDuration(this, value, 1)
// manipulateDuration(this, value, -1)

// e.g. .add(duration.months() * (-1), 'month')
```

also, one thing that confused me, is why adding `value.asMilliseconds()`  is not the correct result. 
It should be the real `ms` between the two dates.

alexander-ulaev: thanks for the tip to improve the code
value.asMilliseconds() executed without date context and In this case, month is equal to 30 days. I think it's correct. (moment js has same behavior)

src/plugin/duration/index.js
```diff
@@ -257,6 +257,15 @@ class Duration {
...
 }
 
+const manipulateDuration = (date, duration, k) =>
+  date.add(duration.years() * k, 'year')
+    .add(duration.months() * k, 'month')
+    .add(duration.days() * k, 'day')
+    .add(duration.hours() * k, 'hour')
+    .add(duration.minutes() * k, 'minute')
+    .add(duration.seconds() * k, 'second')
+    .add(duration.milliseconds() * k, 'millisecond')
```
iamkun, line 267: we can replace 'second' with 's' according to https://github.com/iamkun/dayjs/blob/dev/src/utils.js#L31

to save some bytes

iamkun:
Cheers

PR #2338, opened 2023-06-15
fix: duration.minutes() should guaranteed to be in [0, 60)

### LINKED ISSUE #2329: duration({ minutes: 359 }).minutes() is not in [0, 60)
**Describe the bug**

```
// actual
dayjs.duration({ minutes: 359 }).minutes() === 59
```

**Expected behavior**

```
// expected
dayjs.duration({ minutes: 359 }).minutes() === 0
```

Per https://day.js.org/docs/en/durations/minutes:
> As with the other getters for durations, `dayjs.duration().minutes()` gets the minutes (0 - 59).

**Information**
 - version: `/dayjs@1.11.8` with `integrity: sha512-LcgxzFoWMEPO7ggRv1Y2N31hUf2R0Vj7fuy/m+Bg1K8rr+KAs1AEy4y9jd5DXe8pbHgX+srkHNS7TH6Q6ZhYeQ==`
 - OS: Windows 11 Pro / OS build 22621.1702
 - Browser: Chrome Version 114.0.5735.110 (Official Build) (64-bit)
 - Time zone: GMT-07:00 DST (Pacific Daylight Time)

KronosDev-Pro:
fix #2329


test/plugin/duration.test.js
```diff
@@ -141,80 +141,180 @@ describe('Clone', () => {
...
   describe('Negative number', () => {
-    expect(dayjs.duration(-500).milliseconds()).toBe(-500)
-    expect(dayjs.duration(-1500).milliseconds()).toBe(-500)
-    expect(dayjs.duration(-15000).milliseconds()).toBe(0)
-    expect(dayjs.duration(-500).asMilliseconds()).toBe(-500)
-    expect(dayjs.duration(-1500).asMilliseconds()).toBe(-1500)
-    expect(dayjs.duration(-15000).asMilliseconds()).toBe(-15000)
-  })
-})
+    it('milliseconds', () => {
```
iamkun, line 164: Negative number

test/plugin/duration.test.js
```diff
@@ -141,80 +141,180 @@ describe('Clone', () => {
...
+      expect(dayjs.duration({ seconds: -15 }).milliseconds()).toBe(0)
+    })
 
-describe('Add', () => {
-  const a = dayjs.duration(1, 'days')
-  const b = dayjs.duration(2, 'days')
-  expect(a.add(b).days()).toBe(3)
-  expect(a.add(1, 'days').days()).toBe(2)
-  expect(a.add({ days: 5 }).days()).toBe(6)
+    it('asMilliseconds', () => {
```
iamkun, line 172: Add

test/plugin/duration.test.js
```diff
@@ -141,80 +141,180 @@ describe('Clone', () => {
...
+      expect(dayjs.duration({ minutes: -59 }).asMinutes()).toBe(-59)
+      expect(dayjs.duration({ minutes: -60 }).asMinutes()).toBe(-60)
+      expect(dayjs.duration({ minutes: -1, seconds: -1 }).asMinutes().toFixed(2)).toBe('-1.02')
+    })
+  })
 })
 
 describe('Hours', () => {
-  expect(dayjs.duration(10000000).hours()).toBe(2)
-  expect(dayjs.duration(10000000).asHours().toFixed(2)).toBe('2.78')
```
iamkun, line 217: can you please revert this deletion

test/plugin/duration.test.js
```diff
@@ -141,80 +141,180 @@ describe('Clone', () => {
...
+      expect(dayjs.duration({ seconds: -60 }).asSeconds()).toBe(-60)
+      expect(dayjs.duration({ minutes: -1, seconds: -1 }).asSeconds()).toBe(-61)
+    })
+  })
 })
 
 describe('Minutes', () => {
-  expect(dayjs.duration(100000).minutes()).toBe(1)
-  expect(dayjs.duration(61000).minutes()).toBe(1) // 1 minute 1 second
-  expect(dayjs.duration(100000).asMinutes().toFixed(2)).toBe('1.67')
```
iamkun, line 212: also this one

iamkun:
after careful thought, the main issue leading to this bug is not related to the display `get` getter, but the setter and the constructor.


dayjs.duration({ minutes: 359 }).toJSON() should return "PT5H59M" rather than ""PT359M""

that is to say, you can try to move your calc logic to https://github.com/iamkun/dayjs/blob/dev/src/plugin/duration/index.js#L78

that will make the getter and setter both correct.

ps: IMHO, you have made a nice refactor to the test cases. However, it will be better to keep all the existing test cases to make the lib stronger. Any new PRs should only add new cases or modify the error cases. 😀

iamkun:
after checking the PR again, still I think, we should update the input part (setter and the constructor) to make the inner data correct, not in the output part

faraango (approved):

PR #2355, opened 2023-06-27
fix: Add null to min and max return type

YuLingCheng:
Currently `max` and `min` can return explicit `null` when there are no dates to compare ([see implementation](https://github.com/iamkun/dayjs/blob/dev/src/plugin/minMax/index.js#L9)).

This PR adds `null` to the return type of those functions. (I ended up with a crash on my project because I relied on the types)

### DIFF
--- a/types/plugin/minMax.d.ts
+++ b/types/plugin/minMax.d.ts
@@ -4,8 +4,8 @@ declare const plugin: PluginFunc
 export = plugin
 
 declare module 'dayjs' {
-  export function max(dayjs: Dayjs[]): Dayjs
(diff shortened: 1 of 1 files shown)


iamkun:
Thanks

20manas:
This extra null type is slightly annoying to deal with in simple cases or whenever one is certain that at least one date is being passed to `min`/`max`.

Is there no way to guarantee that a date will be returned when `mix`/`max` is called?

Perhaps adding something like `date1.max(...restDates)` can work because then there is a guarantee that at least one date (`date1`) is always passed to `max`.

Issue #1756, opened 2022-01-05
locale en and localizedFormat issue


pm3cx:
**Describe the bug**
When you load en language and trying to get a longDateFormat('LT') from the localeDate() it can't read the 'LT' property because is undefined.  en-ca it's working fine.

**Code**
```javascript


const dayjs = require('dayjs'),
      localizedFormat = require('dayjs/plugin/localizedFormat'),
      localeData  = require('dayjs/plugin/localeData'),
      de = require('dayjs/locale/de'), // HH:mm
      fr = require('dayjs/locale/fr'), // HH:mm
      es = require('dayjs/locale/es'), // HH:mm
      it = require('dayjs/locale/it'), // HH:mm
      pt = require('dayjs/locale/pt'), // HH:mm
      br = require('dayjs/locale/br'), // h[e]mm A
      nl = require('dayjs/locale/nl'), // HH:mm
      ru = require('dayjs/locale/ru'), // HH:mm
      pl = require('dayjs/locale/pl'), // HH:mm
      zh = require('dayjs/locale/zh'), // HH:mm
      zhCn = require('dayjs/locale/zh-cn'), // HH:mm
      ja = require('dayjs/locale/ja'), // HH:mm
      cs = require('dayjs/locale/cs'), // H:mm
      tr = require('dayjs/locale/tr'), // HH:mm
      enGb = require('dayjs/locale/en-gb'), // HH:mm
      enCa = require('dayjs/locale/en-ca'), // h:mm A
      en = require('dayjs/locale/en'); // error

dayjs.locale(en);

dayjs.extend(localizedFormat);
dayjs.extend(localeData );


console.log(dayjs.localeData().longDateFormat('LT'));

```

**Information**
 - Day.js Version 1.10.7
 - OS: Windows
 - Browser Node.js
 - Time zone: [e.g. GMT-07:00 DST (Pacific Daylight Time)]

bitlab-code:
same issue with `dayjs.localeData().longDateFormat('L')`

lpmusicon:
I've made the PR that should fix the issue, English locale didn't seem to have formats provided.

iamkun:
Tested ok here

<img width="628" alt="image" src="https://github.com/iamkun/dayjs/assets/17680888/89991504-6c35-47e4-8b43-0dcf3101b546">
https://runkit.com/embed/2hb83tk2z3wx

pm3cx:
Great!

lpmusicon:
[...]

PR #2357, opened 2023-06-29
fix: Missing formats for en locale

lpmusicon:
Hi!

This patch should fix the issue #1756 for English (default) locale. I've tested it locally with my repo and it seems to do the trick.

Kind regards,
Greg


iamkun:
I've tested the case in https://github.com/iamkun/dayjs/issues/1756#issuecomment-1614157073, and it works as expected.

ad-boschung:
@iamkun 
Is this fix included in the latest version 1.11.10?
Because I'm using it and it seems that the problem is still there.

PR #2362, opened 2023-06-30
fix: duration plugin - MILLISECONDS_A_MONTH const calculation

sanjaynishad:
`MILLISECONDS_A_DAY * 30` will generate wrong value for a larger diffs, it's generating wrong durations for months

it should be either `MILLISECONDS_A_DAY * 30.416666666666668` (365 / 12) or `MILLISECONDS_A_YEAR / 12`

### DIFF
--- a/test/plugin/duration.test.js
+++ b/test/plugin/duration.test.js
@@ -85,8 +85,9 @@ describe('Parse ISO string', () => {
   it('ISO string with week', () => {
     const d = dayjs.duration('P2M3W4D')
     expect(d.toISOString()).toBe('P2M25D')
-    expect(d.asDays()).toBe(85) // moment 85, count 2M as 61 days
-    expect(d.asWeeks()).toBe(12.142857142857142) // moment 12.285714285714286
+    expect(d.asDays()).toBe(85.83333333333333) // moment 86, count 2M as 61 days
+    expect(d.asWeeks()).toBe(12.261904761904763) // moment 12.285714285714286
+    expect(d.asMonths()).toBe(2.8219178082191783) // moment 2.8213721020965523
   })
   it('Invalid ISO string', () => {
     expect(dayjs.duration('Invalid').toISOString()).toBe('P0D')
--- a/src/plugin/duration/index.js
+++ b/src/plugin/duration/index.js
@@ -8,7 +8,7 @@ import {
 } from '../../constant'
 
 const MILLISECONDS_A_YEAR = MILLISECONDS_A_DAY * 365
-const MILLISECONDS_A_MONTH = MILLISECONDS_A_DAY * 30
+const MILLISECONDS_A_MONTH = MILLISECONDS_A_YEAR / 12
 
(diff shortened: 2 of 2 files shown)


sanjaynishad:
actually one more fixes required with `parseFromMilliseconds`

need to check if duration ms is having leap year, then need to recalculate `MILLISECONDS_A_YEAR` and `MILLISECONDS_A_MONTH`

something like 

```
parseFromMilliseconds() {
    let { $ms } = this
    let msInYear = MILLISECONDS_A_YEAR;
    let msInMonth = MILLISECONDS_A_MONTH;
    let years = roundNumber($ms / MILLISECONDS_A_YEAR);
    if (roundNumber(years / 4)) {
      msInYear = MILLISECONDS_A_DAY * 365.25;
      msInMonth = msInYear / 12;
    }

    this.$d.years = roundNumber($ms / msInYear)
    $ms %= msInYear
    this.$d.months = roundNumber($ms / msInMonth)
    $ms %= msInMonth
    this.$d.days = roundNumber($ms / MILLISECONDS_A_DAY)
    $ms %= MILLISECONDS_A_DAY
```

iamkun:
can you fix the CI as well pls?

test/plugin/duration.test.js
```diff
@@ -85,8 +85,9 @@ describe('Parse ISO string', () => {
   it('ISO string with week', () => {
     const d = dayjs.duration('P2M3W4D')
     expect(d.toISOString()).toBe('P2M25D')
-    expect(d.asDays()).toBe(85) // moment 85, count 2M as 61 days
-    expect(d.asWeeks()).toBe(12.142857142857142) // moment 12.285714285714286
+    expect(d.asDays()).toBe(85.83333333333333) // moment 86, count 2M as 61 days
+    expect(d.asWeeks()).toBe(12.261904761904763) // moment 12.285714285714286
+    expect(d.asMonths()).toBe(2.8219178082191783) // moment 2.8213721020965523
```
iamkun, line 90: It will be better to not have different results from moment.js.

This will cause a huge problem during migration.

sanjaynishad: already a huge difference with moment.js right now, after my changes it's very close to moment.js result, you can check.

e.g.
existing code
days --> 85 --> moment.js 86
weeks --> 12.142857142857142 --> moment.js 12.285714285714286
months --> 2.8333333333333335 --> moment.js 2.8213721020965523

with my changes (improved and close to moment.js now)
days --> 85.83333333333333
weeks --> 12.261904761904763
months --> 2.8219178082191783

iamkun: seems you are correct

iamkun:
THX

mh1622:
This redefinition of a month from `30 days` to `1 year / 12` should have been included in a major release of dayjs as it is a breaking change (and has led to issues in our production code). I am not against this redefinition (it makes sense), however, I am disappointed that it got released as part of minor release v1.11.10.

Furthermore, it needs to be much clearer in the documentation what dayjs chooses as its definition of a month (when attempting to support ISO-8601 durations), as the specification itself is not explicit (probably hence the initial decision to set it to 30 days).

It would be good to hear how these sorts of breaking changes are going to be identified and managed better in the future.

PR #2383, opened 2023-07-21
fix: fix isDayjs check logic

iamkun:
ref: https://github.com/moment/moment/blob/develop/src/lib/moment/constructor.js#L78

`isDayjs` should return `true` in different versions of dayjs.

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -6,7 +6,10 @@ let L = 'en' // global locale
 const Ls = {} // global loaded locale
 Ls[L] = en
 
-const isDayjs = d => d instanceof Dayjs // eslint-disable-line no-use-before-define
+const IS_DAYJS = '$isDayjsObject'
+
+// eslint-disable-next-line no-use-before-define
+const isDayjs = d => d instanceof Dayjs || !!((d || {})[IS_DAYJS])
 
 const parseLocale = (preset, object, isLocal) => {
   let l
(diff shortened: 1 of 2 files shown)


avand:
@iamkun I'm experiencing what I think is a bug that this ticket may resolve. When I use `dayjs.isDayjs()` together with a plugin like `dayjs.extend(utc)` and organize my code in multiple files, I get `false` when I expect `true. Let me demonstrate:

``` js
// utcDay.js
import dayjs from 'dayjs';
import utc from 'dayjs/plugin/utc';

function utcDay() {
  dayjs.extend(utc);
  return dayjs.utc();
}
```

``` js
// index.js
import utcDay from 'utcDay';

dayjs.isDayjs(utcDay()) // => false
```

Would the changes you made here fix this issue? And if so, when will they be released?

src/index.js
```diff
@@ -6,7 +6,10 @@ let L = 'en' // global locale
 const Ls = {} // global loaded locale
 Ls[L] = en
 
-const isDayjs = d => d instanceof Dayjs // eslint-disable-line no-use-before-define
+const IS_DAYJS = '$isDayjsObject'
+
+// eslint-disable-next-line no-use-before-define
+const isDayjs = d => d instanceof Dayjs || !!(d && d[IS_DAYJS])
```
jBuchholz590, line 12: I have a problem related to this change. We have a Dayjs object that we store in `location.state` where it get's serialized. Upon retrieving said object we use `isDayjs` to check whether the object is actually of type `Dayjs` or if we have to parse it, something along the lines of

```ts
const maybeBrokenDayjs(obj: any) => {
  if (isDayjs(obj) {
    return obj;
  } else if (obj?.$d) {
    return dayjs(obj.$d)
  }
}
```

This now obviously breaks as `isDayjs` will suddenly yield `true`. While this is of course easily fixable it is a breaking change and thus quite unexpected in a Patch-Release :(

Issue #2388, opened 2023-07-27
The "timezone" plugin does not parse the timestamp in seconds


krav-ets:
**Describe the bug**
The "timezone" plugin does not parse the timestamp in seconds
Example: 
`dayjs.tz(1690278073, 'X', "Europe/London").format(); // => "1970-01-20T14:31:18+01:00"`

**Expected behavior**
`dayjs.tz(1690278073, 'X', "Europe/London").format(); // => ""2023-07-25T10:41:13+01:00"`

**Information**
 - Day.js Version [v1.11.9]
 - OS: [Ubuntu 20.04.6]
 - Browser [chrome 115.0.5790.102]
 - Time zone: [GMT+03:00]

PR #2389, opened 2023-07-27
fix: parsing timestamp in seconds in timezone plugin

krav-ets:
solution to issue #2388


iamkun:
Thanks for the sweet fix.

Sorry for the inconvenience. I have to close the PR because it does not trigger the CI job correctly. 
image

Feel free to open a new PR with the same code, please.

PR #2396, opened 2023-08-02
Update Slovenian translation for relative time

alesf:
Slovenian translation update so it handles dual and some plural edge cases.


iamkun:
Thanks

PR #2403, opened 2023-08-08
refac(dayjs): refactoring duration internal constants 

ssi02014:
Hello 👋  @iamkun Thank you for sharing such a great library.

It would be nice to have `MILLISECONDS_A_YEAR` and `MILLISECONDS_A_MONTH` inside `duration` managed together in "constants.js", what do you think?


src/plugin/duration/index.js
```diff
@@ -25,8 +24,8 @@ const unitToMS = {
 
 const isDuration = d => d instanceof Duration // eslint-disable-line no-use-before-define
 
-let $d
-let $u
+let $d = null
+let $u = null
```
ssi02014, line 28: I initialized it to null to explicitly write that the value is currently empty.

okinawaa:
good👍

src/plugin/duration/index.js
iamkun: any reason to initialize it to null, can we simply use `let $d`?

iamkun:
The reason to have a constant in duration plugin rather than in the common constant is that the `src/constant` will be bundled to the core package, not to the plugin, which will increase the package size.

src/plugin/duration/index.js
ssi02014: @iamkun 
Yes, my initial intention was to explicitly initialize null to indicate that there is no current value, but I'll revert back to that.

ssi02014:
@iamkun 
>The reason to have a constant in duration plugin rather than in the common constant is that the src/constant will be bundled to the core package, not to the plugin, which will increase the package size.

Will this increase the package size and cause big problems?
I thought it would be better to manage them together in line with the purpose of constants.

iamkun:
> @iamkun
> 
> > The reason to have a constant in duration plugin rather than in the common constant is that the src/constant will be bundled to the core package, not to the plugin, which will increase the package size.
> 
> Does this make a big difference in package size? I thought it would be better to manage them together in line with the purpose of constants.

This is because not everyone needs the duration plugin, and should not download these constants in the core module

ssi02014:
@iamkun I understand, I thought it was a very small difference. 🙏

PR #2406, opened 2023-08-10
use lazy function replace switch

zhangenming:


iamkun:
any explanation for this PR, please?

IMHO, `() =>` will be compiled to `function () {}` that does not save the bundle size.

iamkun:
The main reason is lazy function will cause the bundle size larger than the switch-case.

zhangenming:
我不太懂这个 可以稍微解释下么? tks~  为什么字典模式有这个缺点

PR #2414, opened 2023-08-15
docs: corrected incorrect Korean translations

keepinblazing:
Hello, I work as a Node developer in Korea. 
Many developers in Korea also use dayjs. 
While looking at the Korean README for dayjs, I noticed some translations seemed off. 
I've corrected the parts that were strangely translated.

### DIFF
--- a/docs/ko/README-ko.md
+++ b/docs/ko/README-ko.md
@@ -23,7 +23,7 @@
     </a>
 </p>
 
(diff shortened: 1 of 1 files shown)


keepinblazing:
@iamkun If it's not too much trouble and when you have the time, could you please review this? I believe this has been accurately translated into Korean, but if you find anything amiss, please do point it out.

iamkun:
Thanks

PR #2093, opened 2022-10-22
Fix Central Kurdish prepositions and move to `ckb`

roj1512:

PR #2417, opened 2023-08-16
Change ku to ckb locale

AlanD20:
Hey there!

I'm opening this PR to add `ckb` locale to the list of supported locales. 

**Why?**
The Kurdish language has a variety of dialects, leading to inconsistencies in many packages. The `ku` locale, for instance, might represent Central Kurdish translation in one package, but in another package, it could be Northern or Southern Kurdish. These dialect differences are significant and could result in unintentional selection of the locale.

Therefore, this pull request proposes a consistent [CLDR](https://cldr.unicode.org/index) locale code by duplicating the existing `ku` locale, which represents Central Kurdish translation, into its own designated locale code.

**Note:** I chose to not delete `ku` locale since some people might still use this in their projects. But in the future, people could reserve this locale for another Kurdish dialect translation.

Thank you! 

**References:**
- [Picking the right language identifier by Unicode CLDR](https://cldr.unicode.org/index/cldr-spec/picking-the-right-language-code)
- [CKB in Ethnologue](https://www.ethnologue.com/language/ckb/)
- [CKB in ISO 639-3](https://iso639-3.sil.org/code/ckb)

### DIFF
--- a/src/locale/ckb.js
+++ b/src/locale/ckb.js
@@ -1,0 +1,92 @@
+// Central Kurdish [ckb]
+import dayjs from 'dayjs'
+
+export const englishToArabicNumbersMap = {
+  1: '١',
+  2: '٢',
+  3: '٣',
+  4: '٤',
+  5: '٥',
+  6: '٦',
+  7: '٧',
+  8: '٨',
+  9: '٩',
+  0: '٠'
+}
+
+const arabicToEnglishNumbersMap = {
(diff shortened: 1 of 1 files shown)


AlanD20:
Sorry there were two other attempts to move `ku` to `ckb` that I wasn't aware of.

I'm closing this PR and hopefully either #2093 or #1798 is going to be merged.

Thank you!

PR #2435, opened 2023-08-31
Add "calendar" to language "fr"

JulesVision:
Add calendar translation to the French language.


iamkun:
Thanks

iamkun:
Can you please fix the lint error?

Also, is it better to keep the same calendar config as moment.js fr locale?

Issue #2434, opened 2023-08-31
`Do` date format works incorrect


iYanovoy:
In format function, when I am using 'Do' date format, it returns Date with letter `o`, instead of `st` or `th`

It should add `th` or `st`, but returns letter `o`

**Information**
 - Day.js Version [1.11.9]
 - OS: [Mac OS 13.4]
 - Browser [Chrome Version 116.0.5845.110]
 - Time zone: [GMT +2]
 
 Code example: 
 dayjs().format("Do MMMM YYYY"); // returns 31o Aug 2023

PR #2437, opened 2023-09-01
Add `Do` date format logic

iYanovoy:
Add `Do` date format logic

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -300,6 +300,8 @@ class Dayjs {
           return this.$D
         case 'DD':
           return Utils.s(this.$D, 2, '0')
+        case 'Do':
+          return Utils.n(Utils.s(this.$D, 2, '0'))
         case 'd':
           return String(this.$W)
         case 'dd':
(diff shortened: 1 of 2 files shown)


Pareder:
What if user changes locale and it is not `en`?

Also I would recommend to cover this format with tests.

You can add `do` and `DDDo` formats as well.

iYanovoy:
Ok, I will think about it. But also have a question, Is plugins
`isSameOrBefore` and `isSameOrAfter` work properly, for now? Because it
brokes our validations.

пн, 4 вер. 2023 р. о 09:50 Anton Cherniavskyi ***@***.***>
пише:

iamkun:
https://github.com/iamkun/dayjs/issues/2434#issuecomment-1702845542

Atulkumar112:
> Add `Do` date format logic

Hi @iYanovoy 

i am using 'Do' but it's given me the wrong value.
```
.format(`dddd, DD Do MMMM YYYY`)   
```
and o/p is Friday, 08 8o December 2023 and it should be Friday, 08th December 2023

please have a look on this.

Issue #2442, opened 2023-09-07
Time validation not working properly for languages does not have AM/PM


ahamedalthaf:
Time formatting not working properly for languages that has different value apart from AM and PM for meridiem.

Ex:

Chinese Language:

Current:
dayjs('9:00 午後', 'h:mm A') - 9:00 午前 (Wrong)

The value should be like below.
dayjs('9:00 午後', 'h:mm A') - 9:00 午後 (Correct)

Pareder:
@iamkun I think the problem is in `CustomParseFormat` plugin. It determines afternoon by such condition https://github.com/iamkun/dayjs/blob/dev/src/plugin/customParseFormat/index.js#L55 (`i > 12`), but `meridiem` method of locale returns `12` for afternoon (https://github.com/iamkun/dayjs/blob/dev/src/locale/ja.js#L24), wo we will get `i === 12`, so `this.afternoon` is `false.

Also there are some locales that have more than two periods, not only AM/PM, for example `zh-cn`, `bn-bd`, etc. and this method is broken for such locales.

ahamedalthaf:
For other locales, there is no value in between 11 and 12 so it should not be a problem. This problem will come only for three languages as of now. 'zh', 'zh-cn', 'zh-tw'

ahamedalthaf:
Have added relevant changes and added tests for the same. Please check once @iamkun 
https://github.com/iamkun/dayjs/pull/2443

josephmathew900:
I'm also facing the same issue with Japanese Language. I think updating `isAfternoon = i >= 12` in [customParseFormat](https://github.com/iamkun/dayjs/blob/dev/src/plugin/customParseFormat/index.js#L55) should fix the bug for Japanese Language. @iamkun, can we update it or will it cause any other problems?

Ref: https://github.com/iamkun/dayjs/pull/2032#issue-1345053433

PR #2443, opened 2023-09-08
Fixes issue #2442 Time validation Fix for other languages

ahamedalthaf:
Fixes below issue
https://github.com/iamkun/dayjs/issues/2442


iamkun:
Thanks for the PR. It fixed the bug in the related issue.

However, is it possible not to add `meridiemHour` function to the locale file?  Instead of using `meridiem` as moment.js does.

PR #2444, opened 2023-09-10
Add usage trend link

cwtuan:
As an active user of Day.js, I believe it would greatly benefit the community if we could include a link to the npm usage trend for Day.js and Moment.js. 

This link will provide users with valuable insights into the usage trend, allowing them to make more informed decisions when choosing a modern date-time library. Providing this information will empower users to choose Day.js as a more modern alternative, considering its thriving community and ongoing development.

I kindly request you to consider this addition. Thanks


iamkun:
Seems good, maybe an embedded chart or image will be better?

iamkun:
how about a screenshot then?

PS: found another website that can make a comparison https://npmtrends.com/dayjs-vs-moment

iamkun:
That looks great, thanks.

Before this PR gets merged, can you add a clickable link to the image the same as the link text above? 

Or, maybe we can remove the text link and use an image combined with a a tag instead?

iamkun:
Cheers

PR #2450, opened 2023-09-19
fix: parsing timestamp in seconds in timezone plugin

krav-ets:
solution to issue https://github.com/iamkun/dayjs/issues/2388

### DIFF
--- a/src/plugin/timezone/index.js
+++ b/src/plugin/timezone/index.js
@@ -1,5 +1,7 @@
 import { MIN, MS } from '../../constant'
 
+const matchTimestampInSeconds = /^\d{10}(?:\.\d{1,3})?$/
+
 const typeToPos = {
   year: 0,
   month: 1,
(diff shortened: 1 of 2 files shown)


krav-ets:
@iamkun  new PR, according to the comment https://github.com/iamkun/dayjs/pull/2389#issuecomment-1725372843

iamkun:
Thanks.

If we want to properly parse `dayjs.tz(1690278073, 'X')` from timestamp to date, maybe updating the logic in the costomParseFormat plugin is better?

Not to hard code the logic in the timezone plugin

krav-ets:
> If we want to properly parse `dayjs.tz(1690278073, 'X')` from timestamp to date, maybe updating the logic in the costomParseFormat plugin is better?

Sorry for the long answer.
Execution does not reach CustomParseFormat plugin, because it falls into this If (https://github.com/iamkun/dayjs/blob/305f54099172ada45d10ba4c9bb5ec95e2a9d441/src/plugin/timezone/index.js#L132), and there it is processed incorrectly, because it contains 10 digits, not 13.

Issue #2501, opened 2023-11-10
UTC plugin: getting incorrect clone after .utcOffset(0, true)


ostgals:
**Describe the bug**
After a zero offset is set while keeping local time the clones I get with `dayjs` constructor are not the same.

Simple reproduction:
```
var d1 = dayjs('2023-10-23 14:00:00').utcOffset(0, true)
var d2 = dayjs(d1)
d1.format() // '2023-10-23T14:00:00Z'
d2.format() // '2023-10-23T11:00:00Z'
```

**Expected behavior**
Both the original instance and its clones must be the same.

**Information**
 - Day.js Version v1.10.4 (but reproducible with the version used on the homepage)
 - OS: Windows 11
 - Browser: Chrome 119
 - Time zone: GMT+03:00

**Possible reason**
After a short investigation I came up to `utc` plugin source and I found out that if the second parameter of `.utcOffset` is set to `true` the plugin forcibly sets UTC flag (`$u` ) to true if the first parameter is zero. That's incorrect because if the flag is `true` then the hours for a clone are initialized with plain `Date.getUTCHours`, but the date in the original instance is stored with the local timezone.

iamkun:
fixed in [#2505](https://github.com/iamkun/dayjs/pull/2505)

PR #2505, opened 2023-11-15
fix: .utcOffset(0, true) result and its clone are different

ostgals:
Actually `.utcOffset(0, flag)` is the same as `.utc(flag)` but `.utcOffset` with `keepLocalTime = true` didn't modify the internal date accordingly. That caused a bug described in #2501

Fixed `.utcOffset` by returning `.utc(keepLocalTime)` if passed offset is zero.

Added some tests to cover possible mismatches.


iamkun:
Thanks

Issue #2386, opened 2023-07-26
the updateLocale Plugin should merge config rather than replace it


lanmch:
**Describe the bug**
updateLocal Plugin set default Value

**Expected behavior**
the updateLocale Plugin should merge config rather than replace it
![image](https://github.com/iamkun/dayjs/assets/37408756/104a5c0a-ba1a-428b-a37b-7c2ac780c6cd)

**Information**
 - Day.js Version 1.11.9
 - OS: mac
 - Browser chrome 100+

lanmch:
eg.
in moment, i try to updateLocale relative time, it will just replace the future key.
but in dayjs, i try to updateLocale relative time, it replace the whole object.

```javascript
moment.updateLocale('zh-cn', {
    relativeTime: {
         future: '%s test in the future'
    }
})
dayjs.updateLocale('zh-cn', {
    relativeTime: {
         future: '%s test in the future'
    }
})
```
moment and dayjs
![image](https://github.com/iamkun/dayjs/assets/37408756/b244bdac-229f-4783-aebf-93e117725884)
![image](https://github.com/iamkun/dayjs/assets/37408756/b8a464fb-cbd9-41f1-b8b6-2bc3dbcae283)

iamkun:
nice catch, thanks. PR is welcome

PR #2558, opened 2024-01-20
chore: merge config rather than replace

liuseen-l:
close: #2386

### DIFF
--- a/src/plugin/updateLocale/index.js
+++ b/src/plugin/updateLocale/index.js
@@ -5,7 +5,16 @@ export default (option, Dayjs, dayjs) => {
     if (!localeConfig) return
     const customConfigKeys = customConfig ? Object.keys(customConfig) : []
     customConfigKeys.forEach((c) => {
-      localeConfig[c] = customConfig[c]
+      const originConfig = localeConfig[c]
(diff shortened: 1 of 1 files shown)


src/plugin/updateLocale/index.js
```diff
@@ -5,7 +5,16 @@ export default (option, Dayjs, dayjs) => {
     if (!localeConfig) return
     const customConfigKeys = customConfig ? Object.keys(customConfig) : []
     customConfigKeys.forEach((c) => {
-      localeConfig[c] = customConfig[c]
+      const originConfig = localeConfig[c]
+      const customOriginConfig = customConfig[c]
+      if (originConfig !== null && typeof originConfig === 'object' && typeof customOriginConfig === 'object') {
+        localeConfig[c] = {
+          ...originConfig,
```
iamkun, line 12: Thanks for the PR.

Day.js is a lib targeting es5, please do not use es6 syntax like `...` (Destructuring assignment) or `Object.assign` in the code. This will lead significant bundle size increase.

Please consider using some simple code like `for` loop or `Object.keys` to iterate the object.

Issue #2582, opened 2024-02-16
Dayjs subtract does not support week based dayjs durations => expected?


yassuge:
**Describe the bug**
Dayjs subtract does not support week based dayjs durations => expected?

 x=dayjs()
 x.toISOString()
>> '2024-02-16T12:14:02.958Z'

x.subtract(1,'weeks').toISOString()
>> '2024-02-09T12:14:02.958Z' // expected
x.subtract(dayjs.duration('P2W')).toISOString()
>> '2024-02-16T12:14:02.958Z' // !!! NO IMPACT => IS THIS NOT SUPPORTED?
x.subtract(dayjs.duration('P3M')).toISOString()
>> '2023-11-16T12:14:02.958Z' // expected

**Information**
 - Day.js Version dayjs@npm:1.11.10
 - Browser [e.g. chrome 62]
 - Time zone: [e.g. GMT-07:00 DST (Pacific Daylight Time)]

Issue #2479, opened 2023-10-16
Adding weeks in duration plugin is broken


manchicken:
**Describe the bug**

I haven't upgraded since `1.11.7` because of various bugs, and this is the big one now:

Sample code:

```javascript
const dayjs = require('dayjs')
dayjs.extend(require('dayjs/plugin/duration'))

const today = dayjs()
const twoWeeksFromNow = dayjs().add(dayjs.duration({weeks:2}))

console.log({today,twoWeeksFromNow})
```

Here is what I get in `1.11.10`:

```
❯ node ./dayjs-bug.js
{
  today: M {
    '$L': 'en',
    '$d': 2023-10-16T17:38:24.875Z,
    '$y': 2023,
    '$M': 9,
    '$D': 16,
    '$W': 1,
    '$H': 13,
    '$m': 38,
    '$s': 24,
    '$ms': 875,
    '$x': {},
    '$isDayjsObject': true
  },
  twoWeeksFromNow: M {
    '$L': 'en',
    '$d': 2023-10-16T17:38:24.875Z,
    '$y': 2023,
    '$M': 9,
    '$D': 16,
    '$W': 1,
    '$H': 13,
    '$m': 38,
    '$s': 24,
    '$ms': 875,
    '$x': {},
    '$isDayjsObject': true
  }
}
```

**Expected behavior**

I expect this calculation to happen similarly to how it does in `1.11.7`:

```
❯ node ./dayjs-bug.js
{
  today: M {
    '$L': 'en',
    '$d': 2023-10-16T17:41:20.524Z,
    '$x': {},
    '$y': 2023,
    '$M': 9,
    '$D': 16,
    '$W': 1,
    '$H': 13,
    '$m': 41,
    '$s': 20,
    '$ms': 524
  },
  twoWeeksFromNow: M {
    '$L': 'en',
    '$d': 2023-10-30T17:41:20.524Z,
    '$x': {},
    '$y': 2023,
    '$M': 9,
    '$D': 30,
    '$W': 1,
    '$H': 13,
    '$m': 41,
    '$s': 20,
    '$ms': 524
  }
}
```

**Information**
 - Day.js Version: 1.11.10
 - OS: macOS 14.0
 - Browser: Node in the CLI
 - Time zone: `America/New_York`

TonyRL:
Possible duplicate of #2464

vad99lord:
looks like #2583 addresses this issue, if I'm not mistaking

[...]

Issue #2464, opened 2023-09-26
Duration is wrong on 1.11.10


adrien-may:
**Describe the bug**
The duration is wrong.

Let's take this example

```javascript
// Init
const dayjs = require('dayjs')
const duration = require( 'dayjs/plugin/duration');

dayjs.extend(duration);

// two dates
const first = dayjs('2020-01-01T12:00:00.000Z')
const start = dayjs('2020-01-01T12:00:00.000Z').subtract(10, 'weeks')

// compute duration
const milliseconds = first.diff(start, 'milliseconds')
dayjs.duration(milliseconds).toISOString()
```

With dayjs 1.11.09, we get 'P2M10DT1H'
With dayjs 1.11.10 we get 'P2M9DT5H' which I don't even understand

**Expected behavior**

Previous behavior is right

**Information**
 - Day.js Version 1.11.09 vs 1.11.10

bombillazo:
Any update on this error?

pvds:
@iamkun @sanjaynishad can this behaviour be caused by #2362?

sanjaynishad:
@pvds I don't think, because duration was broken in `1.11.9` as well, and maybe someone pushed the fixes and that's causing the issue

#2479 when I tried this with `1.11.9`

```
const today = dayjs()
const twoWeeksFromNow = dayjs().add(dayjs.duration({ weeks: 2 }))

console.log({ today, twoWeeksFromNow })
```

I got 

```
{
  today: M {
    '$L': 'en',
    '$d': 2023-11-30T14:03:05.248Z,
    '$x': {},
    '$y': 2023,
    '$M': 10,
    '$D': 30,
    '$W': 4,
    '$H': 19,
    '$m': 33,
    '$s': 5,
    '$ms': 248
  },
  twoWeeksFromNow: M {
    '$L': 'en',
    '$d': Invalid Date,
    '$x': {},
    '$y': NaN,
    '$M': NaN,
    '$D': NaN,
    '$W': NaN,
    '$H': NaN,
    '$m': NaN,
    '$s': NaN,
    '$ms': NaN
  }
}
```

I'll try to investigate more on this

yanny7:
Issue is caused by wrong parsing from milliseconds 

```
    _proto.parseFromMilliseconds = function parseFromMilliseconds() {
        var $ms = this.$ms;
        this.$d.years = roundNumber($ms / MILLISECONDS_A_YEAR);
[...]

PR #2583, opened 2024-02-17
fix: missing handling of week on manipulating duration

adlanarifzr:
This fix solve issue #2582

### DIFF
--- a/src/plugin/duration/index.js
+++ b/src/plugin/duration/index.js
@@ -266,6 +266,7 @@ const manipulateDuration = (date, duration, k) =>
     .add(duration.minutes() * k, 'm')
     .add(duration.seconds() * k, 's')
     .add(duration.milliseconds() * k, 'ms')
+    .add(duration.weeks() * k, 'w')
 
 export default (option, Dayjs, dayjs) => {
   $d = dayjs


vad99lord:
does this pr address #2479 , #2464?

manchicken:
> does this pr address #2479 , #2464?

I hope so? I’d love to be able to update.

iamkun:
Thanks, can you please fix the failed unit test before this PR gets merged?

vad99lord:
@iamkun, could you please clarify what is the logic behind `as` and `get` methods in the duration plugin? 

Is my understanding correct that **as** converts the whole duration to the specified unit, ie how many weeks does this instance represent? 
And **get** retrieves only a part of the duration relevant to the instance unit, i.e I've provided "2 weeks" as part of the initial duration string => the method returns 2? 

I'm not quite sure that I understand **get** method for weeks fully, because currently there seems to be a specific case for the [weeks](https://github.com/iamkun/dayjs/blob/3737c8c35459562d4e345136cbb83290e88c6692/src/plugin/duration/index.js#L198) that basically makes this method act as **as** converting the whole duration in terms of weeks.

There is also a difference in parsing between milliseconds input not using weeks and other types of inputs accepting weeks.
With that said, I've suggested changes for weeks logic in the **get** method, hoping this may speed up merging this pr.

```js
  get(unit) {
    let base = this.$ms
    const pUnit = prettyUnit(unit)
    if (pUnit === 'milliseconds') {
      base %= 1000
    } else if (pUnit === 'weeks') {
      const weeks = this.$d[pUnit] || 0
      const days = this.$d.days || 0
      base = weeks + roundNumber(days / 7)
    } else {
      base = this.$d[pUnit]
    }
    return base || 0 // a === 0 will be true on both 0 and -0
  }
```

Issue #1713, opened 2021-11-25
Abbreviated timezone from `advancedFormat` not showing


p6l-richard:
**Describe the bug**
Using the advanced formatting for the abbreviated timezone outputs the offset instead of the abbreviation of the name:
`GMT+1` instead of `CET` for `Europe/Berlin`

Please see the behavior in [this stackblitz](https://stackblitz.com/edit/js-cp8b7x?file=index.js)

**Expected behavior**
`dayjs().tz('Europe/Berlin').format('hh:mmA z');`
should output e.g. 
`11:07AM CET`

**Information**
 - Day.js Version [1.10.7]
 - OS: [Windows]
 - Browser [Version 96.0.4664.45 (Official Build) (64-bit)]
 - Time zone: [e.g. GMT+1 CET (Central European Standard Time)]

vibhorpahare:
I am facing the same issue 
Dayjs version : [1.10.6] 
time zone : IST(Indian slandered time).

m-heinemann:
I am facing the same issue
Dayjs version: [1.11.7]
time zone: most of them!

Bhavin789:
I am facing the same issue. Any idea when this would be fixed or at least have the docs updated? Thanks. cc// @iamkun 
<img width="218" alt="image" src="https://github.com/iamkun/dayjs/assets/15359575/a42a9b2e-bbd0-40ba-89a4-425d76920605">

phalkone:
I know this is an old topic, but I am facing the same issue. Would have loved to use this feature.

MikePrus:
The same thing, somebody found a solution for this?

borkanee:
Same issue here, did anyone find a solution?

stefanobartoletti:
I wanted to open an issue for this, but I see that it is an already known problem.

I can only add that it still happens with version `1.11.10`

hollandmatt:
I'm having the same issue with the `AEST` (Australian Eastern Standard Time) TZ. Present in version `1.11.12`.

PR #2637, opened 2024-04-23
Fix: Advanced format output with short offset name

klm-turing:
- Used of long timezone name by default in `Intl.DateTimeFormat` 
- Extract and cached short offset name from the long one in the `offsetName` function 
- Writing related unit test

This fix [1713](https://github.com/iamkun/dayjs/issues/1713) issue

### DIFF
--- a/src/plugin/timezone/index.js
+++ b/src/plugin/timezone/index.js
@@ -12,23 +12,26 @@ const typeToPos = {
 // Cache time-zone lookups from Intl.DateTimeFormat,
 // as it is a *very* slow method.
 const dtfCache = {}
-const getDateTimeFormat = (timezone, options = {}) => {
-  const timeZoneName = options.timeZoneName || 'short'
-  const key = `${timezone}|${timeZoneName}`
-  let dtf = dtfCache[key]
+const getDateTimeFormat = (timeZone) => {
+  /* Long output is always correct whatever the timezone.
+  *  and whatever the locale, en-US / en-GB etc...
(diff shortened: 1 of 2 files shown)


et:
Thanks for adding this! I definitely could use this in a moment migration. 
@iamkun - curious if we can get this in?

Alternatively, have we considered replicating `moment-timezone`'s `zoneAbbr` functionality? 
https://momentjs.com/timezone/docs/#/using-timezones/formatting/

src/plugin/timezone/index.js
```diff
@@ -107,11 +110,21 @@ export default (o, c, d) => {
...
     // type: short(default) / long
     const zone = this.$x.$timezone || d.tz.guess()
+    const cachedOffsetName = this.$x.$offsetName || {}
+    const cacheKey = `${zone}-${type}}`
+    if (cachedOffsetName[cacheKey]) {
+      return cachedOffsetName[cacheKey]
+    }
     const result = makeFormatParts(this.valueOf(), zone, { timeZoneName: type }).find(m => m.type.toLowerCase() === 'timezonename')
-    return result && result.value
+    if (type === 'short') {
```
alexedwinroberts, line 122: Thanks for putting this PR together, I found it really helpful! 🎉  

There's a small issue where the `value` for Istanbul is 'GMT+03:00' so the returned abbreviation becomes `G`.

Maybe you could add a check to see if there are any digits in the value and then return the value instead of abbreviating it?

Issue #2604, opened 2024-03-13
Error in parsing array format date


hbalxzdl:
**Describe the bug**
```
import dayjs from 'dayjs'
import arraySupport from 'dayjs/plugin/arraySupport'
dayjs.extend(arraySupport)

use:
dayjs([2010, 1, 14]).format('YYYY/MM/DD')) return  2010/02/14

```

**Expected behavior**
```
dayjs([2010, 1, 14]).format('YYYY/MM/DD')) return  2010/01/14
```
**Information**
 - Day.js Version 1.11.8
 - OS: Mac OS
 - Browser: Google Chrome v115
 - Time zone: Eastern Daylight Time

PR #2639, opened 2024-04-24
Fix: Error in parsing array format date

klm-turing:
- Add a `getStringDate` function which parse an array of date and return a string to dayjs
- Add related test units

This fix [2604](https://github.com/iamkun/dayjs/issues/2604) issue

### DIFF
--- a/test/plugin/arraySupport.test.js
+++ b/test/plugin/arraySupport.test.js
@@ -33,12 +33,26 @@ const testArrs = [
   [2010, 6, 10]
 ]
 
-describe('parse array local', () => {
-  testArrs.forEach((testArr) => {
-    it(testArr, () => {
-      expect(dayjs(testArr).format())
-        .toBe(moment(testArr).format())
-    })
+describe('parse full array local', () => {
+  it('parse and format full date array', () => {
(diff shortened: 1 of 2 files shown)


test/plugin/arraySupport.test.js
```diff
@@ -33,12 +33,26 @@ const testArrs = [
   [2010, 6, 10]
 ]
 
-describe('parse array local', () => {
-  testArrs.forEach((testArr) => {
-    it(testArr, () => {
-      expect(dayjs(testArr).format())
-        .toBe(moment(testArr).format())
```
iamkun, line 40: we should keep the same return with moment.js. otherwise, it will be a huge breaking change.

klm-turing: The fact is the same bug is present in « moment ». JavaScript Date parses month number from zero to eleven. But most of the time if not all time, we use month from 1 to 12. Excepting the same result as moment will be impossible unless moment fix it. Another alternative is to update the docs and let users know that they need to maybe change their habits and consider zero as first month and 11 as the last month in order to properly use arraySupport plugin. What would you suggest?

PR #2693, opened 2024-07-18
fix: timezone plugin currect parse UTC tz

iamkun:
fix https://github.com/iamkun/dayjs/issues/2248


zhishaofei3:
starting from this fix, the problem of cross-time zone was introduced. If it is Germany, one hour will be calculated less when crossing the domain during winter time.

franco-marzano:
This problem appear to have been reintegrated in version 1.11.14

PR #2704, opened 2024-08-07
update locale types and add getTimezone from Dayjs instance

LiuTing0403:
- update locale types
-  add method to get current timezone name from Dayjs instance


ray007:
How about returning the timezone with a `instance.tz()` call without parameters?

PR #2709, opened 2024-08-15
feat: add customParseFormat support for isoweek

amoydavid:
1. add customParseFormat support for isoweek(W and WW)
2. fix test for weekOfYear

### DIFF
--- a/src/plugin/customParseFormat/index.js
+++ b/src/plugin/customParseFormat/index.js
@@ -1,6 +1,6 @@
 import { u } from '../localizedFormat/utils'
 
-const formattingTokens = /(\[[^[]*\])|([-_:/.,()\s]+)|(A|a|Q|YYYY|YY?|ww?|MM?M?M?|Do|DD?|hh?|HH?|mm?|ss?|S{1,3}|z|ZZ?)/g
+const formattingTokens = /(\[[^[]*\])|([-_:/.,()\s]+)|(A|a|Q|YYYY|YY?|ww?|WW?|MM?M?M?|Do|DD?|hh?|HH?|mm?|ss?|S{1,3}|z|ZZ?)/g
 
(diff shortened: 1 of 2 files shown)


src/plugin/customParseFormat/index.js
```diff
@@ -100,6 +100,8 @@ const expressions = {
   }],
   w: [match1to2, addInput('week')],
   ww: [match2, addInput('week')],
+  W: [match1to2, addInput('week')],
+  WW: [match2, addInput('week')],
```
iamkun, line 104: seems there no logic dealing with ISOWeek?

amoydavid: In the customParseFormat scenario, similar to the code in pull request #2705, it's sufficient to support parsing uppercase W/WW for ISO week notation. For stricter validation, one could additionally check if the week number exceeds 53. However, this basic approach is adequate for most use cases.

Take, for instance, date strings like '2024-W24' or '2024-w24'. In these cases, it's ambiguous whether '24' denotes a regular weekday or an ISO week number.

Consider the following practical scenario: Suppose we generate a string using the format 'YYYY-[W]WW'. Later, we need to verify if a given string adheres to this format. In such cases, we would typically employ customParseFormat, reusing the original format string. This is where support for 'WW' in ISO week notation becomes crucial. (It's worth noting that using 'ww' would yield consistent results in parsing scenarios, but 'WW' maintains format consistency.)

```
const fileDateFormat = "YYYY-[W]WW";

function buildFileName(date) {
   return dayjs(date).format(fileDateFormat);
}

function checkNoteName(fileName) {
  // here we reused the fileDateFormat
  const parsedDate = dayjs(fileName, fileDateFormat);
  return parsedDate && parsedDate.format(fileDateFormat) === fileName;
}
```

PR #2723, opened 2024-09-10
Update hr.js

tomastan:
Fixing Croatian translations of relativeTime

image


alexandermirzoyan (approved):

tomastan:
We have noticed that selection of the month name is incorrect due to wrong localized format detection regexp. So modified this as well.
Here is an example, how existing test should fail: https://jsfiddle.net/k6Lze0ta/

PR #2725, opened 2024-09-13
fix: test case failed issue fixed in the tesstzone.test.js

saddamhr:
### There was a test that failed in the `timezone.test.js`. I've fixed it

- Before (1 failed test case)
Screenshot 2024-09-13 at 10 46 45 PM

- After (all test cases pass)
image

### DIFF
--- a/test/plugin/timezone.test.js
+++ b/test/plugin/timezone.test.js
@@ -308,7 +308,7 @@ describe('CustomPraseFormat', () => {
 
 describe('startOf and endOf', () => {
   it('corrects for timezone offset in startOf', () => {
-    const originalDay = dayjs.tz('2010-01-01 00:00:00', NY)
+    const originalDay = dayjs.tz('2009-12-31 00:00:00', NY)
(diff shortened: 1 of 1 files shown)


test/plugin/timezone.test.js
```diff
@@ -308,7 +308,7 @@ describe('CustomPraseFormat', () => {
 
 describe('startOf and endOf', () => {
   it('corrects for timezone offset in startOf', () => {
-    const originalDay = dayjs.tz('2010-01-01 00:00:00', NY)
```
alexandermirzoyan, line 311: Do you have any thoughts why it was being failed with this date?

saddamhr: The test failed because dayjs's startOf('day') recalculates the start of the day in local time, which can cause timezone offset changes. On '2010-01-01', the timestamp was adjusted due to the way dayjs handles timezone shifts, leading to a mismatch.

alexandermirzoyan (approved):

PR #2753, opened 2024-10-20
Fix: performance issue in tz plugin

ikeyan:
In this PR, I have improved the performance of the timezone plugin.
I didn't adopt the previous PR #2019 because [it contained a breaking change](https://github.com/iamkun/dayjs/pull/2019#issuecomment-1211723494) and [it introduced a bug](https://github.com/iamkun/dayjs/pull/2019#issuecomment-1662043227).
I also shrunk the tz plugin after gzip from 1110bytes to 991bytes.

The changes are summarized below:

* `getDateTimeFormat`
  * Use `hourCycle: 'h23'` so we can use single DateTimeFormat options and remove [this workaround](https://github.com/iamkun/dayjs/blob/dev/src/plugin/timezone/index.js#L57-L60C32)
  * Set all parts to numeric
* `tzToFirstOffsetCache`
  * first guess of timezone offset in fixOffset (moved from [here](https://github.com/iamkun/dayjs/blob/dev/src/plugin/timezone/index.js#L138))
* change code to make it more compressible

### DIFF
--- a/src/plugin/timezone/index.js
+++ b/src/plugin/timezone/index.js
@@ -1,4 +1,4 @@
-import { MIN, MS } from '../../constant'
+import { INVALID_DATE_STRING, MILLISECONDS_A_MINUTE, MIN, MS } from '../../constant'
 
 const typeToPos = {
   year: 0,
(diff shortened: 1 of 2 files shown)


vaakian:
same issue, why there is no any progress? the performance really sucks

PR #2780, opened 2024-11-30
feat(dayjs): Optimize performance by reducing unnecessary computation…

Jerry965:
In most cases, users will not obtain the UTC difference. I think it can be moved to the switch case and then generated to reduce unnecessary code running.


alexandermirzoyan (approved):
Nice catch!

Ghostz14 (approved):

PR #2800, opened 2025-01-03
Add isoWeekYear setter for Day.js ISO week handling

FAL-coffee:
This pull request introduces an isoWeekYear setter, allowing the ISO week year to be adjusted while preserving the current isoWeek and isoWeekday values. This change ensures consistent ISO-based date calculations and avoids unexpected behavior when using native .year() in combination with ISO week methods.


FAL-coffee:
@iamkun 
We’re encountering a bug in certain cases when using the ISO plugin. Could you please review it? 
Thank you in advance 🙏

Issue #2810, opened 2025-01-16
When using .add(1,'month') month becomes correct but year becomes previous year. 2023-01-01 -> +1 month -> 2022-02-01


sbdd:
**Describe the Bug**  
When using `dayjs.utc(...).tz('utc').startOf('day')` and then adding a month (`.add(1, 'month').startOf('month')`), the resulting date sometimes jumps back by a year if the start date is January 1st. This only happens in the browser and not in Node.js.  It also seems to only happen when you have a utc instances and call .tz('utc') on it.

**Expected Behavior**  
The date should stay in the correct year without rolling back.  

**Steps to Reproduce**  
1. Initialize the date with `dayjs.utc('2023-01-01').tz('utc').startOf('day')`.  
2. Add one month with `.add(1, 'month').startOf('month')`.  
3. Note how the year can shift back.  

**Demo**  
[[StackBlitz Link](https://stackblitz.com/edit/dayjs-playground-pcr9xsfg?file=index.js)](https://stackblitz.com/edit/dayjs-playground-pcr9xsfg?file=index.js)

**Environment**  
- Day.js Version: 1.11.13
- OS: Tested on mac os
- Browser: Chrome
- Time Zone: Europe/Oslo

Code from stackblitz:

````
import './style.css';
import dayjs from 'dayjs';
import timezone from 'dayjs/plugin/timezone.js';
import utc from 'dayjs/plugin/utc.js';
dayjs.extend(utc);
dayjs.extend(timezone);

const doDateStuff = (start = '2023-01-01') => {
  const startDate = dayjs.utc(start).tz('utc');

  let current = startDate;
  const dates = [];

  for (let i = 0; i < 10; i++) {
    dates.push(current.format());
    current = current.add(1, 'month').startOf('month');
  }

  return {
    startDate: startDate.format(),
    dates,
  };
};

const doDateStuffBroken = (start = '2023-01-01') => {
  const startDate = dayjs.utc(start).tz('utc').startOf('day'); // broken because of startOf it seems

  let current = startDate;
  const dates = [];

  for (let i = 0; i < 10; i++) {
    dates.push(current);
    current = current.add(1, 'month').startOf('month');
  }

  return {
    year: startDate.year(),
    startDate: startDate.format(),
    dates,
  };
};

const datesCalc = doDateStuff();
const datesCalcBroken = doDateStuffBroken();

// Write Javascript code!
const appDiv = document.getElementById('app');
appDiv.innerHTML = `
[...]

PR #2813, opened 2025-01-17
Fixes Issue #2810: Inconsistent Behavior with add after startOf call

github-roushan:
This fixes the issue: https://github.com/iamkun/dayjs/issues/2810 by correctly calling the right code when timezone is present as object in DayJS obj


github-roushan:
@iamkun 

can you please merge this trivial pr

PR #2827, opened 2025-02-17
feat: add customParseFormat support for isoweek, the usage is consistent with moment

lizhihangdev:
Background
After using dayjs to replace moment, I found that dayjs does not support moment's ability to parse ISO 8601 strings.

For example
`dayjs('2024-W01', 'GGGG-[W]WW')`

Solution
`dayjs(newDate).add(7, 'day').isoWeek(isoWeek).isoWeekday(1)`

- Get the correct ISO year,`dayjs(newDate).add(7, 'day')`
- Generate dayjs according to ISO week,`dayjs(newDate).add(7, 'day').isoWeek(isoWeek)`
- Take the value of Monday as the final result,`dayjs(newDate).add(7, 'day').isoWeek(isoWeek).isoWeekday(1)`


lizhihangdev:
dependent [IsoWeek ](https://day.js.org/docs/en/plugin/iso-week)plugin

PR #2858, opened 2025-04-15
Add module and exports field to package.json

lennondotw:
## Summary

Add `module` and `exports` fields to `package.json` for proper ESM support.

## Changes

- Add `"module": "./esm/index.js"` for bundler ESM resolution
- Add `"exports"` map covering root, `./esm`, `./plugin/*`, `./esm/plugin/*`, `./locale/*`, `./esm/locale/*`
- Normalize `main` and `types` paths to use relative `./` prefix

## Benefits

- Modern bundlers (Vite, webpack 5, Rollup) can resolve ESM entry points correctly
- Node.js conditional exports work for both `import` and `require`
- Fully backward compatible — no breaking changes

### DIFF
--- a/package.json
+++ b/package.json
@@ -2,8 +2,57 @@
   "name": "dayjs",
   "version": "0.0.0-development",
   "description": "2KB immutable date time library alternative to Moment.js with the same modern API ",
-  "main": "dayjs.min.js",
-  "types": "index.d.ts",
+  "main": "./dayjs.min.js",
+  "module": "./esm/index.js",
+  "types": "./index.d.ts",
+  "exports": {
+    ".": {
+      "import": "./esm/index.js",
+      "require": "./dayjs.min.js",
+      "types": "./index.d.ts"
+    },
+    "./esm": {
+      "import": "./esm/index.js",
+      "types": "./esm/index.d.ts"
+    },
+    "./esm/index.js": {
+      "import": "./esm/index.js",
+      "types": "./esm/index.d.ts"
+    },
+    "./plugin/*": {
+      "require": "./plugin/*.js",
+      "types": "./plugin/*.d.ts"
+    },
+    "./plugin/*.js": {
+      "require": "./plugin/*.js",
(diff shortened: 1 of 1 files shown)


haakonflatval-cognite:
@iamkun I'm seeing problems in some ESM-projects projects I'm working on that require extra patching of import paths from `day.js` in our config files due to malformed/missing export fields in the `package.json`, which is what this PR is trying to address. It would be greatly appreciated if you could consider merging this or one of the similar ones (e.g. https://github.com/iamkun/dayjs/pull/2543/files)

lennondotw:
Rebased on latest `dev`, squashed to a single commit, and cleaned up the exports map — root entry now properly resolves ESM imports (no more `"import": null`), and removed the unintended `packageManager` field.

@iamkun would appreciate a look when you have a moment — this addresses the ESM import issues mentioned by @haakonflatval-cognite above.

lennondotw:
@iamkun I rebased this PR onto the latest `dev` and re-checked whether the issue still exists.

Current `dev` and the latest npm package (`dayjs@1.11.21`) still only expose `main`/`types`, with no `module` or `exports`. In a Node ESM fixture, `dev` still fails for extensionless ESM subpaths such as:

- `import 'dayjs/esm'` -> `ERR_UNSUPPORTED_DIR_IMPORT`
- `import 'dayjs/esm/plugin/utc'` -> `ERR_UNSUPPORTED_DIR_IMPORT`
- `import 'dayjs/esm/locale/zh-cn'` -> `ERR_MODULE_NOT_FOUND`

The updated PR resolves those to the ESM files, while keeping existing CommonJS and explicit file subpaths working, including `require('dayjs')`, `require('dayjs/plugin/utc')`, `require('dayjs/plugin/utc.js')`, `require('dayjs/locale/zh-cn.js')`, and `require('dayjs/package.json')`.

So this is still relevant for ESM consumers, and the branch is now current with `dev`.

PR #2880, opened 2025-05-21
Unify chinese locales

guoyunhe:
`zh` locale is identical to `zh-cn`, `zh-hk` should be identical to `zh-tw`. Unify locales by importing can make bundle size smaller and easier to maintain strings.

### DIFF
--- a/src/locale/zh-hk.js
+++ b/src/locale/zh-hk.js
@@ -1,66 +1,12 @@
 // Chinese (Hong Kong) [zh-hk]
 import dayjs from 'dayjs'
+import tw from './zh-tw'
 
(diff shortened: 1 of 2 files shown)


iamkun:
please fix the lint error

guoyunhe:
@iamkun fixed

iamkun:
image
After testing the build, I found that the spread operator (...) introduces a lot of compatibility-related JavaScript code in the output, which increases the bundle size. Therefore, it might be better to avoid this change. Although maintaining the source code could be slightly more cumbersome, the resulting bundle size would be more ideal.

PR #2885, opened 2025-06-08
Fix for UTC plugin type definitions in combination with CustomParseFormat plugin

WRMSRwasTaken:
Using the `CustomParseFormat` plugin together with the `UTC` plugin seems possible:

```js
dayjs.extend(utc);
dayjs.utc('2025-4-05', ['YYYY-M-D', 'YYYY-MM-D', 'YYYY-M-DD', 'YYYY-MM-DD'], true).day();
// 5
dayjs.utc('2025-04-5', ['YYYY-M-D', 'YYYY-MM-D', 'YYYY-M-DD', 'YYYY-MM-DD'], true).day();
// 5
```

But the types don't line up, giving me an error in my TS project: 
webstorm64_qWojpP1hqZ

Seems to be an easy fix by just orienting on what's already defined as function parameters in `index.d.ts`.

### DIFF
--- a/.github/workflows/check.yml
+++ b/.github/workflows/check.yml
@@ -1,26 +1,12 @@
 name: Lint & Unit Test
 on:
   push:
-    branches: [ master, dev ]
+    branches: [master, dev]
   pull_request:
     branches: [dev]
+
 jobs:
   check:
-    runs-on: ubuntu-latest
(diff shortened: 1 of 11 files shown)


iamkun:
Thanks

PR #2894, opened 2025-06-21
Little improvements on npm commands + Upgrade of cross-env

Genarito:
Hi! First of all, thanks for this amazing project! 

This pull request introduces performance enhancements to the `daysInMonth` function as well as a few improvements in the `package.json` file. The changes are divided into two categories:

### Updates in `daysInMonth` and leap year handling:

Updated the `daysInMonth` function to use a new `MONTH_DAYS` constant and the `yearIsLeap` utility (I've centralized a function to use here and in the `isLeapYear` plugin) for determining the number of days in February during leap years. This is simpler than calling the `startOf` method and doing complex stuff.

Also, I've improved the Typescript definition of the `daysInMonth` function to return the correct possible numbers `28 | 29 | 30 | 31`.

### Improvements in `package.json`:

- Simplified the `lint` script by removing the relative path to `eslint`.
- Updated the `test` script to use `cross-env` for better cross-platform compatibility in the [`package.json`](diffhunk://#diff-7ae45ad102eab3b6d7e7896acd08c427a9b25b346470d7bc6507b6481575d519L8-R10) file.
- Updated `cross-env` to version `5.2.1`.

Hope it helps 😄

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -6,22 +6,6 @@ let L = 'en' // global locale
 const Ls = {} // global loaded locale
 Ls[L] = en
 
-/** Index to get the number of days */
-const MONTH_DAYS = {
-  0: 31, // January
-  1: 28, // February
-  2: 31, // March
-  3: 30, // April
-  4: 31, // May
-  5: 30, // June
-  6: 31, // July
-  7: 31, // August
-  8: 30, // September
-  9: 31, // October
-  10: 30, // November
-  11: 31 // December
-}
-
 const IS_DAYJS = '$isDayjsObject'
 
 // eslint-disable-next-line no-use-before-define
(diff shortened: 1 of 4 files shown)


src/index.js
```diff
@@ -6,6 +6,22 @@ let L = 'en' // global locale
...
+  4: 31, // May
+  5: 30, // June
+  6: 31, // July
+  7: 31, // August
+  8: 30, // September
+  9: 31, // October
+  10: 30, // November
+  11: 31 // December
+}
+
```
iamkun, line 24: Introducing the MONTH_DAYS constant might increase the bundle size, which may not be an ideal optimization?

iamkun: I've noticed this PR introduces some additional code into both `utils` and `index`.

A core principle of Day.js is its minimal bundle size, and we're highly sensitive to any changes that might increase it. For a change like this to be a worthwhile trade-off, we would need to see a significant performance improvement over the current `endOf` implementation.

Have you conducted any performance benchmarks to demonstrate the gains of this new approach? This data would be very helpful for us to evaluate if the change is a net positive for the project.

Looking forward to your thoughts!

Genarito:
Hi!

I have discarded the changes related to calculating the days of the month, as I don't have time to measure whether the library now weighs 1 byte more. I left the `package.json` and `cross-env` changes that simplify the execution of tests on all operating systems. If they are useful, I will be delighted to have contributed to the project; otherwise, feel free to close the PR. I have also adjusted the title of the PR.

Best regards and thank you!

Issue #2882, opened 2025-05-27
date UTC format is wrong


ykx456654:
**Describe the bug**

![Image](https://github.com/user-attachments/assets/471d5ef3-7658-4dc7-ace3-a1c2777b0668)
`dayjs('2025-01-31', 'UTC').startOf('month').add(3, 'month').format('YYYY-MM-DD')`
//  2024-04-30
why the result is 2024-04-30?  it should be 2025-04-01?
**Expected behavior**

**Information**
 - Day.js 1.11.12
 - OS: MacOS
 - Browser Chrome  137.0.7151.41 (Official Build) (arm64)
 - Time zone: Asia/Shangehai

tupb-ici:
@ykx456654 
I faced the same issue.
I don't know why but we can use utcOffset(0) like this as temporary solution.
https://stackoverflow.com/questions/66705548/dayjs-startofday-shifts-two-hours-back

PR #2895, opened 2025-06-23
fix: fix startOf + add chain bug in timezone plugin

eunjios:
#### Problem

- https://github.com/iamkun/dayjs/issues/2882
- I found this issue where `startOf()` followed by `add()` in timezone plugin returns incorrect dates. 

#### Solution

- *modified:* `startOf` method in `plugin/timezone` to properly preserve timestamps
- *added:* test cases for this edge case

### DIFF
--- a/src/plugin/timezone/index.js
+++ b/src/plugin/timezone/index.js
@@ -127,9 +127,10 @@ export default (o, c, d) => {
       return oldStartOf.call(this, units, startOf)
     }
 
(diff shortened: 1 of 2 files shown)


sdg9670f:
Previously, even if the time zone was set to 'tz', did 'startOf' work in UTC time?

literaryverses:
> Previously, even if the time zone was set to 'tz', did 'startOf' work in UTC time?

No it did not

PR #2896, opened 2025-06-25
Fix S and SS format/parsing

alxhotel:
The `S` and `SS` are broken. This PR fixes it.


alxhotel:
Please @iamkun take a look

PR #2898, opened 2025-07-05
fix: dont instantiate regexes within ar locale functions to avoid performance overhead

VehpuS:
Background for this issue: https://stackoverflow.com/questions/9750338/dynamic-vs-inline-regexp-performance-in-javascript

### DIFF
--- a/.github/workflows/check.yml
+++ b/.github/workflows/check.yml
@@ -1,7 +1,7 @@
 name: Lint & Unit Test
 on:
   push:
-    branches: [master, dev]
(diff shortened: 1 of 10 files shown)


VehpuS:
codecov issue seems like it's a problem with the codecov package - hopefully will be fixed soon so it can be rerun...

VehpuS:
Could there be a "SECRET" issue? https://github.com/typestack/class-validator/issues/2544

iamkun:
Thanks

PR #2900, opened 2025-07-13
feat: add Brazilian Portuguese (pt-br) locale

1750yi:
## What's changed?
- Added Brazilian Portuguese locale (`pt-br.js`)
- Includes translations for weekdays, months, etc.
- Added basic test cases.


iamkun:
only an "Empty file"

PR #2930, opened 2025-09-06
fix: replace italian "un' ora fa" with "un'ora fa", add tests for it locale

### LINKED ISSUE #2918: Fix for Italian locale relative time translation
The phrase 'One hour ago' is currently translated as 'un’ ora fa', but the correct translation is 'un’ora fa' (without the space after the apostrophe)

fcarne:
Fixes #2918

I also added tests for the it locale. However, while adding them, I noticed that "a few seconds" differs between Moment.js and Day.js:
- Moment.js → "alcuni secondi"
- Day.js → "qualche secondo"

Since both are correct but not identical, I had to split that case out from the other relative time tests.

### DIFF
--- a/src/locale/it.js
+++ b/src/locale/it.js
@@ -23,7 +23,7 @@ const locale = {
     s: 'qualche secondo',
     m: 'un minuto',
     mm: '%d minuti',
-    h: 'un\' ora',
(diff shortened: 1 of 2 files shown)


iamkun:
Thanks

PR #2945, opened 2025-10-22
fix(diff): allow .diff(unit) shorthand for diffing with current time

### LINKED ISSUE #2943: .diff('days') should be allowed
The `diff` function's documentation mention an example with no date being supplied allowing you to 'diff' with the current time, which implies to me that something like `.diff('days')` or some other unit should also work, but it seems that actually it tries to parse the unit string passed in as a date, gets an invalid date, and then returns NaN.  Since 'days' _is_ a string, this mistake also doesn't produce any type errors.

Ayo1984:
### Summary
This PR adds shorthand support for calling `.diff(unit)` as equivalent to `.diff(dayjs(), unit)`.

### Problem
The documentation suggests that if no date is supplied, `.diff()` compares with the current time.
However, calling `.diff('days')` or `.diff('hours')` currently treats the string as a date, resulting in `Invalid Date` → `NaN`.

### Solution
- Detect when `.diff()` is called with a single recognized unit string (e.g. 'days').
- Fully backward compatible — date-like strings continue to be parsed as dates.
- Added test cases.

### Example
```js
// Before (returns NaN)
dayjs('2025-01-10').diff('days')

// After
dayjs('2025-01-10').diff('days') // same as dayjs('2025-01-10').diff(dayjs(), 'days')


----
###
```
Resolves #2943

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -348,7 +348,13 @@ class Dayjs {
   }
 
   diff(input, units, float) {
-    const unit = Utils.p(units)
+    let unit
+    if (typeof input === 'string' && !units) {
+      unit = Utils.p(input)
+      input = dayjs()
+    } else {
+      unit = Utils.p(units)
+    }
     const that = dayjs(input)
     const zoneDelta = (that.utcOffset() - this.utcOffset()) * C.MILLISECONDS_A_MINUTE
     const diff = this - that
--- a/test/display.test.js
+++ b/test/display.test.js
@@ -1,8 +1,8 @@
-import moment from 'moment'
 import MockDate from 'mockdate'
+import moment from 'moment'
 import dayjs from '../src'
-import th from '../src/locale/th'
 import '../src/locale/ja'
+import th from '../src/locale/th'
 
 beforeEach(() => {
   MockDate.set(new Date())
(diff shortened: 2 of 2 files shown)


dominik-deak:
👌

Ayo1984:
@iamkun this is pretty small

iamkun:
Thanks for the contribution! Really appreciate the effort to make the API feel more intuitive.

After checking the behavior in **Moment.js**, it turns out that Moment **does not** support this shorthand syntax either — you always need to provide a comparison date as the first argument to `.diff()`. For example:

```js
moment('2025-01-10').diff('days') // invalid → tries to parse "days" as a date
moment('2025-01-10').diff(moment(), 'days') // ✅ valid usage
```

So in order to stay consistent with Moment.js (and to avoid unexpected implicit behavior), maybe, Day.js has also required a comparison date to be passed explicitly.

Could you share which specific part of that documentation gave you this impression?

 https://day.js.org/docs/en/display/difference

Ayo1984:
Thank you for the feedback, didn't realise MomentJS was still being used as a benchmark

I think the lack of compile time error must have made the issue creator confused, we can mark this PR as closed

iamkun:
You can actually see that most of the Day.js unit tests use Moment.js results as a reference — a lot of community members also adopt Day.js as a drop-in replacement for Moment.js.
So keeping the core API behavior aligned between the two libraries makes sense and is probably the best long-term direction for consistency and developer expectations.

At the same time, I think we could improve the developer experience by adding an error or warning for this specific case in the https://day.js.org/docs/en/plugin/dev-helper dev helper plugin.
That would help developers quickly spot incorrect usages like `.diff('days')` and understand why it’s invalid.

What do you think about that approach?

Ayo1984:
I see
And yeah, updating the dev helper plugin will be the best choice of action here
I'll create a PR for it if it's not already in your list of plans

iamkun:
Really excited for your PR! Appreciate you taking this on. 🙌

Ayo1984:
Just created a PR for that in #2948 , happy to close this afterwards

Ayo1984:
closed as addressed, solution available in #2948

PR #2948, opened 2025-10-26
feat(dev-helper): add warning for invalid .diff() usage without comparison date

Ayo1984:
This PR enhances the dev-helper plugin by adding a developer warning when .diff() is called without a valid comparison date.

Added unit tests to cover:
Missing comparison date
Invalid (unparsable) string
Valid date strings and Day.js instances (no warning)

Thanks for the earlier feedback in #2945 — this implementation follows the agreed approach to keep behavior consistent with Moment.js while improving developer guidance through dev-helper.


iamkun:
Looks cool. THX

PR #2952, opened 2025-10-31
Fix 12 PM meridiem displaying as AM instead of PM in ar-sa locale

### LINKED ISSUE #2951: Wrong meridiem in Arabic locale when time is 12 PM
Hi, I noticed a bug when using the Arabic locale (ar-sa) in Day.js

When the time is 12 PM Day.js returns the meridiem as "ص" (AM) instead of "م" (PM)

This only happens at 12 PM other PM hours like 1 PM, 2 PM, etc.. work fine

I also tested it using Vitest and the test fails for the same reason

So it seems like the Arabic locale logic doesn’t handle 12 PM correctly

```

import { describe, expect, test } from "vitest";
import dayjs from "dayjs";
import "dayjs/locale/ar-sa";

describe("Dayjs Arabic & English", () => {
  test("12:41 PM should be م in Arabic", () => {
    const result = dayjs("2025-10-28T09:41:10.000000Z").locale("ar-sa").format("hh:mm a");
    expect(result).toBe("12:41 م");
  });

  test("12:41 PM should be pm in English", () => {
    const result = dayjs("2025-10-28T09:41:10.000000Z").format("hh:mm a");
    expect(result).toBe("12:41 pm");
  });
});
```

```

AssertionError: expected '12:41 ص' to be '12:41 م' // Object.is equality

Expected: "12:41 م"
Received: "12:41 ص"
```

AnslemHack:
Fixes #2951  meridiem display bug in `ar-sa` locale where 12 PM (noon) incorrectly showed as "ص" (AM) instead of "م" (PM).

Changed meridiem condition from `hour > 12` to `hour >= 12` to correctly handle hour 12 in 12-hour format.

### DIFF
--- a/src/locale/ar-dz.js
+++ b/src/locale/ar-dz.js
@@ -17,7 +17,7 @@ const locale = {
     LLL: 'D MMMM YYYY HH:mm',
     LLLL: 'dddd D MMMM YYYY HH:mm'
   },
-  meridiem: hour => (hour > 12 ? 'م' : 'ص'),
+  meridiem: hour => (hour >= 12 ? 'م' : 'ص'),
   relativeTime: {
     future: 'في %s',
     past: 'منذ %s',
--- a/src/locale/ar-iq.js
+++ b/src/locale/ar-iq.js
@@ -18,7 +18,7 @@ const locale = {
     LLL: 'D MMMM YYYY HH:mm',
     LLLL: 'dddd D MMMM YYYY HH:mm'
   },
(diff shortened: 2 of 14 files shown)


AnslemHack:
@iamkun care looking at this fix? it addresses the wrong meridiem in arabic when time is 12pm

a7ridi:
Hey @iamkun because of this issue we're stuck and re-using `moment`. Can you please check at your earliest convenience?

Issue #2957, opened 2025-11-05
Inconsistent results on 25 hour day after time change


bluepnume:
**Describe the bug**

```javascript
const time = 1_762_066_800_000;
const timezone = 'US/Pacific';

const a = dayjs.tz(time, timezone)
    .add(1, 'day');

const b = dayjs.tz(
    dayjs.tz(time, timezone).add(1, 'day').valueOf(),
    timezone
);

const c = dayjs.tz(
    dayjs.tz(time, timezone).add(1, 'day').startOf('millisecond').valueOf(),
    timezone
);

console.log('a', a.valueOf(), a.format('YYYY-MM-DD HH:mm:ss'));
console.log('b', b.valueOf(), b.format('YYYY-MM-DD HH:mm:ss'));
console.log('c', c.valueOf(), c.format('YYYY-MM-DD HH:mm:ss'));
```

Output:

```javascript
a 1762153200000 2025-11-03 00:00:00
b 1762153200000 2025-11-02 23:00:00
c 1762156800000 2025-11-03 00:00:00
```

**Expected behavior**

I expect all 3 results to be:

```javascript
a 1762156800000 2025-11-03 00:00:00
b 1762156800000 2025-11-03 00:00:00
c 1762156800000 2025-11-03 00:00:00
```

**Information**
 - Day.js Version 1.11.19
 - OS: MacOS 15.5
 - Browser Node 22
 - Time zone: US/Pacific

PR #2961, opened 2025-11-07
fix(timezone): preserve timezone context when adding calendar units

SHARMA1525:
Fix issue where adding days/weeks/months/years to timezone-aware dayjs objects did not preserve timezone context correctly, causing incorrect valueOf() results when converting back to timezone.

The fix overrides the add() method for timezone-aware objects to:
1. Format the date in timezone context
2. Add calendar units without timezone conversion
3. Convert back to timezone with keepLocalTime=true

This ensures all three approaches produce identical results:
- dayjs.tz(time, tz).add(1, 'day')
- dayjs.tz(dayjs.tz(time, tz).add(1, 'day').valueOf(), tz)
- dayjs.tz(dayjs.tz(time, tz).add(1, 'day').startOf('millisecond').valueOf(), tz)

Fixes issue where valueOf() and format() results were inconsistent after adding calendar units to timezone-aware objects.

Linked Issue 
#2957

### DIFF
--- a/src/plugin/timezone/index.js
+++ b/src/plugin/timezone/index.js
@@ -132,6 +132,31 @@ export default (o, c, d) => {
     return startOfWithoutTz.tz(this.$x.$timezone, true)
   }
 
+  const oldAdd = proto.add
+  proto.add = function (number, units) {
+    if (!this.$x || !this.$x.$timezone) {
+      return oldAdd.call(this, number, units)
+    }
+
+    const unit = this.$utils().p(units)
+    // For calendar units (day, week, month, year), work in timezone context
+    // For time units (ms, second, minute, hour), use timestamp-based approach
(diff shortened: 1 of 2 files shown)


Xiphoseer:
`add` is not the only function where this goes wrong and it's not the only DST bug as you can see from the issue tracker. `set` is another candidate. What does sort of work is to patch the private `init` function to recompute the offset. I've built a custom `dst` plugin that does that but not sure I can share it.

SHARMA1525:
I’ve raised a new PR that implements the suggested fix by patching the private init() function to recompute the timezone offset.

This ensures DST offsets remain consistent across all methods (.add(), .set(), .startOf(), etc.), not just .add().

All test suites (93/93) pass successfully.

Please take a look when you get a chance  thank you again for your guidance!

Issue #2970, opened 2025-11-12
[Feature request] Disable rounding in `dayjs.duration().humanize()`


KaKi87:
Hi,

It would be nice to have an option for `{ seconds: 10 }` to output `10 seconds` rather than `a few seconds`, and more importantly, for `{ minutes: 45 }` to output `45 minutes` rather than `an hour`.

Thanks

KirtiGautam620:
Hi! I’d like to work on this feature request to add an option to disable rounding in dayjs.duration().humanize(). Could you please assign this issue to me?

bhavyajain-glitch:
Hello @KaKi87  and maintainers,
I'm a new contributor working on a university project, and I would love to implement this feature.

After reviewing the code, I've developed a solution and wanted to propose it here. My approach is to:
1.Update duration.humanize() to accept an options object (e.g., humanize({ round: false })).
2.If round: false is passed, the function bypasses the relativeTime's fromNow() logic and manually formats a precise string (like "45 minutes").
3.If no options are passed (or round is not false), it uses the original fromNow() logic to ensure 100% backward compatibility.
4.To support this, I've also added an ss: '%d seconds' string to the relativeTime locale, as requested in the issue.

I've already implemented this solution on my fork, and it passes all npm run lint and npm test checks, including the 100% test coverage requirement.
May I please be assigned to this issue? If so, I am ready to open a Pull Request for your review immediately.

Thank you!

alex-eri:
Also now `{days:1 , hours:12}` -> `2 days`, 

i want `1 day 12 hours`  not `36 hours`

Round depth as parameter will be cool

`{days:1 , hours:12, seconds:30}`  i want `about 1 day 12 hours`
`{days:1 , minutes :29}`  i want `more then day` (depth=2) and `about day` (depth=1) 

On rounded values is good to have prefixes: `about` ( rounded part is not next to rounded ), `more then`, `less then`

PR #2973, opened 2025-11-17
feat(plugin:duration): Add { round: false } option to humanize()

bhavyajain-glitch:
Hi maintainers!

This PR adds a round: false option to duration.humanize() to allow for unrounded, precise output.

Closes: #2970

Summary
Updated duration.humanize() to accept an options object for { round: false } and { withSuffix: true }.

If round: false is passed, it now bypasses the relativeTime thresholds and formats the string manually.

If no options are passed, it defaults to the original, rounding behavior (fully backward-compatible).

Added ss: '%d seconds' to the relativeTime plugin for the new "10 seconds" output.

Added full tests for the new feature, backward compatibility, and fallbacks, maintaining 100% test coverage.

I'm a student working on a university project to learn open source, and I'd be happy to make any changes you suggest. Thanks!

### DIFF
--- a/src/plugin/duration/index.js
+++ b/src/plugin/duration/index.js
@@ -23,6 +23,16 @@ const unitToMS = {
   weeks: MILLISECONDS_A_WEEK
 }
 
+const RELATIVE_TIME_UNITS = ['years', 'months', 'days', 'hours', 'minutes', 'seconds']
+const RELATIVE_TIME_UNIT_MAP = {
+  years: 'y',
+  months: 'M',
+  days: 'd',
+  hours: 'h',
+  minutes: 'm',
+  seconds: 's'
+}
+
 const isDuration = d => d instanceof Duration // eslint-disable-line no-use-before-define
 
 let $d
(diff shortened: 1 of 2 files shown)


test/plugin/duration.test.js
```diff
@@ -309,3 +309,30 @@ describe('Format', () => {
...
   })
 })
+
+it('humanize() with round: false option', () => {
+  expect(dayjs.duration({ minutes: 45 }).humanize({ round: false }))
+    .toBe('45 minutes')
+  expect(dayjs.duration({ seconds: 10 }).humanize({ round: false }))
+    .toBe('10 seconds')
+  expect(dayjs.duration({ seconds: 1 }).humanize({ round: false }))
+    .toBe('a few seconds')
```
michaelhazan: i feel like if we say 
```ts
dayjs.duration({ seconds: 1 }).humanize({ round: false })
``` 
it should be `a second`, to be consistent with 
```ts
dayjs.duration({ minutes: 1 }).humanize({ round: false })
```
being `a minute`


same applies to `withSuffix: true`

src/plugin/duration/index.js
```diff
@@ -229,11 +239,58 @@ class Duration {
     return wrapper(this.$ms, this)
   }
 
-  humanize(withSuffix) {
-    return $d()
-      .add(this.$ms, 'ms')
-      .locale(this.$l)
-      .fromNow(!withSuffix)
+  humanize(options) {
+    let withSuffix = false
```
michaelhazan: i feel like `withSuffix` should rather be `interval`, referring to it being a time interval.

after all this isn't necessarily a `suffix`, it changes the entire sentence, not only the end of it.

michaelhazan (changes requested):
i'm not a maintainer but i do feel these changes should be applied...

PR #3012, opened 2026-02-20
fix(updateLocale): merge nested objects for partial locale updates

### LINKED ISSUE #1118: UpdateLocale Plugin does not allow partial updates to locale

veeceey:
## Summary
- Fixes the `updateLocale` plugin to shallow-merge nested object properties (like `formats`, `relativeTime`) instead of replacing the entire object when only a subset of keys is provided
- Arrays and non-object values continue to be replaced entirely, preserving existing behavior
- Adds tests for partial nested object updates and array replacement

Fixes #1118

## Test plan
- [x] Existing `updateLocale` tests continue to pass
- [x] New test: partial update to `formats` preserves unmodified keys (e.g., updating only `L` keeps `LT`, `LL`, etc.)
- [x] New test: array values (like `months`) are still replaced entirely, not merged
- [x] 100% code coverage on the updateLocale plugin

### DIFF
--- a/src/plugin/updateLocale/index.js
+++ b/src/plugin/updateLocale/index.js
@@ -5,7 +5,15 @@ export default (option, Dayjs, dayjs) => {
     if (!localeConfig) return
     const customConfigKeys = customConfig ? Object.keys(customConfig) : []
     customConfigKeys.forEach((c) => {
-      localeConfig[c] = customConfig[c]
+      if (localeConfig[c] && customConfig[c] && typeof localeConfig[c] === 'object' && typeof customConfig[c] === 'object'
+        && !Array.isArray(localeConfig[c])) {
+        localeConfig[c] = {
+          ...localeConfig[c],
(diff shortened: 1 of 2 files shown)


veeceey:
hey, still around if any changes are needed here — just lmk!

iamkun:
Thanks.

PR #3016, opened 2026-03-12
fix: preserve unsupported year tokens in format (#3015)

### LINKED ISSUE #3015: Bug: undocumented Y and YYY tokens fall through to ZZ formatting
### Summary

`dayjs().format()` currently treats `Y` and `YYY` as matched tokens, but does not format them as year tokens.

Instead, they fall through to the timezone-offset fallback used for `ZZ`, producing output such as `+0800`.

### Expected Behavior

The formatting documentation does not mention `Y` or `YYY` as supported `format()` tokens.

Because of that, `Y` and `YYY` should not be formatted using timezone-offset logic.

A more reasonable behavior would be to preserve them literally, for example:

```js
dayjs('2024-01-02').format('Y')   // 'Y'
dayjs('2024-01-02').format('YYY')  // 'YYY'
```

At minimum, they should not behave like `ZZ`.

### Actual Behavior

```js
dayjs('2024-01-02').format('Y')   // return: +0800
dayjs('2024-01-02').format('YYY')// return: +0800
```

The exact output depends on the local timezone, because these tokens fall through to timezone-offset formatting.

### Minimal Reproduction

```js
dayjs('2024-01-02').format('Y')
dayjs('2024-01-02').format('YYY')
```

### Root Cause

- `REGEX_FORMAT` matches `Y{1,4}`
- `format()` only implements `YY` and `YYYY`
- unmatched tokens fall through to:

```js
matches(match) || zoneStr.replace(':', '')
```

That fallback is effectively the implementation used for `ZZ`, so `Y` and `YYY` are formatted as timezone offsets.

### Why This Is a Bug

- `Y` and `YYY` are not documented `format()` tokens
- users would not expect undocumented year-like tokens to behave like timezone formatting
- the current output is an internal fallback artifact, not a meaningful formatting result

Ryokki:
## Summary

This changes `format()` so unsupported year tokens are preserved literally instead of falling through to the timezone-offset fallback.

Closes #3015.

## Root Cause

`REGEX_FORMAT` matches `Y{1,4}`, but `format()` only implements `YY` and `YYYY`.

Previously, unmatched tokens fell through to:

```js
matches(match) || zoneStr.replace(':', '')
```

That fallback was intended for `ZZ`, but it also affected unsupported tokens such as `Y` and `YYY`, causing them to be formatted as timezone offsets.

## Changes

- handle `ZZ` explicitly in `format()`
- preserve unsupported matched tokens literally
- add regression coverage in `display.test.js` for:
  - `Y`
  - `YYY`

### DIFF
--- a/src/constant.js
+++ b/src/constant.js
@@ -27,4 +27,4 @@ export const INVALID_DATE_STRING = 'Invalid Date'
 
 // regex
 export const REGEX_PARSE = /^(\d{4})[-/]?(\d{1,2})?[-/]?(\d{0,2})[Tt\s]*(\d{1,2})?:?(\d{1,2})?:?(\d{1,2})?[.:]?(\d+)?$/
-export const REGEX_FORMAT = /\[([^\]]+)]|Y{1,4}|M{1,4}|D{1,2}|d{1,4}|H{1,2}|h{1,2}|a|A|m{1,2}|s{1,2}|Z{1,2}|SSS/g
+export const REGEX_FORMAT = /\[([^\]]+)]|YYYY|YY|M{1,4}|D{1,2}|d{1,4}|H{1,2}|h{1,2}|a|A|m{1,2}|s{1,2}|Z{1,2}|SSS/g
--- a/src/index.js
+++ b/src/index.js
@@ -332,14 +332,13 @@ class Dayjs {
           return Utils.s(this.$ms, 3, '0')
         case 'Z':
           return zoneStr // 'ZZ' logic below
-        case 'ZZ':
-          return zoneStr.replace(':', '')
         default:
-          return match
+          break
       }
+      return null
     }
 
-    return str.replace(C.REGEX_FORMAT, (match, $1) => $1 || matches(match))
+    return str.replace(C.REGEX_FORMAT, (match, $1) => $1 || matches(match) || zoneStr.replace(':', '')) // 'ZZ'
   }
 
   utcOffset() {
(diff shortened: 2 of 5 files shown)


iamkun:
Perhaps it would be better to update the REGEX_FORMAT regex constant to YYYY|YY to address this issue at the source.
A copy of the current REGEX_FORMAT could be preserved in src/plugin/duration/index.js.
This would fix the problem at its root and keep the implementation simpler and cleaner.
The current test cases could remain as they are.

Ryokki:
Thanks, I agree that tightening `REGEX_FORMAT` is the better fix here, since it resolves the mismatch at the source instead of handling it later in `format()`.

I also agree that `src/plugin/duration/index.js` may need to keep its own copy of the current regex, because its supported year tokens differ from core `format()`.

One point I want to clarify is the expected behavior for `YYY` once the core regex is narrowed to `YYYY|YY`.

My current preference is to treat `YYY` as `YY` token plus a literal `Y`, rather than preserving the whole sequence as literal `YYY`. That seems like the simpler rule: supported tokens are matched normally, and only the unsupported remainder falls back to literal text.

If you agree with that interpretation, I can adjust the implementation and tests in that direction.

iamkun:
Thanks for the clarification. I agree that interpreting YYY as YY + a literal Y makes sense. It keeps the rule simple: supported tokens are parsed normally, and the remaining characters fall back to literal text. That behavior looks reasonable to me.

Ryokki:
Thanks for confirming. I’ve updated the implementation accordingly.

iamkun:
THX

PR #3023, opened 2026-03-15
fix(customParseFormat): preserve UTC mode when format is an array

### LINKED ISSUE #3013: Passing array value to format param to dayjs.utc() silently disables utc mode
**Describe the bug**
When an array is passed as the format argument to dayjs.utc(), UTC mode is not used: the date is parsed in the local timezone.
dayjs() accepts an array for the format parameter (with the customParseFormat plugin), so it’s easy to assume dayjs.utc() does too. With an array, dayjs.utc() does not throw; it silently parses in local time, which can cause subtle bugs (e.g. tests passing only when TZ=UTC).

To reproduce
```

dayjs.extend(utc);

// With a single format, result is UTC midnight (correct)
const single = dayjs.utc('1995-05-01', 'YYYY-MM-DD');
console.log(single.toISOString()); // '1995-05-01T00:00:00.000Z'

// With a format array, result is local midnight (incorrect)
const withArray = dayjs.utc('1995-05-01', ['YYYY-MM-DD', 'YYYY-M-D']);
console.log(withArray.toISOString()); // e.g. '1995-05-01T23:00:00.000Z' in UTC+1
```
Stackblitz link https://stackblitz.com/edit/dayjs-playground-8cua8kja?file=index.js

**Expected behavior**
Either:
- Support format arrays and parse in UTC (same semantics as single-format dayjs.utc(date, format)), or
- Reject array format (e.g. throw error)

**Likely cause**
 In `customParseFormat`, when format is an array, the code uses `d.apply(this, args)` (the default constructor) instead of the UTC constructor, so each format is tried in local time.

**Information**
 - Day.js Version 1.11.13
 - OS: macos
 - Browser: NA
 - Time zone: NA

Jadu07:
## What this PR does

Fixes #3013 — `dayjs.utc()` ignores UTC mode when the format argument is an array.

---

### The bug

If you pass an array of formats to `dayjs.utc()` (via the `customParseFormat` plugin), the date gets parsed in **local time** instead of UTC. No error is thrown, which makes the bug really hard to spot — your tests might pass fine in `TZ=UTC` but silently break everywhere else.

```js
dayjs.extend(utc)
dayjs.extend(customParseFormat)

// single format — works fine
dayjs.utc('1995-05-01', 'YYYY-MM-DD').toISOString()
// → '1995-05-01T00:00:00.000Z' ✓

// array format — parsed in local time before this fix
dayjs.utc('1995-05-01', ['YYYY-MM-DD', 'YYYY-M-D']).toISOString()
// → '1995-05-01T23:00:00.000Z' in UTC+1 ✗
```

---

### Why it happened

In `customParseFormat/index.js`, the `format instanceof Array` branch loops through each format and tries it with `d.apply(this, args)`. The problem is that `d.apply` just calls the regular `dayjs()` constructor, which has no idea about `utc: true`. So every format attempt ends up using local time.

Meanwhile, the single-format code path calls `parseFormattedInput(date, format, utc, d)`, which correctly threads the `utc` flag through.

---

### The fix

One line change — once we find a valid format, compute `this.$d` using `parseFormattedInput` with the `utc` flag instead of blindly copying `result.$d`:

```diff
  if (result.isValid()) {
-   this.$d = result.$d
+   this.$d = utc
+     ? parseFormattedInput(date, format[i - 1], utc, d)
+     : result.$d
```

---

### Tests

Added two regression tests to `test/plugin/utc.test.js`:

- first format in the array matches → result is UTC midnight ✓
- second format in the array matches → result is UTC midnight ✓

Full suite: **774 tests, 93 suites — all passing**. `npm run lint` is clean too.

### DIFF
--- a/src/plugin/customParseFormat/index.js
+++ b/src/plugin/customParseFormat/index.js
@@ -256,7 +256,9 @@ export default (o, C, d) => {
         args[1] = format[i - 1]
         const result = d.apply(this, args)
         if (result.isValid()) {
-          this.$d = result.$d
+          this.$d = utc
+            ? parseFormattedInput(date, format[i - 1], utc, d)
+            : result.$d
           this.$L = result.$L
           this.init()
           break
--- a/test/plugin/utc.test.js
+++ b/test/plugin/utc.test.js
@@ -219,6 +219,26 @@ describe('UTC with customParseFormat', () => {
     expect(instant.format()).toBe('2011-02-02T03:04:05Z')
     expect(instant.format()).toBe(momentInstant.format())
   })
+
+  it('Array format - parses in UTC, not local time', () => {
+    dayjs.extend(customParseFormat)
+    const date = '1995-05-01'
+    const formats = ['YYYY-MM-DD', 'YYYY-M-D']
+    const result = dayjs.utc(date, formats)
+    expect(result.isValid()).toBe(true)
+    expect(result.toISOString()).toBe('1995-05-01T00:00:00.000Z')
+    expect(result.isUTC()).toBe(true)
+  })
+
+  it('Array format - second format matches, parsed in UTC', () => {
+    dayjs.extend(customParseFormat)
+    const date = '1995-5-1'
+    const formats = ['YYYY-MM-DD', 'YYYY-M-D']
+    const result = dayjs.utc(date, formats)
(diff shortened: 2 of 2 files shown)


Jadu07:
@iamkun @pavan-sh pls review this pr.

PR #3024, opened 2026-03-18
fix: handle invalid duration humanize

### LINKED ISSUE #3006: duration.humanize(): returns "a month" for `NaN`
**Describe the bug**

```console
> dayjs.duration(NaN).humanize()
"a month"
```

**Expected behavior**

```console
> dayjs.duration(NaN).humanize()
"invalid duration"
```

**Information**
 - Day.js Version: 1.11.19
 - OS: macOS 26.3
 - Runtime: deno 2.6.9
 - Time zone: GMT-06:00 CST (Central Standard Time)

DeoJin:
## Summary
Return \invalid duration\ from \duration.humanize()\ when the duration value is not finite, instead of letting relative-time formatting turn \NaN\ into an incorrect humanized string.

## Changes
- guard \Duration#humanize\ against non-finite millisecond values
- add regression coverage for \dayjs.duration(NaN).humanize()\ with and without suffixes

## Testing
- \
px jest test/plugin/duration.test.js --runInBand\

Fixes #3006

### DIFF
--- a/src/plugin/duration/index.js
+++ b/src/plugin/duration/index.js
@@ -230,6 +230,10 @@ class Duration {
   }
 
   humanize(withSuffix) {
+    if (!Number.isFinite(this.$ms)) {
+      return 'invalid duration'
+    }
+
     return $d()
       .add(this.$ms, 'ms')
       .locale(this.$l)
--- a/test/plugin/duration.test.js
+++ b/test/plugin/duration.test.js
@@ -147,6 +147,11 @@ describe('Humanize', () => {
     expect(dayjs.duration(1, 'minutes').humanize(true)).toBe('en un minuto')
     dayjs.locale('en')
   })
+
+  it('Invalid duration', () => {
+    expect(dayjs.duration(NaN).humanize()).toBe('invalid duration')
+    expect(dayjs.duration(NaN).humanize(true)).toBe('invalid duration')
+  })
 })
 
 describe('Clone', () => {


yfwmaniish:
Root-cause note for reviewers, in case it's useful: the reason this specifically surfaces as `"a month"` (not some other nonsense string) is `monthDiff` in `src/utils.js`:

```js
const monthDiff = (a, b) => {
  const c = b - anchor < 0
    (anchor2 - anchor)))) || 0)
}
```

Once `$ms` is `NaN`, `humanize()`'s `$d().add(this.$ms, 'ms')` produces a Dayjs wrapping an Invalid Date. The `relativeTime` threshold loop's `diff()` calls for the `y`/`yy`/`M`/`MM` buckets route through `monthDiff`, whose final `|| 0` silently turns the resulting `NaN` into `0` (rather than propagating it) -- so the loop reads "0 months", the `abs <= 1 && i > 0` demotion kicks in, and it prints the singular `'M'` bucket's string, `"a month"`, instead of ever recognizing the instant as invalid.

Worth knowing this `|| 0` fallback exists independently of `duration.humanize()` -- it'd produce the same silent-zero behavior for *any* `.diff(..., 'year'|'month')` call where either side is an Invalid Date, not just this path. Out of scope for this PR (which correctly fixes the reported symptom at the `humanize()` boundary), but potentially worth a follow-up issue if others hit it via a different route.

PR #3027, opened 2026-03-21
fix(test): make dayOfYear and localizedFormat tests timezone-independent

### LINKED ISSUE #3009: dayOfYear/localizedFormat of Plugin Tests Fail in Non-UTC Timezones Due to Missing Parsed Timezone Info
# Bug Report: Timezone Handling in Day.js Plugins

**Describe the bug**  
When the local computer timezone is set to US/Los_Angeles, most plugin test cases fail. After the latest fixes, almost all test cases pass except for two:  
- `test/plugin/localizedFormat.test.js`  
- `test/plugin/dayOfYear.test.js`  

These two are hard to fix because Day.js does not store the timezone information when parsing a string. Therefore, functions like `dayOfYear` and `format` can only operate based on the current environment's timezone. Fixing this properly would require a core feature change in Day.js to store the original string’s timezone, which would be a significant modification.  

**Expected behavior**  
All plugin test cases should pass regardless of the local system timezone, correctly handling `dayOfYear` and `format` when the input string contains a timezone offset.  

**Information**  
- Day.js Version: ALL  
- OS: macOS / Windows (any OS, but tested on US/LA timezone)  
- Browser: N/A (Node.js environment)  
- Time zone: US/Pacific (GMT-07:00 DST, Pacific Daylight Time)  

**Test cases demonstrating the issue**  
```js
describe('Timezone conversion format', () => {
  it('converts between fixed-offset timezones correctly', () => {
    expect(dayjs.tz('2025-03-09 12:00', 'Etc/GMT-5').tz('Etc/GMT+5').format())
      .toBe('2025-03-09T02:00:00-05:00');
    expect(dayjs.tz('2025-03-09 12:30', 'Etc/GMT-5').tz('Etc/GMT+5').format())
      .toBe('2025-03-09T02:30:00-05:00');
  });

  it('converts from fixed-offset timezone to IANA timezone correctly', () => {
    expect(dayjs('2025-03-08T21:00:00.000Z')
      .tz('Etc/GMT-5')
      .tz('America/Argentina/Buenos_Aires')
      .format()).toBe('2025-03-08T18:00:00-03:00');
  });
});

<img width="2370" height="1846" alt="Image" src="https://github.com/user-attachments/assets/6d33e4f5-afaf-4422-916a-98dba169c767" />

mathieumaf:
### Summary

Fixes #3009

The `dayOfYear` and `localizedFormat` plugin tests fail when run in non-UTC timezones (e.g. `US/Pacific`). The issue is **not in the plugin code** but in the tests themselves, which implicitly assume `UTC` as the local timezone.

Two problems:

- **`dayOfYear.test.js`**: Uses UTC strings (`2015-01-01T00:00:00.000Z`) then asserts results with `toISOString()`. In UTC-8, this date resolves to Dec 31 2014 locally, so `dayOfYear(4)` operates on the wrong year.
- **`localizedFormat.test.js`**: Uses `new Date(0)` (epoch = Jan 1 1970 00:00 **UTC**), which in UTC-8 is Dec 31 1969. The test then asserts `YYYY` equals `1970`.

### Changes

- `dayOfYear.test.js`: Use local date strings (without `Z` suffix) and `format()` instead of `toISOString()`
- `localizedFormat.test.js`: Use `new Date(1970, 0, 1)` (local) instead of `new Date(0)` (UTC)

### Test plan

- [x] `npm run lint` passes
- [x] All tests pass in `TZ=UTC`
- [x] All tests pass in `TZ=US/Pacific`
- [x] All tests pass in `TZ=Asia/Kolkata`
- [x] All tests pass in `TZ=Pacific/Auckland`

### DIFF
--- a/test/plugin/dayOfYear.test.js
+++ b/test/plugin/dayOfYear.test.js
@@ -24,27 +24,27 @@ it('DayOfYear get', () => {
 
 it('DayOfYear set', () => {
   expect(dayjs().dayOfYear(4).dayOfYear()).toBe(moment().dayOfYear(4).dayOfYear())
-  expect(dayjs('2015-01-01T00:00:00.000Z')
+  expect(dayjs('2015-01-01T00:00:00.000')
     .dayOfYear(4)
     .dayOfYear()).toBe(4)
 
-  expect(dayjs('2015-01-01T00:00:00.000Z')
+  expect(dayjs('2015-01-01T00:00:00.000')
     .dayOfYear(4)
-    .toISOString()).toBe('2015-01-04T00:00:00.000Z')
+    .format('YYYY-MM-DDTHH:mm:ss.SSS')).toBe('2015-01-04T00:00:00.000')
 
-  expect(dayjs('2015-01-01T00:00:00.000Z')
+  expect(dayjs('2015-01-01T00:00:00.000')
     .dayOfYear(32)
     .dayOfYear()).toBe(32)
 
-  expect(dayjs('2015-01-01T00:00:00.000Z')
+  expect(dayjs('2015-01-01T00:00:00.000')
     .dayOfYear(32)
-    .toISOString()).toBe('2015-02-01T00:00:00.000Z')
+    .format('YYYY-MM-DDTHH:mm:ss.SSS')).toBe('2015-02-01T00:00:00.000')
 
-  expect(dayjs('2015-01-01T00:00:00.000Z')
+  expect(dayjs('2015-01-01T00:00:00.000')
     .dayOfYear(365)
     .dayOfYear()).toBe(365)
 
-  expect(dayjs('2015-01-01T00:00:00.000Z')
+  expect(dayjs('2015-01-01T00:00:00.000')
     .dayOfYear(365)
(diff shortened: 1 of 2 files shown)


mathieumaf:
Friendly ping on this one — still relevant and still applies cleanly.

Quick status check against `dev` (as of `73017be`):

- **Still needed** — #3009 is open, and neither `test/plugin/dayOfYear.test.js` nor `test/plugin/localizedFormat.test.js` has changed upstream since 2022, so the failures still reproduce on any non-UTC machine.
- **No conflicts** — the patch applies cleanly to current `dev`; nothing to rebase.
- **Zero risk to users** — test-only change, 10 lines added / 10 removed. No `src/` code is touched, so it cannot affect published behaviour.

The practical benefit is for contributors: right now `npm test` fails out of the box for anyone whose machine isn't set to UTC, which is a confusing first experience when you're trying to validate an unrelated change.

Verified passing under `TZ=UTC`, `TZ=US/Pacific`, `TZ=Asia/Kolkata` and `TZ=Pacific/Auckland`.

Happy to rebase or split it further if that helps — just let me know.

PR #3034, opened 2026-04-04
fix: resolve #894 — Feature: add locale format string to localeData or somewhere else

### LINKED ISSUE #894: Feature: add locale format string to localeData or somewhere else
I want to set a datefields default format based on the language preference which is set on start but the only way to get the format string for the correct locale is seen below. I would like to have a valid function or constant that I can use. The right place for that would be in localeData or in `locale()` IMO but I don't even undestand why there are two `locale` and `$locale`?

```js
dayjs().$locale().formats.L
```

Bildschirmfoto 2020-05-03 um 16 20 11

tang-vu:
## Summary

fix: resolve #894 — Feature: add locale format string to localeData or somewhere else

## Problem

**Severity**: `Low` | **File**: `src/plugin/localeData/index.js`

Currently, there is no public, documented way to retrieve the localized format strings (like 'L', 'LL', 'LT', etc.) for a specific locale without accessing internal properties like `$locale().formats`. This enhancement adds the `longDateFormat` method to the `localeData` plugin, providing a Moment.js-compatible API to retrieve these format strings. This is useful for UI components that need to know the date format pattern (e.g., 'MM/DD/YYYY') based on the user's current locale.

## Solution

Update the `localeData` plugin to include the `longDateFormat` method in the object returned by `dayjs().localeData()` and `dayjs.localeData()`.

## Changes

- `src/plugin/localeData/index.js` (modified)

## Testing

- [ ] Existing tests pass
- [ ] Manual review completed
- [ ] No new warnings/errors introduced

---
*Generated by [ContribAI](https://github.com/tang-vu/ContribAI) v5.7.1*

### DIFF
--- a/src/plugin/localeData/index.js
+++ b/src/plugin/localeData/index.js
@@ -13,8 +13,11 @@ export default (o, c, dayjs) => { // locale needed later
     return result.map((_, index) => (result[(index + (weekStart || 0)) % 7]))
   }
   const getDayjsLocaleObject = () => dayjs.Ls[dayjs.locale()]
-  const getLongDateFormat = (l, format) =>
-    l.formats[format] || t(l.formats[format.toUpperCase()])
+  const getLongDateFormat = (l, format) => {
+    const { formats } = l
+    const upper = format.toUpperCase()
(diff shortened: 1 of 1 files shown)


iamkun:
test case please

PR #3041, opened 2026-04-10
fix: ja locale date format

### LINKED ISSUE #3038: fix: ja locale date format
**Describe the bug**
The format "L" and "l" in "ja" are the same.
It is different from "zh".

**Expected behavior**
The month and day are single digits.
l: 'YYYY/M/D'

**Information**
 - Day.js Version [v1.11.20]

mata02:
A modification of format 'l'.

Closes #3038


iamkun:
maybe we can check this as a reference? https://github.com/moment/moment/blob/develop/src/locale/ja.js

mata02:
I'm closing the pull request because it would break compatibility with Moment.js.

PR #3052, opened 2026-05-07
fix: Update Switzerland locale(de-ch) yearStart 4

rahulpratik:
Fix issue https://github.com/iamkun/dayjs/issues/3049


Ayan-Amir:
Fix looks right. One suggestion: the test cases all cover week 1 and 2 boundaries, but none assert week 53, which is the symptom reported in #3049.

Might be worth adding:

  ['2020-12-31', 53, 'Dec 31 2020 is week 53'],

2020-12-31 was a Thursday, so under ISO, the week runs 28 Dec to 3 Jan, and that week's Thursday falls in 2020, making it week 53. It's the exact case from the issue, and it would fail without this change.

PR #3069, opened 2026-05-24
fix: fixed hour verbiage

rahulpratik:
Fix hour verbiage in Croatian.
Screenshot 2026-05-23 at 5 14 48 PM


iamkun:
seems we can not use a fixed string https://github.com/moment/moment/blob/develop/src/locale/hr.js

rahulpratik:
> seems we can not use a fixed string https://github.com/moment/moment/blob/develop/src/locale/hr.js

Thanks for catching, I fixed the formats by cross checking the translations.

PR #3138, opened 2026-06-16
fix: add meridiem to Turkish locale

FatahChan:
## Summary

- Adds `meridiem` to the Turkish (`tr`) locale so 12-hour formatting uses correct Turkish AM/PM markers (`ÖÖ` / `ÖS`).

## Motivation

The Turkish locale was missing a `meridiem` function, which caused incorrect or missing output when formatting times with the `A` token in 12-hour mode.

## Test plan

- [x] `npm run lint`
- [x] `npx jest test/locale/keys.test.js`

Made with [Cursor](https://cursor.com)

### DIFF
--- a/src/locale/tr.js
+++ b/src/locale/tr.js
@@ -17,6 +17,7 @@ const locale = {
     LLL: 'D MMMM YYYY HH:mm',
     LLLL: 'dddd, D MMMM YYYY HH:mm'
   },
+  meridiem: hour => (hour < 12 ? 'ÖÖ' : 'ÖS'),
   relativeTime: {
     future: '%s sonra',
     past: '%s önce',


nrps9909 (approved):
Verified on exact head `3a25939d9094946b42b9e93447980c90ce7ae4b1`.

The implementation correctly makes `A` format as `ÖÖ` before noon and `ÖS` from noon onward. I ran the existing locale-key suite (143/143), ESLint on `src/locale/tr.js`, and an independent focused regression covering 00:00, 11:59, 12:00, and 23:59; all passed.

This also addresses the remaining root cause behind ant-design/ant-design#48235: rc-picker already delegates its `cellMeridiemFormat: 'A'` to the active Day.js locale, but Day.js currently falls back to English because `tr` has no `meridiem` function.

Non-blocking suggestion: a committed `test/locale/tr.test.js` assertion across the noon boundary would protect the exact `ÖÖ`/`ÖS` behavior; `keys.test.js` only proves that the function returns a value.

Codex-assisted review: I traced the downstream rc-picker path, checked for duplicate Day.js work, and ran the exact-head and independent boundary verification above.

Issue #3123, opened 2026-06-09
dayjs.tz(ms, 'UTC').startOf('day') returns prior-day UTC midnight on hosts with DST


bluepnume:
**Describe the bug**

`dayjs.tz(ms, 'UTC').startOf('day')` returns the start of the previous day in UTC instead of the correct day when the host's local timezone has a DST transition on the date being processed.

The Dayjs returned by `dayjs.tz(ms, 'UTC')` is supposed to be pinned to UTC for chained operations, but `startOf('day')` appears to fall back to host-local-time semantics. Using the UTC plugin directly (`dayjs.utc(ms)`) produces the correct result.

**Reproducing test case**

Run with the host's `TZ` set to a zone that has a DST transition on 2024-11-03 (e.g. `TZ=America/Los_Angeles`):

```javascript
const dayjs = require('dayjs');
const utc = require('dayjs/plugin/utc');
const timezone = require('dayjs/plugin/timezone');
dayjs.extend(utc);
dayjs.extend(timezone);

// 2024-11-03 23:00 UTC. Local TZ America/Los_Angeles ends DST at
// 02:00 PDT on this date (PDT -> PST), so 2024-11-03 straddles a
// DST transition in local time.
const ms = new Date('2024-11-03T23:00:00Z').getTime();

const buggy   = dayjs.tz(ms, 'UTC').startOf('day').valueOf();
const correct = dayjs.utc(ms).startOf('day').valueOf();

const expected = new Date('2024-11-03T00:00:00Z').getTime();

console.log('buggy   ->', new Date(buggy).toISOString());
console.log('correct ->', new Date(correct).toISOString());
console.log('expected->', new Date(expected).toISOString());

// Expected output:
//   buggy   -> 2024-11-03T00:00:00.000Z
//   correct -> 2024-11-03T00:00:00.000Z
//   expected-> 2024-11-03T00:00:00.000Z
//
// Actual output (on a host with US DST on 2024-11-03):
//   buggy   -> 2024-11-02T23:00:00.000Z   <-- off by one hour
//   correct -> 2024-11-03T00:00:00.000Z
//   expected-> 2024-11-03T00:00:00.000Z
```

**Expected behavior**

`dayjs.tz(ms, 'UTC').startOf('day').valueOf()` should equal `dayjs.utc(ms).startOf('day').valueOf()` — i.e., the start of the UTC day containing `ms`. The choice of host-local timezone should not affect the result when the chained timezone is explicitly UTC.

[...]

PR #3144, opened 2026-06-17
fix in timezone: ensure UTC chain is not affected by host DST in startOf

EzequielArce:
**Description**
This PR fixes a bug in the `timezone` plugin where calling `.startOf('day')` on a UTC instance returns an incorrect date if the host machine's local time is undergoing a Daylight Saving Time (DST) transition on that specific date.

**The Problem**
In the original implementation, `proto.startOf` handles timezone shifting by formatting the date into a naive string (`YYYY-MM-DD HH:mm:ss:SSS`) and parsing it back using the default factory `d()`. 

When the target timezone is `UTC` but the host system (e.g., `America/Los_Angeles`) experiences a DST transition (such as a 25-hour day), the core's default parsing falls back to the host's local clock constraints. This causes the underlying local `Date` mathematics to miscalculate the midnight boundary, leaking the host's DST anomaly into the final UTC object.

**Solution presented**
When `this.$x.$timezone` is `'UTC'`:
I use `{ utc: true }` in the configuration.

**Type of Change**
- [x] Change in the plugin timezone.js
- [x] Added the following test: transitionDSTtoUTC

Addresses [#3123](https://github.com/iamkun/dayjs/issues/3123#issue-4618852197)

### DIFF
--- a/src/plugin/timezone/index.js
+++ b/src/plugin/timezone/index.js
@@ -126,9 +126,13 @@ export default (o, c, d) => {
     if (!this.$x || !this.$x.$timezone) {
       return oldStartOf.call(this, units, startOf)
     }
-
-    const withoutTz = d(this.format('YYYY-MM-DD HH:mm:ss:SSS'), { locale: this.$L })
-    const startOfWithoutTz = oldStartOf.call(withoutTz, units, startOf)
+    let ins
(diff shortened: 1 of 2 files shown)


idanlevitski:
🤩🤩

PR #3152, opened 2026-06-29
fix(locale): use correct English ordinal in en-ca, en-ie, en-il, en-sg

mahirhir:
### What

`en-ca`, `en-ie`, `en-il` and `en-sg` still carry the placeholder `ordinal: n => n`, so `format('Do')` returns the bare day number for them:

```js
dayjs('2021-01-21').locale('en-au').format('Do') // "21st"
dayjs('2021-01-21').locale('en-ca').format('Do') // "21"  (expected "21st")
```

(`Do` comes from the `advancedFormat` plugin.)

The other English locales already define the standard English ordinal function: `en`, `en-au`, `en-gb`, `en-in`, `en-nz`, `en-tt`. English ordinals (1st/2nd/3rd, 11th/21st) are the same across these regions, so the four locales above were simply missing the function rather than intentionally different. This is the same change #2878 made for `en-au` ("Use the same ordinal as moment"); moment also defines the full ordinal for `en-ca`, `en-ie`, `en-il` and `en-sg`.

### Fix

Reuse the existing English ordinal function in the four locales.

### Test

Added an ordinal parity check in `test/locale/en.test.js` that compares `format('Do')` against moment for days 1 to 31 in each of the four locales. It fails on the old `n => n` and passes with the change. The full `jest` suite is green and `eslint` passes.

### DIFF
--- a/src/locale/en-ca.js
+++ b/src/locale/en-ca.js
@@ -8,7 +8,11 @@ const locale = {
   weekdaysShort: 'Sun_Mon_Tue_Wed_Thu_Fri_Sat'.split('_'),
   monthsShort: 'Jan_Feb_Mar_Apr_May_Jun_Jul_Aug_Sep_Oct_Nov_Dec'.split('_'),
   weekdaysMin: 'Su_Mo_Tu_We_Th_Fr_Sa'.split('_'),
-  ordinal: n => n,
+  ordinal: (n) => {
+    const s = ['th', 'st', 'nd', 'rd']
+    const v = n % 100
+    return `[${n}${(s[(v - 20) % 10] || s[v] || s[0])}]`
+  },
   formats: {
     LT: 'h:mm A',
     LTS: 'h:mm:ss A',
--- a/src/locale/en-ie.js
+++ b/src/locale/en-ie.js
@@ -9,7 +9,11 @@ const locale = {
   weekdaysShort: 'Sun_Mon_Tue_Wed_Thu_Fri_Sat'.split('_'),
   monthsShort: 'Jan_Feb_Mar_Apr_May_Jun_Jul_Aug_Sep_Oct_Nov_Dec'.split('_'),
   weekdaysMin: 'Su_Mo_Tu_We_Th_Fr_Sa'.split('_'),
-  ordinal: n => n,
+  ordinal: (n) => {
+    const s = ['th', 'st', 'nd', 'rd']
+    const v = n % 100
+    return `[${n}${(s[(v - 20) % 10] || s[v] || s[0])}]`
+  },
   formats: {
     LT: 'HH:mm',
     LTS: 'HH:mm:ss',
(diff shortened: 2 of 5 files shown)


mahirhir:
No review on this since June. Here it is measured against the published package rather than described, plus the denominator, since "four locales" is only interesting next to how many there are.

Reading every English locale on `dev`:

```
locale   ordinal declaration   carries the th/st/nd/rd table
en       (n) => { ...          true
en-au    (n) => { ...          true
en-gb    (n) => { ...          true
en-in    (n) => { ...          true
en-nz    (n) => { ...          true
en-tt    (n) => { ...          true
en-ca    n => n,               false
en-ie    n => n,               false
en-il    n => n,               false
en-sg    n => n,               false

English locales = 10, implemented = 6, returning the number unchanged = 4
```

Six of ten already carry the same suffix table, including the base `en`. The four in this PR are the ones that were never filled in, and the identity function reads like a deliberate choice rather than an omission, which is probably why it lasted.

What a caller sees, running `dayjs@1.11.23` with `advancedFormat` and the `Do` token:

```
locale       1     2     3    11    21    22    23    31
en         1st   2nd   3rd  11th  21st  22nd  23rd  31st
en-gb      1st   2nd   3rd  11th  21st  22nd  23rd  31st
en-ca        1     2     3    11    21    22    23    31
en-ie        1     2     3    11    21    22    23    31
en-il        1     2     3    11    21    22    23    31
en-sg        1     2     3    11    21    22    23    31

locales differing from base en = 4  (en-ca, en-ie, en-il, en-sg)
```

So `format('Do')` in `en-ca` produces `1` where `en` and `en-gb` produce `1st`. `Do` is the ordinal-day token, so in those four locales it is indistinguishable from `D`.

**A note on that table, because my first two attempts at it were wrong.** Loading the locale files beside a separately-required copy of dayjs registers them against a different instance, and every `.locale()` call then falls back to `en` in silence. My first run showed all six locales agreeing, which looked like "the bug does not reproduce" and was really "no locale was ever active". The check that caught it:

```
CONTROL en-gb becomes active = true, unknown locale stays en = true
```

An unknown locale string also returned `1st`, which is what gave it away. Worth knowing for anyone else testing locale behaviour here: a wrong locale name is not an error, so a passing locale test can be testing nothing.

The four changed functions are byte-identical to the one already in `en-gb`. The added test walks all 31 days of a month for each of the four locales and compares against `moment` with the same locale, so the expectation is not a table I typed, and 11/12/13 against 21/22/23 are covered by construction rather than by my remembering to include them.

One thing I am not claiming:

```console
$ gh pr checks 3152 --repo iamkun/dayjs
no checks reported on the 'fix/en-variant-ordinal' branch
```

Nothing runs on this branch, so the tables above are mine, produced locally against the published package rather than by your CI.

Happy to rebase, or to take the four functions directly and close this. No attribution needed.

PR #3167, opened 2026-07-29
fix: parse fractional seconds shorter than 3 digits as a decimal fraction

lenamonj:
`REGEX_PARSE` captures fractional seconds as bare digits and `parseDate`
truncates them with `substring(0, 3)` without right-padding, so a fraction
shorter than three digits is read as an integer count of milliseconds:

```js
dayjs('2019-03-25T06:41:00.5').millisecond()  // 5  - moment and new Date() give 500
dayjs('2019-03-25T06:41:00.05').millisecond() // 5  - moment and new Date() give 50
```

ISO 8601 fractional seconds are a decimal fraction. Strings with a trailing
`Z` already take the native `Date` path and are unaffected; only the regex
path is hit.

Fix: right-pad the capture to three digits before truncating. Fractions of
three or more digits are unchanged (`.999999999` still parses as `999`), and
the no-fraction path still yields `0`.

Test: one new case in `test/parse.test.js` beside the existing
`parses unlimited millisecond` test, comparing `valueOf()` and
`millisecond()` against moment for 1-, 2-, and 3-digit fractions. Without
the fix it fails (1 failed, 26 passed in the file); with it the file is 27
passed and the full suite is 774 passed across 93 suites, line coverage
unchanged at 100%.

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -69,7 +69,7 @@ const parseDate = (cfg) => {
     const d = date.match(C.REGEX_PARSE)
     if (d) {
       const m = d[2] - 1 || 0
-      const ms = (d[7] || '0').substring(0, 3)
+      const ms = `${d[7] || '0'}00`.substring(0, 3) // '.5' is 500ms: pad before truncating
       if (utc) {
         return new Date(Date.UTC(d[1], m, d[3]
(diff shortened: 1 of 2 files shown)


lenamonj:
Friendly ping. This is clean against dev; happy to adjust anything you'd like before it goes in.

PR #3187, opened 2026-08-26
fix: diff(year/month/quarter) returns NaN instead of 0 for Invalid Date

### LINKED ISSUE #3186: diff(..., 'year'|'month', true) returns 0 instead of NaN for an Invalid Date operand
**Describe the bug**

`monthDiff` in `src/utils.js` silently swallows `NaN` into `0` via its final `|| 0`:

```js
const monthDiff = (a, b) => {
  const c = b - anchor < 0
    (anchor2 - anchor)))) || 0)
}
```

This is used by `.diff(other, 'year'|'month', true)`. When either side of the diff is an Invalid Date, the intermediate arithmetic naturally produces `NaN`, but the `|| 0` at the end turns that into a plausible-looking `0` instead of propagating the invalidity -- so `.diff()` silently reports "0 months apart" for a comparison that is actually meaningless.

**Reproduction**

```js

const invalid = dayjs('not a date');
const now = dayjs();

invalid.diff(now, 'month', true); // 0 -- should be NaN
invalid.diff(now, 'year', true);  // 0 -- should be NaN

// Compare with day/hour/etc units, which use raw ms diff and correctly propagate NaN:
invalid.diff(now, 'day', true);   // NaN (correct)
invalid.diff(now, 'hour', true);  // NaN (correct)
```

**Expected behavior**

`.diff(..., 'year'|'month', true)` should return `NaN` when either operand is an Invalid Date, consistent with every other unit (`day`, `hour`, `minute`, `second`, `week`, `quarter` all correctly return `NaN` already, since they compute from the raw millisecond `diff = this - that` without an `|| 0` fallback).

**How I found this**

While looking at #3006 / PR #3024 (`duration.humanize()` returning `"a month"` for `dayjs.duration(NaN)`) -- that bug's actual root cause is exactly this `monthDiff` fallback (`humanize()` ends up diffing an Invalid-Date-wrapping instant against "now" in Y/M units, and `monthDiff`'s `|| 0` turns the resulting `NaN` into `0`, which the `relativeTime` threshold loop then reads as "0 months" and prints `"a month"`). PR #3024 already fixes the `humanize()` symptom directly at that boundary, so this issue is just the underlying `monthDiff` behavior itself, which can still surface through any other code path that calls `.diff(..., 'year'|'month')` with an invalid operand (not just via `duration`).

**Information**
- Day.js Version: 1.11.22 (current `dev`)

koding88:
Fixes #3186.

`monthDiff` in `src/utils.js` ended with `|| 0`, which converted the `NaN` produced by Invalid Date operands into a plausible-looking `0`. As a result, `.diff(other, 'year'|'month'|'quarter', true)` silently reported "0 units apart" for a meaningless comparison, while every other unit (`day`, `hour`, `minute`, `second`, `week`) correctly returned `NaN` from the raw millisecond difference.

Replaced `|| 0` with `+ 0` to normalize `-0` to `+0` (the original purpose of the fallback) while letting `NaN` propagate.

```js
dayjs('not a date').diff(dayjs(), 'month', true) // 0 → NaN
dayjs('not a date').diff(dayjs(), 'year', true)  // 0 → NaN
dayjs('not a date').diff(dayjs(), 'day', true)   // NaN (unchanged)
```

Added test covering year/month/quarter for both operand orderings, plus a regression check against moment.js for valid dates. All 798 tests pass, lint clean.


nrps9909 (commented):
Reviewed `07c5726c8a0ac211af5744e9ccaf743a70289af0` against base `0f6c19e3b63bcc3ff74917cb3a60125020c75648`. The `monthDiff` NaN change looks correct: across UTC, New York and Kathmandu, 216 invalid-date observations fail on base and none fail on head (648 observations total). Another 3,600 comparisons of valid dates against Moment agree, including month ends, leap dates, local/UTC instances, floating/integer results and zero sign. Full `npm test` passes all eight timezone legs and 94 suites / 798 tests, meeting the 100% line-coverage gate; lint passes. The previously flagged two-instance millisecond assertion is already fixed on this head.

One scope question before approving the whole diff: `prettyUnit` also adds `special[lower]`, independently of the NaN fix. This changes more than mixed-case `ms`. With a fixed `d = dayjs('2024-06-15T12:34:56.789')`:

| Expression | Base | This head |
| --- | ---: | ---: |
| `d.add(1, 'H').valueOf() - d.valueOf()` | 1 ms | 3,600,000 ms |
| `d.add(1, 'S').valueOf() - d.valueOf()` | 1 ms | 1,000 ms |
| `d.add(1, 'W').valueOf() - d.valueOf()` | 1 ms | 604,800,000 ms |

These were unsupported spellings taking the existing fallback, rather than documented valid inputs regressing. However, [the Add documentation](https://day.js.org/docs/en/manipulate/add) explicitly treats short forms as case-sensitive. Could the unit-normalization change and its get/set tests be separated from this invalid-date fix, or could maintainers confirm the intended API expansion and corresponding documentation/types? The invalid-date correction can stand on its own without that decision.

No upstream status checks were exposed for this head. This is a scoped validation comment, not an approval of the unrelated unit-contract change. Review and local validation performed with Codex.

PR #3188, opened 2026-08-27
fix: return NaN from year/month/quarter diff with Invalid Date

### LINKED ISSUE #3186: diff(..., 'year'|'month', true) returns 0 instead of NaN for an Invalid Date operand

22elix3r:
## Summary
`.diff(..., 'year'|'month'|'quarter', true)` no longer reports `0` when either operand is an Invalid Date. The result is `NaN`, matching every other unit.

## Related Issue (Fixes #3186)
Fixes #3186

## Changes Made
- `monthDiff` in `src/utils.js` ended with `|| 0` to turn `-0` into `+0`. That same fallback also turned `NaN` into `0` when year/month arithmetic involved an Invalid Date.
- Replaced `|| 0` with `+ 0` so `-0` is still normalized to `+0` while `NaN` propagates.
- Added tests in `test/display.test.js` for year/month/quarter (float and integer) in both operand orders, plus a `+0` regression check for equal valid dates.

## Testing
- `npx jest test/display.test.js --coverage=false`
- `npm test` (795 tests, 100% line coverage)

## Notes
`day`, `hour`, `minute`, `second`, and `week` already returned `NaN` from the raw millisecond difference. Only the calendar-unit path through `monthDiff` was affected.


22elix3r:
Ready for review whenever you have a moment. I cannot request a reviewer from this account.

22elix3r:
Withdrawing: I missed that this issue already had earlier open PRs. Sorry for the noise — leaving the field to those patches.

22elix3r:
Withdrawing this PR so the earlier open work on the issue can proceed.

PR #3204, opened 2026-09-03
[LOW] Pin actions/setup-node to its reviewed commit

OskarEichler:
## Summary

Both test and privileged release workflows referenced actions/setup-node through the mutable v3 tag. A moved tag or upstream compromise could cause unreviewed action code to execute.

## Fix

Pin both uses to the reviewed v3 commit 3235b876344d2a9aa001b8d1453c930bba69e610 while retaining an inline v3 annotation.

## Validation

- Diff checks passed

## Breaking changes

None.


lbesecker195 (approved):
Verified: `3235b876` is exactly what the `actions/setup-node@v3` tag resolves to today. LGTM.

PR #3205, opened 2026-09-03
[LOW] Pin actions/checkout to its reviewed commit

OskarEichler:
## Summary

Both test and privileged release workflows referenced actions/checkout through the mutable v3 tag. If that tag were moved or the upstream action were compromised, unreviewed action code would execute in these workflows.

## Fix

Pin both uses to the reviewed v3 commit a37ce9120846195fa4ece8f58b268e6043cb2f26 while retaining an inline v3 annotation for maintainability.

## Validation

- Diff checks passed

## Breaking changes

None.


lbesecker195 (approved):
Verified: `a37ce912` is exactly what the `actions/checkout@v3` tag resolves to today. LGTM.

PR #3206, opened 2026-09-03
[LOW] Pin the Codecov action to its reviewed commit

OskarEichler:
## Summary

The reusable test workflow referenced codecov/codecov-action through the mutable v4 tag and passes it the Codecov upload token. A moved tag or upstream compromise could expose that token or alter uploaded coverage data.

## Fix

Pin the action to reviewed commit b9fd7d16f6d7d1b5d2bec1a2887e65ceed900238 while retaining an inline v4 annotation.

## Validation

- Diff checks passed

## Breaking changes

None.


lbesecker195 (approved):
Verified: `b9fd7d16` is exactly what the `codecov/codecov-action@v4` tag resolves to today. Pinning the action that receives the token is the right one to prioritize. LGTM.

PR #3215, opened 2026-09-13
fix(duration): parse decimal commas in duration strings

Hugohong258:
The duration parser's regular expression already accepts commas in numeric components, but converting those captures directly with `Number()` produces NaN. For example, `dayjs.duration('PT1,5S').asMilliseconds()` returns `0` instead of `1500`.

Normalize decimal commas to points before converting captured strings to numbers. Add tests for comma/point equivalence across duration units, negative seconds and the normalized ISO output.

Validation: the two new tests fail with the original implementation. After the fix, all 32 duration tests pass, ESLint passes for the changed files, and the full Jest run with coverage passes all 796 tests across 93 suites. The separate multi-time-zone npm script matrix was not run.


lbesecker195 (approved):
`DURATION_REGEX_PARSE` on dev already accepts `[0-9,.]` in each component, but `Number('1,5')` is `NaN`, so `PT1,5S`, `P1,5Y` and `PT-1,5S` parse to NaN components today. With the replace they become 1.5 / 1.5 / -1.5, matching the point forms. ISO 8601 allows a comma as the decimal sign, and a malformed `PT1,5,5S` is still NaN either way since only one comma per component is normalized. LGTM.

